PR #4, opened 2020-04-18
inserted copyright notice into all iHAMOCC files

JorgSchwinger:
Inserted a copyright notice (format as proposed by Mats) into all files in the /hamocc and /trc sub-directories.


TomasTorsvik:
I was reviewing the changes, and I noticed that several files have been removed. Perhaps this should be mentioned specifically in the comment to the commit?

Deleted files:
- hamocc/c14_correction.F90
- hamocc/chksumbgc.F
- hamocc/co2_ems.F90
- hamocc/mo_gridremap.F90
- hamocc/mo_kind.F90
- hamocc/mo_satm.F90
- hamocc/mo_timeser_bgc.F90

matsbn:
For the ocean specific copyright statements I was planning to add a year range for when modifications done by persons were done (e.g. Copyright (C) 2012-2019 Person A, Person B). I see that you mostly use 2020 for the copyright notice. Using 'git log --follow <file name>' it is possible to see the complete commit history for a file (also before a file was renamed or moved). Although clearly more work, would it be an idea to put in the actual year or year range when the modifications were done?

mvhulten (approved):

JorgSchwinger:
Thanks for your comments. I have updated my commit taking into account most of the comments above.
-Commit message mentions deleted files
-Files in the /trc sub-directory are no longer included

Regarding the year of the copyright: I it would take quite some time to work all this out, plus it would not really be correct in the end (e.g. I don't believe I would be able to track back Karens work). Maybe more important, we obtained the permission to publish iHAMOCC from the original copyright holder this year, I do not think we had any copyright before (although this is probably legally complicated matter).

Since it is a new commit in my fork of BLOM I'll close this pull request now and open a new one (not sure this is necessary, github seems to have recognized my updates in this pull request already, but I'll close it anyway to be sure).

PR #5, opened 2020-04-19
Inserted copyright notice into all iHAMOCC files. Deleted unused file…

### LINKED ISSUE #2: License for BLOM
It would be useful to clarify the license for BLOM. Assuming we are free to chose whatever license, I would suggest (having consulted with Marco) to go for either a GNU Lesser General Public License v3.0 (lgpl-3.0) or an Apache license 2.0 (apache-2.0).

JorgSchwinger:
…s from iHAMOCC directory.

Updated as requested. Files in /trc directory not included any longer. Kept year of copyright as before (2020).


TomasTorsvik:
About the copyright statement, I think the main thing is that original contributions are recognized, and probably the exact year of each contribution is less important. Perhaps one could consider setting a range instead of single year for contributions prior to iHAMOCC (e.g. <start year> - 2019), but personally I'm fine with keeping it with a single year as Jörg already has done it.

TomasTorsvik:
I assigned Mats to this pull request, to sign off on it.

PR #7, opened 2020-05-05
Replace MICOM with BLOM

matsbn:
This pull request is related to other pull requests in NorESMhub repositories (NorESM, cime, CAM and CESM_CICE5) to complete the transition from MICOM to BLOM.

I believe the most profound change is that now output and restart files are called \*.blom.\*.nc compared to \*.micom.\*.nc before. NorESM can still branch from old experiments using \*.micom.\*.nc restart files. Other notable changes is that BLOM is placed in the components/blom folder of NorESM and that $DIN_LOC_ROOT/ocn/blom is the root of BLOM input data compared to $DIN_LOC_ROOT/ocn/micom before.


JorgSchwinger (commented):
Hi Mats,

looks very good to me, it's good to have this done. Just one thing I saw in the iHAMOCC code: It great that the model can read old restart files (with .micom. in the file name). But if PNETCDF is active this seems not to be the case for iHAMOCC (file aufr_bgc.F90)? Do I misunderstand something or is this intentionally?

matsbn:
Hi Jörg,

This is not intentional and must be corrected in the PR before merging. Thanks for spotting this.

matsbn:
I have updated aufr_bgc.F90 and verified correct restart with PnetCDF enabled.

JorgSchwinger (approved):

PR #10, opened 2020-05-09
Updated generation of input data list and netCDF settings

matsbn:
The input data list generated by BLOM should now be correct. Also netCDF settings are now taken from variables set in env_run.xml (PIO_TYPENAME and PIO_NETCDF_FORMAT).

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -24,9 +24,24 @@ setenv BLOM_RIVER_NUTRIENTS `./xmlquery BLOM_RIVER_NUTRIENTS --value`
 setenv BLOM_N_DEPOSITION `./xmlquery BLOM_N_DEPOSITION --value`
 setenv BLOM_N_DEP_FILE `./xmlquery BLOM_N_DEP_FILE --value`
 setenv RUN_STARTDATE `./xmlquery RUN_STARTDATE --value`
-echo $CASE ${EXEROOT} $CASEBUILD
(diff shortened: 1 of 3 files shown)


JorgSchwinger (approved):
Hi Mats

looks good so far, but for HAMOCC there are still a few things missing.

1) The initial conditions files for tracers, which partly depend on some options (whether C-isotopes are activated, although this is still not really supported I would say)
2) CFC concentration input files if CFCs are activated

I'll approve this now and I suggest we'll open a new issue to fix the remaining files for HAMOCC.

Just one more question: The part 

if ($RUN_TYPE == startup) then
...  
else if ($ICFILE =~ *.blom.r.*) then
    set icfile = `echo $ICFILE | rev | cut -c 2- | rev | cut -c 2-`
...
    if ($ecosys == TRUE) then
      set icbgcfile = `echo $ICFILE | sed "s/.blom.r./.blom.rbgc./"`

means you can have a startup run from restart files.  Is this working with HAMOCC?

matsbn:
Hi Jörg,

I felt the HAMOCC input data list generation was more up to date than BLOM and I tried to make sure this PR was consistent with what was already there for HAMOCC. Great if the remaining HAMOCC input data can be listed.

The ability of MICOM and HAMOCC to startup from restart files was committed here: 417670dafeb41adc88fbb5c5141b7f4b5df48386. I am quite sure I tested this with HAMOCC enabled, but it would be great to test again. Typically one would set ICFILE = "'\<blom restart file\>'" in user_nl_blom to test.

Then I will merge this PR.

PR #20, opened 2020-06-15
Added science_support indicators for OMIP type compsets in config_com…

JorgSchwinger:
…psets.xml

As discussed with Aleksi, I didn't add the 1/4-degree since this configuration is not enough tested to call it "scientifically supported"


matsbn (approved):
This looks reasonable, thanks!

PR #22, opened 2020-06-16
Added a "dummy" testlist_blom.xml

matsbn:
Added a "dummy" testlist_blom.xml so that when unsupported compsets and grid combinations are used, a meaningful warning is displayed.


JorgSchwinger (approved):
I don't fully understand how this works, but I assume you have tested this.

PR #35, opened 2020-07-13
Add multiple instance of CESM2 support.

blcc:
The multiple instance support is added.
Following the method of CAM, the phy/mod_instance.F90 is add to get instance suffix string ( '_0001', '_0002' and so on).
The cime_config/buildnml is also modified to generate the namelists for correspond instance.

Tested multi-instance setting:
./xmlchange NINST_ATM=2
./xmlchange NINST_LND=2
./xmlchange NINST_ICE=2
./xmlchange NINST_ROF=2
./xmlchange NINST_GLC=2
./xmlchange NINST_OCN=2
./xmlchange NINST_WAV=2
./xmlchange NINST_ESP=2
./xmlchange MULTI_DRIVER=TRUE

Total tasks will be original NTASKS * NINST with MULTI_DRIVER=TRUE.

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -26,6 +26,8 @@ set HAMOCC_CISO = `./xmlquery HAMOCC_CISO --value`
 set RUN_STARTDATE = `./xmlquery RUN_STARTDATE --value`
 set PIO_TYPENAME_OCN = `./xmlquery PIO_TYPENAME_OCN --value`
 set PIO_NETCDF_FORMAT_OCN = `./xmlquery PIO_NETCDF_FORMAT_OCN --value`
+## for multiple instance
+set NINST_OCN = `./xmlquery NINST_OCN --value`
 cd $EXEROOT/ocn/obj
 set CONFIG_OCN_DIR = $CONFIG_OCN_DIR1/../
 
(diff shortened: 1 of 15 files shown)


matsbn:
Great with multiple instance support in BLOM! Before approving a few things must/should be resolved:

- When I tried to compile BLOM including the PR commits, hamocc/aufr_bgc.F90 failed on lines 206, 216 and 217. I think you want "trim(inst_suffix)" instead of "trim(lenInRstfn)" in these lines.
- "mod_instance.F90". This new file needs the copyright/license notice at the top of the file. See any BLOM source file for a template. For other modules, module name corresponds with file name. Could you consider to call the module "mod_instance"?
- Variable "lenInRstfn". To conform with the existing code style (and since Fortran is case-insensitive) I suggest to only use lower-case letters in variable name ("leninrstfn"). 
- "cime_config/buildnml": I would delete the two comments "## for multiple instance". Confusing with mixed indentation level in the foreach loop (lines 694-1640). Please use indention level of 2 (the cat ... EOF blocks cannot be intended however).
- Several places a space is inserted before trim(inst_suffix) in concatenated strings. I would delete the space to conform with existing style.
- "drivers/cpl_mct/sumsbuff_mct.F": Commented lines 87-88 seems to be left-over from some test and should be deleted.

matsbn:
Another comment: In "mod_instance.F90", do not write to the "*" I/O unit. The message will end up in cesm.log.*, while it should go to ocn.log.*. Use the unit stored in the "lp" variable.

blcc:
Thanks for comments. Pull request will be reopen after all issue fixed.

blcc:
Coding style conformed. And also bugs fixed.

matsbn:
Thanks for making these changes, @blcc. I tested briefly with a NOIIAOC compset and it worked fine. I will merge.

PR #40, opened 2020-08-13
Bugfix multiple instance of CESM2 support

matsbn:
The merged PR https://github.com/NorESMhub/BLOM/pull/35 causes BLOM not to compile in standalone configuration since the PR introduced dependency on CESM modules for all configurations. This PR corrects this. With PR commits included, BLOM passed testing in single instance configuration in NorESM and for standalone configuration. Great if you can test if it works for NorESM multiple instance configuration, @blcc.


blcc:
I tested this PR with 2 instance of NorESM on Betzy. It works perfectly.
Thanks @matsbn .

matsbn:
Thanks for testing, @blcc. I will then merge this PR.

PR #42, opened 2020-08-26
Corrected bug in BLOM's buildnml script, iHAMOCC initial condition fi…

### LINKED ISSUE #37: Write HAMOCC initial condition files to blom.input_data_list
HAMOCC initial condition files need to be written to blom.input_data_list also for runtype/=STARTUP. The model attempts to read them in any case so they need to be present/downloaded. This will resolve https://github.com/NorESMhub/NorESM/issues/149

@matsbn @TomasTorsvik we need to discuss where to put these kind of changes now. Certainly into the master branch (but then master should first be synchronised with the release-1.0 branch). But probably also into release-1.0, since this will be a candidate included in future minor releases?

JorgSchwinger:
…les are now correctly written to blom.input_data_list in any case. This resolves #37 and NorESMhub/NorESM#149.

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -1658,6 +1658,8 @@ cat >> $CASEBUILD/blom.input_data_list << EOF
 sss_climatology_file = `echo $SCFILE | tr -d '"' | tr -d "'"`
 EOF
 endif
+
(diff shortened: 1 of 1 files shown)


matsbn (approved):
This looks good.

PR #46, opened 2020-08-30
Revised time management

matsbn:
This is my first stab at replacing some of the common blocks in header files with modules and using modern Fortran language features and style for new code. The commit of this PR revises the time management with unified calendar routines. To avoid common blocks in new modules, certain groups of variables and constants have been moved from common blocks into modules.

The code has been tested gives bit-identical results in standalone configuration and for startup, hybrid and continue runs coupled to NorESM.


JorgSchwinger (approved):
Looks good (although of course too much  to really check in depth).

TomasTorsvik (approved):
Good to see common blocks being replaced with modules! As Jörg said, it's a bit too many changes to make a thorough review, but I guess this is almost unavoidable when you change things regarding timing and calendar in the code.

PR #48, opened 2020-10-01
Refactor sediment code

### LINKED ISSUE #41: Replace DO statements with seared termination (CONTINUE) with DO ... END DO format
DO with shared termination was declared obsolescent in Fortran 90, but still part of the Standard. 
Example:
      DO 100 J = 1,NJ
      DO 100 I = 1,NI
      X = X + ARRAY(I,J)
100   CONTINUE

This style still appears in some places in BLOM/iHAMOCC. I would like to replace these instances with the more modern DO ... END DO format.

TomasTorsvik:
Rewrite of files powach.F90, powadi.F90, dipowa.F90, bodensed.F90, ocprod.F90

- Replace all DO ... CONTINUE blocks with DO ... ENDDO
- Set intent for variables passed to subroutine
- Replace F77-style relational operators with symbolic equivalents
- Replace fixed-form with free-form spacing convention

Test run for 1 year simulation of NOINYOC_T62_tn21 gives bit-exact same results as current "master" (max/min values zero for diff of all variables in *.hy.nc and *.hbgcy.nc files).

### DIFF
--- a/hamocc/bodensed.F90
+++ b/hamocc/bodensed.F90
@@ -169,5 +169,4 @@ subroutine bodensed(kpie,kpje,kpke,pddpo)
   enddo
 
 
-  return
 end subroutine bodensed
--- a/hamocc/bodensed.F90
+++ b/hamocc/bodensed.F90
@@ -17,7 +17,7 @@
 ! along with BLOM. If not, see https://www.gnu.org/licenses/.
 
 
-      SUBROUTINE BODENSED(kpie,kpje,kpke,pddpo)
+subroutine bodensed(kpie,kpje,kpke,pddpo)
 !**********************************************************************
 !
 !**** *BODENSED* - .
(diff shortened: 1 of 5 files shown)


matsbn:
I think this looks fine, but naturally Jörg's opinion is more important than mine here. I would skip the return statement at the end of subroutines as they have no function. I know they are plentiful in the BLOM code, but I am removing them as I go through the code now.

matsbn (approved):

JorgSchwinger (approved):
Hi Tomas,

looks fine for me too. I am wondering two things: 1) what about Marco's code? Will you introduce his changes later on to of this? 2) regarding ocprod.F90: should we take the opportunity to fix the weird ordering of loops (loop1-loop4 as you call them now). I don't know why the k-loop is innermost in all these loops, I guess this has been some optimization for a specific machine long time ago. The ordering should be k,j,i in all loops except those dealing with sinking. But maybe this would be a separate commit.

TomasTorsvik:
Thanks for your feedback. 
To Mats' point, I agree that it makes sense to remove the return statements as part of this code rewrite, so I will go back and do that. To Jörg's points, (1) This rewrite doesn't contain any of the burst coupling elements from Marco's code, but Marco was rewriting some of thee routines in F90 style while implementing the burst coupling. I hope therefore that introducing the F90 style will create smaller diffs (and more meaningful ones) when Marco's code is introduced later on. (2) That's a good point, I will look at that once more, but as you suggest, I think this is better done as a separate commit.

PR #54, opened 2020-10-16
Added support for different compilers

nordmoen:
This commit updates the build instructions and adds support for multiple
compilers through different `.make` files. A compiler can be selected at
build time as long as the corresponding `.make` file exist for that
compiler. By default the build will use `mpiifort` in order to be
backwards compatible.


nordmoen:
Just an additional note, I had to change some parameters in `phy/dimension.F` as well to get it to work, but I thought I would leave those changes out as they would not be backwards compatible.

TomasTorsvik:
Thanks @nordmoen , this looks good. Since this is a feature branch I will just merge it right away and test it afterwards. Could you perhaps make an issue about the changes you needed to introduce in phy/dimension.F ? Perhaps this could also be useful to others.

PR #64, opened 2021-01-22
Add CI integration with Github Actions

nordmoen:
This change adds CI setup for Github Actions which will automatically test changes to the `meson` branch on Ubuntu and macOS with both the default compiler (`gfortran`) and `ifort` from Intel. This allows one to check if changes introduced is compatible with all systems that BLOM should support.

Additional tests can also be added in future, e.g. testing with different versions of `gfortran` on macOS and Ubuntu or adding Windows support.


matsbn:
This looks very interesting and also non-intrusive. I have not tried GitHub Actions before and would be happy to test it in this branch.

PR #66, opened 2021-01-25
Added support for compiling hamocc with Meson

### LINKED ISSUE #65: Build HAMOCC with Meson
Currently, adding all the source files in `hamocc` does not lead to a valid build with Meson (checkout the [`meson`](https://github.com/NorESMhub/BLOM/tree/meson) branch).

There are quite a few errors when compiling so I'm unsure how to proceed with enabling `hamocc`. Could someone with more experience comment on which files are required by `hamocc` and if there are special flags that are needed to compile. In addition, should `hamocc` be placed behind a feature flag in Meson in the same vein as `MPI` support?

With the new [`Actions` support](https://github.com/features/actions) added to the `meson` branch, any pull requests made are automatically tested. This can be seen under the `Actions` tab in the Github interface. So if you want to experiment, simply create a pull request against the `meson` branch and add `hamocc` support.

nordmoen:
This solves #65.

### DIFF
--- a/hamocc/blom2hamocc.F
+++ b/hamocc/blom2hamocc.F
@@ -153,7 +153,8 @@ c$OMP PARALLEL DO PRIVATE(kn)
       kn=k+nn
       do l=1,isp(j)
       do i=max(1,ifp(j,l)),min(ii,ilp(j,l))
-        ocetra(i,j,k,:)=trc(i,j,kn,itrbgc:itrbgc+ntrbgc-1)*bgc_rho(i,j,k)
+        ocetra(i,j,k,:)=trc(i,j,kn,itrbgc:itrbgc+ntrbgc-1)
+     .                 *bgc_rho(i,j,k)
       enddo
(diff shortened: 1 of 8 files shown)


JorgSchwinger:
Regarding the problem with restart_hamoccwt.F, it would be better to fix the cause of the problem (the truncated line) instead of using a compiler flag to ignore the problem. The truncated line has been introduced just recently.

nordmoen:
> Regarding the problem with restart_hamoccwt.F, it would be better to fix the cause of the problem (the truncated line) instead of using a compiler flag to ignore the problem. The truncated line has been introduced just recently.

That would be a better solution, but I would need a bit of help as I'm not familiar at all with Fortran syntax.

nordmoen:
I merged PR #67 into this PR which enables integration of `iHAMOCC` without adding a new compiler flag.

matsbn (approved):

PR #68, opened 2021-03-05
Added gcc and intel compiler debug flags for Meson build.

matsbn:
Added some debug flags for the gcc and intel compilers that should not cause execution failure. Concerning macOS, the default Meson version in Macports and Brew are currently 0.57.1 which fails to build due to incorrect source dependency generation. This is a know issue (https://github.com/mesonbuild/meson/issues/8395) that will likely be fixed soon.


nordmoen (approved):
Looks good to me!

PR #69, opened 2021-03-08
Meson build system and CI workflow for automatic build testing

matsbn:
Meson is a promising build system for standalone BLOM application. The CI workflow automatically tests building on several systems and with different compilers. Both the Meson build system and CI workflow are non-intrusive and does not change BLOM operation when built as a NorESM component (the Meson branch compared to master gave bit-identical results in NorESM). A failed CI build test will not obstruct merging of commits, and therefore is merely a diagnostic of code portability (currently the macOS build tests fail because of a known Meson issue).


.github/workflows/ci.yml
```diff
@@ -0,0 +1,117 @@
+name: Continuous Integration testing
+on:
+  push:
+    branches:
+      - meson
```
nordmoen: This should probably be changed to `master` so that CI runs on pushes to `master` and not the `meson` branch

.github/workflows/ci.yml
```diff
@@ -0,0 +1,117 @@
...
+  pull_request:
+    branches:
+      - meson
```
nordmoen: This should also be changed to `master` (see comment above)

matsbn: Thank you for pointing this out. I have committed a modified .github/workflows/ci.yml so the tests will be done on master branch.

nordmoen:
It is also possible to simplify the CI configuration a bit ([ref](https://github.com/nordmoen/BLOM/blob/meson/.github/workflows/ci.yml)) which would avoid the failing tests on macOS, but that can be a new feature after this pull request has been merged.

JorgSchwinger (approved):

TomasTorsvik (approved):
I made a small change in the README.md file. Should have included this as a suggestion, but I messed it up a bit. Still, it's should be just a minor thing. Looks fine otherwise.

AleksiNummelin (approved):
Looks good! Latest Ubuntu build seems to be failing now, but I think that is acceptable (and might not have anything to do with the build system).

PR #70, opened 2021-03-10
Added Fukamachi et al. (1995) test case

matsbn:
The Fukamachi et al. (1995) test case is a periodic channel configuration where the length of the channel matches the most unstable wave length of a baroclinic jet. Thus, it simulates the genesis of a single eddy. The test case has a rather small memory footprint and does not require any input files besides a namelist. Revised options and their handling in the Meson build system have also been committed. The modified code has been tested under NorESM (NOIIAOC compset) and gives bit-idential results.


.github/workflows/ci.yml
```diff
@@ -2,10 +2,10 @@ name: Continuous Integration testing
 on:
   push:
     branches:
-      - master
+      - feature-fuk95
```
nordmoen: These should not be changed here (and are also why the checks have not run for this PR, I think)

fuk95/mod_fuk95.F90
```diff
@@ -0,0 +1,355 @@
...
+                          taux, tauy
+   use mod_mxlayr, only: mltmin
+   use mod_state, only:  v, temp, saln, sigma, phi
+   use mod_checksum, only: csdiag, chksummsk
+
+   implicit none
+
+   private
+
+   real(r8), parameter :: &
```
AleksiNummelin, line 48: There could be some benefits from defining these parameters in the 'limits' namelist, especially if this case will be used for something else than just testing the stability of the code. For example, a perturbation study would be very clean to do if one can use the same executable, but just change the namelist. However, this is fine for now and can be changed at later stage if needed.

.github/workflows/ci.yml
matsbn: Thanks for pointing this out, @nordmoen! I keep changing the branch for CI testing while working on it and forget to change back. The CI test on ubuntu fails because fetching http://azure.archive.ubuntu.com/ubuntu/pool/main/n/nvidia-settings/libxnvctrl0_440.82-0ubuntu0.20.04.1_amd64.deb fails. I am not sure why the tests on macOS keep getting cancelled. There might still be a meson 0.57.1 issue on macOS, but the test seems to get cancelled before getting to this. Tests with Intel compilers have persistently passed.

fuk95/mod_fuk95.F90
matsbn: I agree. Should be a namelist group particularly for this test case I guess. I suggest to add this later.

JorgSchwinger (approved):

TomasTorsvik (approved):

.github/workflows/ci.yml
nordmoen: They are simply canceled because they are in the same `matrix` in Github action parlance. We can change the [`fail-fast` behavior](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) of this setup so that this is not the case.

AleksiNummelin (approved):

PR #71, opened 2021-03-15
Single column test case

milicak:
This is a single column test case to try different mixing schemes or vertical remapping methods. To run the test case, one needs an initial condition file and NCEP atmospheric forcing files which will be put in betzy soon.
The folder tree is similar to fuk95 test case, i.e. there is a "single_column" folder at the same level of phy/.
However, in the future we should put all these idealized test cases under a different folder such as "tests" or "examples".
The single column test case has been built by meson and run in betzy using one core.


nordmoen:
This seems to fail on `macOS` with the error `Fatal Error: Cannot open module file 'mo_carbch.mod' for reading at (1): No such file or directory`, maybe something to look into before merging?

milicak:
I am not quite sure why it failed because the code was not supposed to compile hamocc anyhow. 
ecosys boolean is false in meson_options.txt. 
However, there is another problem before that which is;
../idlage/idlage_init.F:26:10:

   26 |       use mod_xc
      |          1
Fatal Error: Cannot open module file 'mod_xc.mod' for reading at (1): No such file or directory.

This cannot be true, because I tried in my mac as well and it compiles without any problem.

PR #72, opened 2021-03-15
Update Github CI to run on all branches and simplify

nordmoen:
This update removes the branch notation from Github CI so that all pushes and pull requests are run through CI. This change simplifies CI and ensures that one can create feature branches separate from `master` and still have them tested.

In addition, this PR simplifies building on the default runners (`Ubuntu` and `macOS`) to use a preconfigured action with Meson.

I've also added `fail-fast: false` to the `strategy` so that if one of the default runners fails the others will not be aborted.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,6 @@
 name: Continuous Integration testing
-on:
-  push:
-    branches:
-      - master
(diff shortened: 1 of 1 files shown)


matsbn (approved):
Seems to solve the failed checks, great!

PR #76, opened 2021-03-23
Feature highres

AleksiNummelin:
This pull request introduces the BLOM 1/8 deg (tnx0125) to NorESM, companion pull requests are made to other components. BLOM edits introduce grid-specific changes and modifications to the default namelist parameters (turning off eddy parameterization for example). Note that functionality for HAMOCC is included, but that has not been activated/tested, and not all forcing files are available.

Tested configurations: N1850frc2NOECO_f09_tnx0125v4, NOINY_T62_tn01254
 
!Note! from the companion CIME pull request: to run on Betzy, one also needs to change either config/cesm/machines/config_compilers.xml or Macros.make in a case folder, so that FFLAGS includes: -mcmodel medium option.


matsbn:
Looks good to me!

PR #77, opened 2021-03-24
Added new pe-layout for OMIP1 compsets with iHAMOCC using the tnx2 gr…

JorgSchwinger:
…id; to be used on betzy

I also tried to start with making the config_pes.xml file a bit nicer by providing some comments.


matsbn (approved):
This looks fine.

PR #78, opened 2021-04-08
Feature channel

AleksiNummelin:
This pull request brings in a new idealized channel setup that is been actively used in the KeyCLIM project. The channel setup is self-contained (needs no input) and easily modified by the user (bottom topography, initial stratification, and wind forcing are all controlled by namelist parameters). One can also easily modify the setup to include more complex forcing or idealized conditions.


AleksiNummelin:
The pull request should finally be ready. The branch is updated with the latest master, so no more conflicts and the code is tested with NorESM.

AleksiNummelin:
Mats pointed out some mistakes in OpenMP syntax and random number generation. Those are now fixed and the channel setup is verified to produce the same results with and without OpenMP. I also discovered the following behavior linked to a combination of OpenMP/pnetcdf/compressed output on Betzy

OpenMP disabled + pnetcdf  + uncompressed restart format → works
OpenMP disabled + netcdf    + uncompressed restart format → works
OpenMP                + pnetcdf  + uncompressed restart format → works
OpenMP                + netcdf    + uncompressed restart format → works

OpenMP disabled + pnetcdf  + compressed restart format → works
OpenMP disabled + netcdf    + compressed restart format → works
OpenMP               + pnetcdf  + compressed restart format → restart write fails (variable exceeds dimensions)


The recommendation is to either run without OpenMP or with OpenMP + pnetcdf + uncompressed restart. Curiously, the compressed output format does not cause the same error (only the restart write).

matsbn (approved):
Looks good now. CI build on macOS failed, but I consider this a library issue and not a problem with modified code in this PR.

PR #82, opened 2021-04-22
Simplify and cache Intel CI setup

nordmoen:
This PR simplifies and speeds up the Intel CI setup. This is accomplished through caching the installed Intel compiler and the building of netCDF-Fortran through a standard Github action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,16 +13,7 @@ jobs:
       matrix:
         os: [ubuntu-20.04, macos-10.15]
         mpi: [true, false]
-
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
-        with:
-          ref: meson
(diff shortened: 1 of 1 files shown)


matsbn (approved):
Looks good! CI build on macOS fails because include file mpif.h cannot be found. I assume this is an issue the openmpi installation, but I'll approve anyway.

nordmoen:
> Looks good! CI build on macOS fails because include file mpif.h cannot be found. I assume this is an issue the openmpi installation, but I'll approve anyway.

The MacOS has nothing to do with this PR. This PR simply speeds up Intel builds (after first caching occurs).

I have tried to look into the installation issue, but I can't find out why. It seems that an update to the brew setup has changed, the challenge is that the installation is correctly picked up by Meson, but then the `include` folder is empty.

PR #83, opened 2021-04-23
Initial set of utility scripts to support standalone BLOM functionality.

### LINKED ISSUE #86: Restructuring of BLOM directory
In a BLOM-core meeting today with @matsbn @TomasTorsvik @milicak and @JorgSchwinger we discussed the need for some restructuring of the code and putting together a minimum standalone running infrastructure. Specifically, there is a need to move the idealized test cases under a specific folder (or possibly outside the BLOM git structure). Overall, these changes are needed for #80 and some sort of initial try was given in #83.

We felt that we should take a bit of time to make a good choice for the structure so I created a [doc](https://docs.google.com/presentation/d/10l86awfuxMBd-3lDs8uHxnicsGpw8q9ERGXZrAVVH8k/edit?usp=sharing) that we can use for planning. I put there an illustration of the current structure and a suggestion.

- Outline different possibilities for the structure
- Agree on a which structure to proceed with
- git-mv existing directories to the new structure
- Make a new pull request or modify #83 to bring in new changes

AleksiNummelin:
This pull request introduces a new folder and a set of scripts that are useful when running BLOM on standalone mode. Most notably, it includes limits.json and limits_from_json.py that create the limits namelist file needed by BLOM. At a later stage, this python-based functionality could replace the current shell script cime_config/buildnml used in the NorESM system.

The limits_from_json.py reads in the master limits.json file and optionally a user-provided user_nl_blom namelist and creates a final limits file for BLOM. In future, one could either have a configuration specific limits.json file, or a configuration specific user_nl_blom file (perhaps renamed to something else). I believe the latter option would be better.

The limits.json file in this pull request is mostly specific to the 'channel' setup and does not include iHAMOCC namelists. This can functionality can be easily added.

In addition, this pull request includes a run script that can be used on Betzy and a HOWTO file (pending proper documentation).

### DIFF
--- a/utils/HOWTO_standalone
+++ b/utils/HOWTO_standalone
@@ -1,0 +1,25 @@
+########################################
+# Quick HOWTO on running BLOM on Betzy #
+########################################
+#
+# (1) create a case directory 
+mkdir ../../cases/BLOM_channel
+# (2) Copy the contents of the utils 
+#     folder to the case folder
+cp * ../../cases/BLOM_channel/
+#
+# (3) Move to the case folder and
+#     create the namelist.
+#     If needed, edit the user_nl_blom
+#     prior to running the limits_from_json.py.
+#     Note that the script assumes python3.
+cd ../../cases/BLOM_channel/
+module load Python/3.8.6-GCCcore-10.2.0
+python limits_from_json.py
+#
+# (4) Edit and run the run.sh_normal_betzy. 
+#     Usually the EXPID should match the RUNID 
+#     in the limits file. The run script will
+#     create a folder under /cluster/work and
+#     submit the blom exectutable.
+./run.sh_normal_betzy
--- a/utils/limits.json
+++ b/utils/limits.json
@@ -1,0 +1,543 @@
+{"LIMITS": { 
+    "NDAY1": {
+        "definition": "First day of integration (i)", "value": [0]},
+    "NDAY2": {
+        "definition": "Last day of integration (i)", "value": [5400]},
+    "IDATE": {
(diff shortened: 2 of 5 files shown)


milicak:
I am not quite sure this is the right way to go, because there are multiple idealized experiments in BLOM as of now.
We should dicuss how to run all these idealized experiments stand alone. 
I suggested @matsbn  before to create another project such as BLOM_examples and commit all the necessary scripts such as in this pull to that project. That is one way, there might be others. It is better to discuss this.

AleksiNummelin:
We should probably have a separate discussion. I think the main contribution here is the functionality to create limits files and I think that type of functionality could live in a 'utils' folder. I think @matsbn has created some infrastructure for creating a case when running outside NorESM.

AleksiNummelin:
Whatever the optimal format of including these changes, I realized that actually the idealized setups should be accompanied with `limits` file, because otherwise, one doesn't know the namelist settings those cases are supposed to be run with.

matsbn:
Not sure what is best, having run configurations in a separate repository or in the BLOM repository itself. So a good idea to have a discussion on this.

TomasTorsvik (approved):
Hi, this PR has been dormant for way too long. I suggest we include it in the `master` branch since (1) it includes useful information on running BLOM on Betzy, (2) it doesn't break anything, and (3) we don't have any immediate plans to store this information elsewhere.

Should we decide to make a new structure for external utilities, we can always pull it out again in the future.

TomasTorsvik:
This one has been stalled for a long time. I don't think this PR will be merged in at this point, so I close it now.

PR #88, opened 2021-05-03
OpenMP fix and disabling answer changing optimizations

matsbn:
Corrected an OpenMP directive and added an Intel compiler flag that disables optimizations changing floating point results.


TomasTorsvik:
Looks good to me. Do you need something more for the review process?

PR #89, opened 2021-05-04
Initial unit testing implementation

nordmoen:
This PR contains the initial implementation for unit testing in BLOM. The PR introduces a new sub-directory, `tests`, which contains unit tests to run. Currently only `fuk95` is complete and running. The README has been updated to reflect the unit testing setup.

This PR also contains an updated CI configuration which will run both build and test for all targets and separates out OpenMP to catch problems earlier. In addition this PR skips MacOS builds with MPI as that has given some strange dependency issues. Tests are currently not run with MPI as that is not working locally either.

The next steps for unit testing would be to parse the output of the completed runs and compare the result with known good output, so that deviations from this can automatically be detected.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,44 +11,64 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-20.04, macos-10.15]
+        os: [ubuntu-latest, macos-latest]
         mpi: [true, false]
+        openmp: ['enabled', 'disabled']
+        exclude:
+          # Do not build on macOS with MPI as that is having some dependency issues
+          - os: macos-latest
+            mpi: true
     steps:
       - name: Install dependencies - Ubuntu
         run: |
           sudo apt update
-          sudo apt install -y libnetcdff-dev mpi-default-dev
+          sudo apt install -y libnetcdff-dev mpi-default-dev ninja-build
         if: runner.os == 'Linux'
 
       - name: Install dependencies - macOS
-        run: brew install netcdf open-mpi
+        run: brew install netcdf open-mpi ninja
         env:
           HOMEBREW_NO_INSTALL_CLEANUP: 1
         if: runner.os == 'macOS'
 
-      - name: Checkout code
-        uses: actions/checkout@v2
-
       - name: Setup Python for newer version of Meson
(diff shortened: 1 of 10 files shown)


AleksiNummelin:
I think this is a good start, we can get the physical tests in later. A minor point is that some of the new idealized cases (like the single_column and channel, maybe even fuk95) could be used for real science as well, not just for testing. So I wonder if `tests` directory should be renamed to `examples` or something similar, or if there should be another directory called `examples` and then in the `tests`directory one would just have the `ci.yml` and in future some validation data/scripts related to the physical tests? This way the `tests` directory would just be a case directory for testing while the actual fortran files would live in `examples/<example_name>` directory. @nordmoen would this approach be possible/problematic in terms of CI/meson?

nordmoen:
It doesn't really matter what the `test` folder is called as long as it is added with `subdir` in the main `meson.build` files. Currently the `test` folder only contains `meson.build` files to create an executable which is then added as a Meson test, so no Fortran files have been moved there.

The `ci.yml` file must live under `.github/workflow` folder or else it will not work.

I think we maybe should add an additional `example/` folder instead of using the `test/` folder. The unit tests, that should be run by CI, should not be changed by users so that we can always be certain that the tests are equivalent on all machines.

matsbn:
I tried the test procedure on Betzy and it worked well! We have an ongoing discussion on the directory structure of the repository (https://github.com/NorESMhub/BLOM/issues/86), but it seems fairly straightforward to relocate things when that has been settled.

One thing I would have preferred is that the output from tests is relative to the "builddir" given as argument to Meson. I typically locate the "builddir" outside of the repository and writes output on my scratch area. The updated .gitignores ensures that the test output does not interfere with git, but cleaning up after a test is a bit cumbersome since the "limits" file should be preserved in the test run folder.

nordmoen:
> One thing I would have preferred is that the output from tests is relative to the "builddir" given as argument to Meson. I typically locate the "builddir" outside of the repository and writes output on my scratch area. The updated .gitignores ensures that the test output does not interfere with git, but cleaning up after a test is a bit cumbersome since the "limits" file should be preserved in the test run folder.

I change the configuration around so that this is indeed the case.

matsbn:
Thanks for making that change, @nordmoen!

PR #92, opened 2021-05-10
Bugfix cisonew

### LINKED ISSUE #90: Combine CPP flags with module "use <module>, only:"
@JorgSchwinger @matsbn 

I found a bug when compiling with the CPP flag "cisonew" true. In beleg_param.F90 the variables iatmc13 and iatmc14 are not defined, because they are not loaded from the mo_param1_bgc module. This module is loaded with 
use mo_param1_bgc,  only: iatmco2,iatmnco2,iatmo2,iatmn2

The problem is that iatmc13 and iatmc14 are only defined when the "cisonew" flag is true, so it is not trivial to add these variables to the list that is loaded from the module. 
- Solution 1: remove "only" condition when using the module
- Solution 2: have different variants of the module load  directive, depending on CPP flags
- Solution 3: Remove the CPP flag settings, find some other method to turn these effect on and of

For now I suggest we use solution 1 in these cases, but maybe we should think about solution 3 for the long term?

TomasTorsvik:
Some bugs appeared when I set the 'cisonew' CPP flag TRUE. These bug fixes makes it possible to compile the code.

I have not done a test run with this code yet. Do we have a testcase that I can compare with?

### DIFF
--- a/hamocc/beleg_vars.F90
+++ b/hamocc/beleg_vars.F90
@@ -40,7 +40,7 @@
       use mo_vgrid, only: kmle,kbo
       USE mod_xc,   only: mnproc
 
-      implicit none      
+      implicit none
 
       INTEGER, intent(in) :: kpaufr,kpie,kpje,kpke,kbnd
       REAL,    intent(in) :: pddpo(kpie,kpje,kpke)
@@ -51,6 +51,9 @@
 
       ! local variables
       INTEGER :: i,j,k,l
+#ifdef cisonew
+      REAL :: rco213,rco214,beta13,beta14
+#endif
 
 
 #ifdef FB_BGC_OCE
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -1749,8 +1749,8 @@ inisil_file = `echo $INISIL | tr -d '"' | tr -d "'"`
 EOF
   if ($HAMOCC_CISO == TRUE) then
 cat >> $CASEBUILD/blom.input_data_list << EOF
-inic13_file = `echo $INIC13 | tr -d '"' | tr -d "'"`
-inic14_file = `echo $INIC14 | tr -d '"' | tr -d "'"`
+inic13_file = `echo $INID13C | tr -d '"' | tr -d "'"`
+inic14_file = `echo $INID14C | tr -d '"' | tr -d "'"`
 EOF
   endif
   if ($RUN_TYPE == startup) then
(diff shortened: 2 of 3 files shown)


hamocc/beleg_vars.F90
```diff
@@ -51,6 +51,9 @@ SUBROUTINE BELEG_VARS(kpaufr,kpie,kpje,kpke,kbnd,pddpo,prho,omask,      &
 
       ! local variables
       INTEGER :: i,j,k,l
+#ifdef cisonew
+      REAL :: rco213,rco214,beta13,beta14
```
JorgSchwinger, line 55: This bug was introduced when splitting beleg.F90 into beleg_vars.F90 and beleg_parm.F90. The variables rco213,rco214 are only used in the latter, but are still defined in the former. So they could be removed from beleg_parm.F90.

JorgSchwinger (changes requested):
Would be great if you could also remove rco213 and rco214 from beleg_parm.F90, where they are actually not used. Otherwise it looks fine.

JorgSchwinger:
Regarding a test case, if the model is still bit-identical with the CMIP6 version, you could compare with this output 

/projects/NS2980K/schwinger/NOINYOC_T62_tn21_27_betzy

This was produce using 
create_newcase --case NOINYOC_T62_tn21_27_betzy --res T62_tn21 --mach betzy --compset NOINYOC --project nn9560k

JorgSchwinger (approved):

TomasTorsvik:
@JorgSchwinger 
NOINYOC_T62_tn21 with default settings is unchanged from the reference. When I set HAMOCC_CISO=TRUE I get an error message during build:
INIC13: Undefined variable

I guess this has something to do with missing input files? Should I merge this anyway, or is there a simple fix to get the test running?

JorgSchwinger:
Hi @TomasTorsvik 

it is simple to fix (it is a typo in the buildnml script in cime_config):

at the very end in the two lines

inic13_file = `echo $INIC13 | tr -d '"' | tr -d "'"`
inic14_file = `echo $INIC14 | tr -d '"' | tr -d "'"`


INIC13 and INIC14 have to be replaced by INID13C and INID14C (INIC13 and INIC14 are indeed never defined). Would be great if you could add this to this commit.

TomasTorsvik:
Fixed namelist variables. Now the model runs with HAMOCC_CISO=TRUE.

PR #95, opened 2021-06-04
Add CONTRIBUTING.md file.

TomasTorsvik:
The CONTRIBUTING.md file is partly based on the one for NorESM, but some changes were necessary. I have indicated that there is info about the BLOM repository structure in the wiki, although this is not presently the case. However, I think this would be nice to have.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,6 +18,9 @@ Please consider providing additional information with your bug report, e.g.
 - Error messages, log output
 
 ## Suggesting new features or feature enhancements
+An issue can also be raised to make suggestions for new features or feature enhancements, or to get in touch with members of the core BLOM/iHAMOCC development team regarding more general questions about the model system. If a suggestion is approved by the development team, we might also provide suggestions or limited assistance for the implementation of new solutions.
+
+The basic steps for adding code to the BLOM/iHAMOCC model is as follows:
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,25 @@
...
+## Reporting bugs
+If you find bugs, strange output or something puzzling, please raise an issue. No issue is too big nor too small. Before creating a new issue, we would be grateful if you first check existing issues in case the bug has already been reported. In this case you may want to add your comments to the existing bug report.
+
+Please consider providing additional information with your bug report, e.g.
+- model version for BLOM/iHAMOCC and/or NorESM
+- HPC platform and environment settings
+- Model settings (compset, grid resolution, etc.)
+- Error messages, log output
+
+## Suggesting new features or feature enhancements
```
JorgSchwinger, line 20: I wonder whether we should add something about discussing new features with the relevant members of the core BLOM/iHAMOCC group? If someone wishes to implement a new feature this has to be discussed beforehand to ensure that we actually want to include this feature and to provide some guidance on how to implement this best.

JorgSchwinger (approved):
Thanks Tomas, this looks good to me, I just had one small comment that we might consider (but I approve this anyway).

CONTRIBUTING.md
TomasTorsvik: Yes, I agree. I will add something about this.

PR #98, opened 2021-06-11
Update readme

### LINKED ISSUE #97: Document stand alone BLOM/iHAMOCC in wiki?
@matsbn @nordmoen @AleksiNummelin @JorgSchwinger @milicak 
I suggest that we can use the wiki to document building and running stand alone BLOM/iHAMOCC. This is now in the README.md file, but that file is not really suitable for a long documentation text. The wiki is a better option for this because it can be structured over several pages for different sub-topics. At the moment I have just copied over the meson info from the README.md file. I was also thinking to include some description about the different test cases. What do you think, is this a way to go or do you prefer the current format with info in the README.md file?

TomasTorsvik:
Here is a suggestion for an update of the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ $ meson test -C builddir "run single_column"
 ## Contribute to BLOM/iHAMOCC development
 
 The [CONTRIBUTING.md](CONTRIBUTING.md) file includes instructions on how to contribute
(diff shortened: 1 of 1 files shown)


nordmoen (approved):
LGTM

JorgSchwinger (approved):

PR #101, opened 2021-06-18
Replace type REAL with LOGICAL for bgcwrt.

TomasTorsvik:
Hope it's OK that I make some minor adjustments like this.


JorgSchwinger (approved):
Looks good, sorry that it took so long before I looked into this!

PR #102, opened 2021-06-24
Update copyright and comment sections for various iHAMOCC files.

TomasTorsvik:
Mostly trivial updates. Only one question about reference to IAUFR in hamocc4bcm (see comment dated 2021-06-18).

### DIFF
--- a/hamocc/hamocc4bcm.F90
+++ b/hamocc/hamocc4bcm.F90
@@ -124,13 +124,15 @@
 
 
 !--------------------------------------------------------------------
-! Increment bgc time step counter of run (initialized in INI_BGC).
+! Increment bgc time step counter of run (initialized in HAMOCC_INIT).
 !
       ldtrunbgc = ldtrunbgc + 1
 
 
 !--------------------------------------------------------------------
 ! Increment bgc time step counter of experiment (initialized if IAUFR=0).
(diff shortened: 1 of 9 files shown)


hamocc/hamocc4bcm.F90
```diff
@@ -124,13 +124,15 @@ SUBROUTINE HAMOCC4BCM(kpie,kpje,kpke,kbnd,kplyear,kplmon,kplday,kldtday,&
...
-! Increment bgc time step counter of run (initialized in INI_BGC).
+! Increment bgc time step counter of run (initialized in HAMOCC_INIT).
 !
       ldtrunbgc = ldtrunbgc + 1
 
 
 !--------------------------------------------------------------------
 ! Increment bgc time step counter of experiment (initialized if IAUFR=0).
+! 2021-06-18 : IAUFR does not seem to be defined. ldtbgc is defined in
+!              mo_control_bgc and initialized in hamocc_init.
```
JorgSchwinger: Maybe we should rather remove the text "(initialized if IAUFR=0)" instead of adding a comment that the comment seems to be wrong?

JorgSchwinger (approved):

hamocc/hamocc4bcm.F90
TomasTorsvik: Agree. I just wanted to confirm that was actually incorrect. I will remove the text.

PR #106, opened 2021-08-20
Change reading of BGC namelist

### LINKED ISSUE #99: Extend the fuk95 with option to run iHAMOCC?
@matsbn @JorgSchwinger 
It would be nice to have a stand-alone test case that can run iHAMOCC, so I was wondering if it would be possible to extend the fuk95 case to allow for this. There are some tracers that are initialized with GLODAP and WOA data in beleg_vars.F90. Are there any other obstacles that need to be handled?
I was thinking to just extend the existing mod_fuk95 with an additional subroutine, e.g. inihamocc_fuk95, to set the necessary variables. What do you think about this?

TomasTorsvik:
Change reading of BGC namelist to accept (1) ocn_in file or (2) limits file.

- Use the "newunit" option when opening namelist file, do not reserve io_nml unit for namelist.
- New subroutine in mo_control_bgc : get_bgc_namelist

This change will be convenient if implementing any stand-alone test cases with activated HAMOCC component using "limits" file.

### DIFF
--- a/hamocc/mo_control_bgc.F90
+++ b/hamocc/mo_control_bgc.F90
@@ -40,7 +40,9 @@
 ! Logical unit number for I/O.
 
       INTEGER :: io_stdo_bgc           !  standard out.
-      INTEGER :: io_nml                !  namelist
+
+! File containing namelists
+      CHARACTER(LEN=:), ALLOCATABLE, PROTECTED :: bgc_namelist
 
 ! Control variables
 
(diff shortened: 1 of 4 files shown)


hamocc/mo_control_bgc.F90
```diff
@@ -60,5 +62,40 @@ MODULE mo_control_bgc
...
+        implicit none
+
+        character(len=*), intent(out) :: fname
+        logical :: exists
+
+        if (.not. allocated(bgc_namelist)) then
+           inquire (file='ocn_in'//trim(inst_suffix), exist=exists)
+           if (exists) then
+              allocate(character(len=len('ocn_in'//trim(inst_suffix))) ::       &
+                   bgc_namelist)
```
JorgSchwinger, line 83: I haven't seen this syntax before, I would write
`allocate(bgc_namelist(len(...))`
This is how it is used other places in iHAMOCC

JorgSchwinger (commented):
Looks good to me, just one minor comment below

hamocc/mo_control_bgc.F90
TomasTorsvik: Agree, I will change this.

TomasTorsvik: Turns out that the standard syntax returns an error in this case:

/cluster/projects/nn2980k/tomast/NORESM/noresm-dev/components/blom/hamocc/mo_control_bgc.F90(83): error #8232: If any object being allocated in the ALLOCATE statement has deferred type parameter, either type specification or SOURCE= or MOLD= specifiers shall appear.   [BGC_NAMELIST]

/cluster/projects/nn2980k/tomast/NORESM/noresm-dev/components/blom/hamocc/mo_control_bgc.F90(83): error #8196: An ALLOCATE shape must not be given for a scalar object.   [BGC_NAMELIST]

The reason is that bgc_namelist is a scalar character variable, not a character array. Still, I would like to use the test (.not. allocated(bgc_namelist)) to check if the variable is defined, so that's the reason for making it allocatable in the first place. It works, but it requires this strange syntax, so I'm changing it back again.

JorgSchwinger: If the main purpose is just the test, then one could also initialise `bgc_namelist=''` and use `len_trim(bgc_namelist)/=0` to test whether it has been assigned a value. I'm just not sure whether this syntax is according to a fortran standard. I believe it is not conforming to f90/95, but might be introduced later(?).

TomasTorsvik: Allocatable scalars and character length is part of the F2003 standard (see e.g. https://gcc.gnu.org/wiki/Fortran2003Status), and is supported by gcc from version 4.5 or 4.6, so I think it should be safe to use by now. As you say, we could also use the `len_trim` test for the same effect. 

Perhaps a better solution all together would be to use the `bgc_namelist` variable from `mo_control_bgc` directly, and only invoke `get_bgc_namelist` once. Then we don't need to define fixed length variables in the calling subroutines, the namelist file name could be of arbitrary length, and we would not need to use `trim(bgc_namelist)` to remove trailing whitespace.

TomasTorsvik:
I'm merging this now. I haven't seen problems with the allocated character scalar in the tests, so I think this should be safe to use.

PR #109, opened 2021-08-26
Extend mo intfcblom

### LINKED ISSUE #108: Change blom2hamocc.F and hamocc2blom.F to free source format
@JorgSchwinger @matsbn 
From mo_intfcblom.F90 header:
!   Once BLOM has transitioned to free source format, the interface routines
!   blom2hamocc and hamocc2blom can be incorporated into this module

I think this can be done now. Any reason why not?

TomasTorsvik:
Include blom2hamocc and hamocc2blom subroutines in mo_intfcblom. Ref. issue #108 .


JorgSchwinger (approved):
Hi Tomas

thank you looks good. I assume you tested that you get bit-identical results?

TomasTorsvik:
I compared with 1 year run of NOINYOC_f19_tn21 noresm2.0.5 and confirmed bit-exact output, so it should be OK. There are not much actual changes in the code, just moving it into a module and re-ordering some j- and k-loops, so it looks more extensive than it really is.

PR #112, opened 2021-09-01
Import `nocetra` variable from mo_param1_bgc in mod_tracers.

### LINKED ISSUE #105: Enable tracer count to be determined at run time
For discussion:
If we should have a BLOM/iHAMOCC model that can accommodate different BGC tracer counts at run time, it is necessary to first make some changes in the BLOM/iHAMOCC interface, and the way that the tracer module works. I propose the following changes:

- make "ntr" adjustable at runtime through a new subroutine ("set_trc") and change the attribute from PARAMETER to PROTECTED.
- Make the tracer arrays allocatable: trc, trcold, uflxtr, vflxtr, trflx
- Make a new "allocate_tracers" subroutine
- Set value of "ntrbgc" based on variable settings in input namelist (e.g. "with_cfc", "with_natDIC", "with_AGG", "with_ciso" as .true. or .false.)
  - advantage: No need for "double booking" of i_base, i_iso, i_cfc, i_agg, i_nat_dic 

This setup will require the following procedure:
1. Read namelist for iHAMOCC tracer variables
2. Set "ntrbgc"
3. Set "ntr"
4. Allocate tracer arrays

After that the model structure should be the same as before, so I don't think there is a need for major restructuring of the code. However, the tracer allocation will have to be included somewhere in the initialization procedure.

TomasTorsvik:


JorgSchwinger (approved):
This is very good, excellent! Much easier this way.

PR #113, opened 2021-09-14
Allocatable tracer array

### LINKED ISSUE #105: Enable tracer count to be determined at run time

TomasTorsvik:
This is a suggestion for changing the tracer array from static to allocatable. I have also moved some things around in the module, but this doesn't make any difference for the final outcome.

### DIFF
--- a/trc/mod_tracers.F90
+++ b/trc/mod_tracers.F90
@@ -88,8 +88,7 @@ contains
    ! Set total number of tracers : ntr
    ! Allocate tracer arrays.
    ! ---------------------------------------------------------------------------
-     use dimensions, only: idm, jdm, kdm
-     use mod_xc,     only: mnproc, nbdy, lp, xchalt, xcbcst
+
 #ifdef HAMOCC
      use mo_param1_bgc, only: nocetra
 #endif
--- a/trc/mod_tracers.F90
+++ b/trc/mod_tracers.F90
@@ -78,7 +78,7 @@ module mod_tracers
    public :: ntrocn, ntrtke, ntrgls, ntriag, ntrbgc, ntr, natr,                 &
              itrtke, itrgls, itriag, itrbgc,                                    &
              trc, trcold, uflxtr, vflxtr, trflx,                                &
-             allocate_tracers, inivar_tracers
+             inivar_tracers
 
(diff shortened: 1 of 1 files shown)


TomasTorsvik:
From https://github.com/NorESMhub/BLOM/issues/105#issuecomment-919024257 :

Here are results from some tests I did on Betzy and Fram:
[alloc_array_perf.txt](https://github.com/NorESMhub/BLOM/files/7161071/alloc_array_perf.txt)

It seems that there still can be some issues when replacing static arrays with allocatable ones, but perhaps only for specific compiler settings or case parameters. For most tests the differences were fairly small.

JorgSchwinger:
I don't know why, but github won't let me add an review. 

So I just coment here: looks good to me, I would be fine with including this.

JorgSchwinger (approved):
ok, now it works...

TomasTorsvik:
It turns out that the code runs with ``call xcbcst(ntr)`` even if ``ntr`` is declared ``protected``, so I removed the ``num_tracers`` variable. I also ran the fuk95 case without problems from the branch.

PR #114, opened 2021-09-14
CVMix added as a submodule. mod_difest.F is changed accordingly

milicak:
This is the initial commit of the CVMix as a submodule.
The changes are significant and further testings are required.


matsbn:
Mehmet, can you check your committed mod_cmnfld.F90? Git diff indicates the whole file is changed, while it seems much of the code is unchanged. I guess the file is saved with different end-of-line convention or something. If this is the case, it should be corrected so the diff is meaningful.

TomasTorsvik:
Note also error message from the automatic build

pkgs/meson.build:1:0: ERROR: File CVMix-src/src/shared/cvmix_background.F90 does not exist.

Looks like something should be fixed for the meson build process.

milicak:
Hopefully it should work now. But instead of git clone, it should be used as git clone --recursive From: Tomas Torsvik ***@***.***>Date: Wednesday, September 15, 2021 at 9:48 AMTo: NorESMhub/BLOM ***@***.***>Cc: Mehmet Ilicak ***@***.***>, Author ***@***.***>Subject: Re: [NorESMhub/BLOM] CVMix added as a submodule. mod_difest.F is changed accordingly (#114)Note also error message from the automatic buildpkgs/meson.build:1:0: ERROR: File CVMix-src/src/shared/cvmix_background.F90 does not exist.Looks like something should be fixed for the meson build process.—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android.

matsbn (approved):
I suggest that we fix the failing checks later and merge this PR. I would also recommend "squash and merge" so that the large, but mostly opposite, diffs in mod_cmnfld.F90 in two commits will cancel out.

TomasTorsvik:
> I suggest that we fix the failing checks later and merge this PR. I would also recommend "squash and merge" so that the large, but mostly opposite, diffs in mod_cmnfld.F90 in two commits will cancel out.

That's fine with me, we can probably be a bit less strict with feature branches.

PR #120, opened 2021-09-29
Bottom Brunt-Vaisala frequency computation added. 

milicak:
Bottom Brunt-Vaisala frequency computation added. CVMIX CPP is removed. New method to compute interface heights has been implemented.


matsbn (commented):
Looks good. I just wonder about the purpose of "hcorr" in mod_difest.F. It is set to zero and added to "dh" which seems to be inconsistent with the comment "Take away the accumulated error (could temporarily make dh<0)".

matsbn (approved):

PR #129, opened 2021-11-03
Add pes layout for f19_tn14 resolution in cplhist mode

JorgSchwinger:
This fix makes the existing CPLHIST-compset work for the f19 resolution

### DIFF
--- a/cime_config/config_pes.xml
+++ b/cime_config/config_pes.xml
@@ -187,7 +187,7 @@
       </pes>
     </mach>
   </grid>
-  <grid name="a%0.9x1.25.+oi%tnx1v4">
+  <grid name="a%0.9x1.25.+oi%tnx1v4|a%1.9x2.5.+oi%tnx1v4">
     <mach name="any">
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):
Looks fine to me.

matsbn (approved):
Looks fine.

cime_config/config_pes.xml
```diff
@@ -187,7 +187,7 @@
       </pes>
     </mach>
   </grid>
-  <grid name="a%0.9x1.25.+oi%tnx1v4">
+  <grid name="a%0.9x1.25.+oi%tnx1v4|a%1.9x2.5.+oi%tnx1v4">
```
TomasTorsvik, line 190:
```suggestion
  <grid name="a%0.9x1.25.+oi%tnx1v4 | a%1.9x2.5.+oi%tnx1v4">
```
Question: Will it also work if you include spaces around the pipe symbol? Maybe improves readability.

JorgSchwinger: Not sure, there is no space other places, so I'd suggest to leave it for now.

PR #134, opened 2021-11-09
Fixed accessing arrays beyond sensible range and initialized some variables.

matsbn:
Some small bug fixes. Does not change results (tested compset NOIIAOC), but needed to avoid exceptions during thorough debugging.


TomasTorsvik (approved):
Looks fine to me.

PR #135, opened 2021-11-10
Do not add -DLEVITUS2X option when building with tnx2v1 grid

JorgSchwinger:
This restores the behavior from before the updated build-scripts. With the tn21-grid, we have always used only 35 vertical z-levels in the output. This saves storage (particularly for HAMOCC) and is good enough for the lower resolution ocean grid, so I think we should stick to it.


matsbn (commented):
There was no intention to change behaviour in this respect, so great you found this. But please exclude "buildcppc" in the PR. This is compiled python code of "buildcpp" and is system dependent.

JorgSchwinger:
Ok, I have removed the buildcppc file. I will then proceed and merge into master. This should probably also go into the new release branch before making the release tag

matsbn (approved):

PR #138, opened 2021-12-13
Add mo_clim_swa.F90 to meson.build file.

TomasTorsvik:
mo_clim_swa.F90 was missing in the meson.build file.


matsbn (approved):
Looks good. Great that you are testing iHAMOCC code with standalone build system.

PR #143, opened 2022-02-07
Refactor inventory bgc

TomasTorsvik:
Include option to write iHAMOCC inventory to netCDF file.

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -1383,9 +1383,12 @@ cat >>! $RUNDIR/ocn_in$inststr << EOF
 !   COMPFLAG - switch for compressed/uncompressed output (i) 
 !   NCFORMAT - netcdf format (valid arguments are 0 for classic,
 !              1 for 64-bit offset and 2 for netcdf4/hdf5 format)
-!   INVENTORY- how often to write an inventory of tracers to ocean 
-!              output file
-!  
+!   INVENTORY- how often to write an inventory of tracers to ocean
+!              output file:
+!              0 : no output
+!              1 : write inventory to standard output (log file)
+!              2 : write output to inventory netCDF file
+!
 ! Arguments for diagnostic variables:
 !   0        - variable is not written
 !   2        - variable is written as int2 with scale factor and offset
--- a/hamocc/accfields.F90
+++ b/hamocc/accfields.F90
@@ -53,7 +53,6 @@
       use mo_vgrid, only: dp_min
       use mod_xc
       use mo_riverinpt, only: irdin,irdip,irsi,iralk,iriron,irdoc,irdet,rivinflx
-      use mod_config, only: expcnf
       implicit none
       INTEGER :: kpie,kpje,kpke
       REAL    :: pdlxp(kpie,kpje)
@@ -385,13 +384,6 @@
         endif
       ENDDO
 
-      IF(expcnf.eq.'single_column') THEN
-        IF (mnproc.eq.1) THEN
-         WRITE(io_stdo_bgc,*)' '
-         WRITE(io_stdo_bgc,*)'after BGC flux accumulation: call INVENTORY'
-        ENDIF
-        CALL INVENTORY_BGC(kpie,kpje,kpke,pdlxp,pdlyp,pddpo,omask,0)
-      ENDIF
       atmflx=0. ! nullifying atm flux here to have zero fluxes for stepwise inventory fluxes
       ndepflx=0.
       rivinflx=0.
--- a/hamocc/accfields.F90
+++ b/hamocc/accfields.F90
@@ -44,12 +44,12 @@
 !     *REAL*    *omask*   - land/ocean mask
 !
 !**********************************************************************
-      USE mo_carbch
-      USE mo_sedmnt
       USE mo_biomod
       USE mo_bgcmean
+      USE mo_carbch
       USE mo_control_bgc
-      use mo_param1_bgc 
+      use mo_param1_bgc
+      USE mo_sedmnt
       use mo_vgrid, only: dp_min
       use mod_xc
       use mo_riverinpt, only: irdin,irdip,irsi,iralk,iriron,irdoc,irdet,rivinflx
@@ -378,7 +378,7 @@
         nacc_bgc(l)=nacc_bgc(l)+1
         if (bgcwrt(l)) then
           if (GLB_INVENTORY(l).ne.0) then
-            CALL INVENTORY_BGC(kpie,kpje,kpke,pdlxp,pdlyp,pddpo,omask,0)
+            CALL INVENTORY_BGC(kpie,kpje,kpke,pdlxp,pdlyp,pddpo,omask,l)
           endif
           call ncwrt_bgc(l)
           nacc_bgc(l)=0
--- a/hamocc/accfields.F90
+++ b/hamocc/accfields.F90
@@ -52,7 +52,8 @@
       use mo_param1_bgc 
       use mo_vgrid, only: dp_min
       use mod_xc
-
+      use mo_riverinpt, only: irdin,irdip,irsi,iralk,iriron,irdoc,irdet,rivinflx
+      use mod_config, only: expcnf
       implicit none
       INTEGER :: kpie,kpje,kpke
       REAL    :: pdlxp(kpie,kpje)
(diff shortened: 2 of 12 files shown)


hamocc/mo_inventory_bgc.F90
```diff
@@ -0,0 +1,883 @@
...
+!**********************************************************************
+
+  implicit none
+
+  integer, intent(in) :: kpie,kpje,kpke
+  integer, intent(in) :: iogrp
+  real,    intent(in) :: dlxp(kpie,kpje)
+  real,    intent(in) :: dlyp(kpie,kpje)
+  real,    intent(in) :: ddpo(kpie,kpje,kpke)
+  real,    intent(in) :: WETO(kpie,kpje)
```
JorgSchwinger: we could rename WETO to omask for consistency with other routines

hamocc/mo_inventory_bgc.F90
```diff
@@ -0,0 +1,883 @@
...
+
+subroutine sum2d(kpie,kpje,dlxp,dlyp,var2d,total,mask)
+!**********************************************************************
+!**** Sum 2D scalar fields
+!**********************************************************************
+  implicit none
+  integer, intent(in) :: kpie, kpje
+  real, dimension(kpie,kpje), intent(in) :: dlxp, dlyp, var2d
+  real, intent(out) :: total
+  real, dimension(kpie,kpje), intent(in), optional :: mask
```
JorgSchwinger: Do we have a case where it makes sense to call this routine without "mask"? If not this could be simplified by making mask non-optional

hamocc/mo_inventory_bgc.F90
TomasTorsvik: Renamed WETO to omask.

hamocc/mo_inventory_bgc.F90
TomasTorsvik: mask is now required in sum2d routine.

JorgSchwinger (commented):
Is it necessary to have the rivinflx and ndepflx arrays? As far as I see, they only differ by dtb/356 from rivin and ndep, so this multiplication could also be done in accflds? I find it slightly confusing this way (and waste of memory - although it's not a lot)

jmaerz:
During implementation, I followed the atmflx and differentiated between input and output fields. I agree that rivinflx and ndepflx might not be needed, if you break this barrier between i/o. Being new to betzy: speaking about waste of memory - is this an issue?

JorgSchwinger:
But in contrast to atmflx, ndepflx and rivinflx are not calculated inside hamocc (except for unit conversion, i.e. multiplication with dtb/365). I think one could even apply the factor dtb/365 right away when ndep and rivin are read in, then it would be even more transparent. Regarding memory on betzy: I think iHAMOCC is already quite memory intensive with 20-30 tracers (on two time levels in the trc field, plus a copy of the current time level in octra). So in that sense the addition of ndepflx and rivflx does not really matter. It's more about avoiding confusion  about what the different fields actually are.

jmaerz:
Sorry for the late reply. Another thing that is maybe worth mentioning about the current implementation that it enables a time-step wise flux and inventory check (based on instantaneous and not accumulated fluxes), which requires the nullifying of the output fields every time step (so knowing the inventory before and after application of riverine and deposition fluxes). It's not in such way in Tomas branch thus far (but lives in my branch). It helped me a lot in the single column setup to debug the inventory calculations for iHAMOCC, independent of potential issues inside BLOM (see my emails). Hence, this would be in conflict to the river input that is only read at iHAMOCC initialization.

jmaerz:
Maybe it would be an option to rename the arrays holding the information for the output/inventory calculation in a more meaningful way to avoid confusion?

JorgSchwinger:
Hi Jöran, sorry I do not understand this point really, why is it necessary to nullify the fluxes if you do a step-wise inventory? Could you explain a bit? (Sorry I just might be a bit stupid here)

jmaerz:
The current way of calculating the mass inventory makes use of the accumulated fluxes to backtrace, if something went wrong in the calculations. The accumulation happens after the BGC routines. Since I also saw the intra-run variability caused by BLOM in a single column setup, I aimed for a possibility to track BGC inventory changes immediately after each subroutine and independent of BLOM. Since the accumulation of fluxes happens after BGC, potential errors in the inventory calculations due to riverine and N-deposition fluxes were lumped together (since no debugging and inventory calculation inbetween was done and could not be done in the current model structure - accumulation of the individual fluxes would require to be called right after applying the individual fluxes to do so - or you would have to calculate inventory changes outside of iHAMOCC). Note that we also had to check, if the inventory routines were written correctly to include the riverine and N-dep fluxes. To also distinguish from potential BLOM issues (or potential issues in unit conversions from BLOM to iHAMOCC), I only looked at the instantaneous fluxes happening in individual time steps (in my branch, I calculate the sum over the instantaneous fluxes in the inventory calculation and not the accumulated fluxes when being in single column mode) and hence, it matters again, if I check the inventory before or after fluxes have been applied. With your suggestion, one would have to check in the inventory calculation, if one is before or after the application of the fluxes to potentially apply the fluxes also in the inventory calculations. With the additional field being nullified as I did, such check isn't needed (since it only holds value different from zero, if the fluxes were applied). By this approach, it is enough to look at the inventory changes, when switching on the debugging flags to see, in which step of iHAMOCC, something goes potentially wrong.  Not sure, if this helps as explanation, though... And there might be more elegant ways to achieve the same - like introducing an additional logical switch for the inventory calculations to in/exclude certain flux fields.

JorgSchwinger:
Thanks for the explanation, Jöran, I understand the problem now (sorry, I didn't think of the sub-timestep level inventory). From a structural point of view, I still think it would be less confusing if all this is handled inside the inventory-routine by passing a flag (I would go for one integer-flag) that indicates which fluxes have already been applied (as you already mention). Then, one could use the already existing arrays to subtract the per time-step flux from the total inventory if the respective flux has already been applied.

But on the other hand, I'm not doing the work here, so please go for a solution that works and is easy to implement.

JorgSchwinger:
Thanks for the explanation, Jöran, I understand the problem now (sorry, I didn't think of the sub-timestep level inventory). From a structural point of view, I still think it would be less confusing if all this is handled inside the inventory-routine by passing a flag (I would go for one integer-flag) that indicates which fluxes have already been applied (as you already mention). Then, one could use the already existing arrays to subtract the per time-step flux from the total inventory if the respective flux has already been applied.

But on the other hand, I'm not doing the work here, so please go for a solution that works and is easy to implement.

jmaerz:
Hi Jörg, I guess, in this case, I would go for passing one flag per individual flux via use statements (each defined in the individual fluxes modules - e.g. lriverinp_done, lndep_done). Another question in this context is whether the 'single column mode individual time step check' is of a broader interest? Then we would need to add this to Tomas branch as well. If we settle on this, then I would ask Tomas how to best share the tasks (I presume that Tomas is ok with that, if not please let me know, Tomas). The programming shouldn't take too much time. I would also include the direct time step conversion while reading the files to reduce computational costs.

TomasTorsvik:
Hi @jmaerz @JorgSchwinger ,
I think it's a good idea to use the inventory as a means for debugging, with multiple sub-timstep level calls. I wasn't sure if we should bring in all code changes in this pull request. However, I do think that the "single column mode" is useful enough that it is of general interest to introduce this option in the master branch, and then it would make sense to include the required code changes right away. If we all agree on this, then I think Jöran and I should get together and work out how to share the work.

jmaerz:
Ok, I'll prepare the code as suggested.

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , I realized that for N-deposition (as for dust), the field for holding the fluxes is initialized in hamocc_step, which makes it a bit cumbersome to provide it to the inventory. Thus I would allocate and initialize it in mo_ndep (similar to the river input), which enables to use the field everywhere via a use statement.

JorgSchwinger:
Hi Jöran, just one note on the iHAMOCC structure: We invested some work into separating everything that is grid dependent from the core of iHAMOCC (with this I mean everything below and including the HAMOCC4BCM routine). I.e., everything that requires reading data from a file in a specific format has been moved to hammoc_init and hamocc_step. That is why the rivin, ndep, and dust fields are passed as arguments to HAMOCC4BCM. The reason for this is/was that we wanted to be able to run the same core HAMOCC code with other physical "drivers" with as few changes as possible. Concretely, we worked on implementing a transport matrix environment for fast running experiments (where all these fields would come from different input files). Currently, there is no one pursuing this any further, but I think the basic principle is good and should be kept. Please keep this in mind when making changes as to how these fields are handled. Maybe we can discuss this next week.

jmaerz:
Hi Jörg, so you would prefer funneling e.g. riverinput fluxes and switches from hamocc_step via the subroutine call to hamocc4bcm further into ocprod, where it would only be used for potential inventory calls? - and make accordingly changes to these subroutine calls?

jmaerz:
Hi all, thanks for the good productive discussion this morning! As discussed, I updated the `single_column_inventory` branch towards a more general debugging-oriented time step-based inventory calculation in debugging mode (preprocessor flag `PBGC_CK_TIMESTEP` - note that in the current version, the `ocprod`-related flag `PBGC_OCNP_TIMESTEP` can be switched on in addition). Additionally, I disentangled the memory allocation for the inside-`hamocc4bgm` used inventory calculation memory for the river input from the outside handled riverinput fields (leading to two calls to the river module in `hamocc_init.F`). The latter is still tentative when it comes to a full disentangling, but a step towards handling it fully disentangled and can serve as a guideline for future steps (e.g. also for N-deposition). If you agree, Tomas can merge it into his inventory branch.

TomasTorsvik:
Hi all,
I think we are getting close to a version that can be pulled into master. I included pore water and sediment budgets in the netCDF file, and changed the output routine so that it follows the `GLB_FILEFREQ` settings, and recognize the `fileann_bgc(iogrp)` and `filemon_bgc(iogrp)` settings for annual and monthly output. If needed, we can implement separate flags for inventory output, but I don't see the need at the moment.

I also merged in the `single_column_inventory` branch, and changed it to use the global sum functions. There was one problem with allocation of `rivinflx`, it was done in two places, so I commented out the call to the new subroutine `alloc_mem_riverinpt_bgm`. I think the initiation procedure can be updated later, and I would prefer to do it in a separate step since this pull request is already quite big.

JorgSchwinger (approved):

PR #154, opened 2022-04-04
Hamocc hybrid coord2

TomasTorsvik:
Suggestion for change in iHAMOCC bulk mixed layer index "kmle" from parameter value 2 to value set by OBL depth.

### DIFF
--- a/hamocc/mo_vgrid.F90
+++ b/hamocc/mo_vgrid.F90
@@ -125,7 +125,6 @@ subroutine set_vgrid(kpie,kpje,kpke,pddpo)
 !$OMP END PARALLEL DO
 
 
-  kmle(:,:) =2
   kbo(:,:)  =1
   bolay(:,:)=0.0
 
--- a/hamocc/mo_vgrid.F90
+++ b/hamocc/mo_vgrid.F90
@@ -53,10 +53,11 @@ module mo_vgrid
 !******************************************************************************
   implicit none
 
-  ! INTEGER, PARAMETER :: kmle   = 2        ! k-end index for layers that 
-  !                                         ! represent the mixed layer in BLOM
+  INTEGER, PARAMETER :: kmle_static = 2   ! k-end index for layers that
+                                          ! represent the mixed layer in BLOM.
+                                          ! Default value used for isopycnic coordinates.
   REAL,    PARAMETER :: dp_ez  = 100.0    ! depth of euphotic zone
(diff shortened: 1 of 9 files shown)


hamocc/mo_vgrid.F90
```diff
@@ -114,6 +115,7 @@ subroutine set_vgrid(kpie,kpje,kpke,pddpo)
 !$OMP END PARALLEL DO
 
 
+  kmle(:,:) =2
```
JorgSchwinger: This will overwrite the values set in blom2hamocc every time step.

TomasTorsvik: Thanks! That was left over from the first test implementation. It should not be used when importing the kOBL field from BLOM.

JorgSchwinger: It might be useful to have the possibility though to set kmle=2 for backwards-compatibility(?). Because the cvmix-routine will not (necessarily) return the index 2 if the old vertical layer structure is activated.

TomasTorsvik:
Perhaps there should be some sanity check for `kmle`? Is is best to do this in `blom2hamocc` when copying data from `kOBL`, or in subroutines where `kmle` is actually used?

hamocc/mo_vgrid.F90
TomasTorsvik: I Agree. I can set this in blom2hamocc with a dependency on the value of `vcoord_type_tag`.

JorgSchwinger:
I think if a sanity check would be needed, this should be done in BLOM (and I guess that cvmix already takes care of this).

TomasTorsvik:
This PR may also be postponed until after the hybrid coordinate framework has been merged into the master branch.

matsbn:
There seems to be some conflicts after I merged #170 into feature-hybrid_coord2.

TomasTorsvik:
This branch needs the fixes from #172 . I could cherry-pick them into this branch, but that will become messy when we want to merge with `master` later on. I think the cleaner way is to update `master` with #172, merge `master` into `feature-hybrid_coord2`, and then rebase the `hamocc-hybrid_coord2` branch on top of the updated `feature-hybrid_coord2` branch.

TomasTorsvik:
@matsbn 
I have updated the `hamocc-hybrid_coord2` branch with the latest changes in `master` and rebased my local branch on top of `feature-hybrid_coord2` branch. This should eliminate conflicts in the code. The code compiles with meson with enabled ecosys flag.

PR #155, opened 2022-04-06
bugfix inventory: Allow zero index iogrp for debugging.

TomasTorsvik:


TomasTorsvik:
@jmaerz @JorgSchwinger 
This should solve the problem with iogrp==0 when using the inventory routine for debugging.

jmaerz:
Hi Tomas, thanks for taking care. Yes, this should solve the issue.

Issue #148, opened 2022-03-16
iHAMOCC: explicit use statements in modules / subroutines


jmaerz:
### Explicit use statements in iHAMOCC modules/subroutines

For code readability and safety, the inclusion of explicit use statements throughout the iHAMOCC code would be nice to achieve. A first branch with all necessary coding is done (`expl_use_statem`) and tested to the point that i) all iHAMOCC preprocessor-flags (but `BOXATM`, see below) were compiled and run in single_column mode (partially with dummy input files) and ii) a regular run case for a 1yr historical simulation produces (bit) identical results in NorESM (comparing NorESM version 2.0.5, but with i) BLOM master and ii) branch `expl_use_statem`). 

Before opening a pull request, the updated inventory calculations -now also accounting for river and nitrogen deposition fluxes (#143)-   could be merged in (or the present inventory calculation changes in `expl_use_statem` need to be reverted back to the master branch state). 

**Observations on the way:**
* `mo_boxatm.F90` is missing so that testing of flag BOXATM wasn't possible 
* in a number of subroutines, the explicit declaration of `intent(in/inout)` is still missing - debating, whether we should change it in this process as well
* float numbers could be attributed with an explicit working precision, e.g.`9.1_wp` statement (future task? - mo_kind.F90?)

**Potential impact on merging other branches:**
Explicit use statements might need to be enhanced, if new parameters, diagnostics, etc. are handled in the modules/subroutines. At present state, one file experienced renaming and transformation to a module (`get_pi_ph.F90` -> `mo_get_pi_ph.F90`).

PR #157, opened 2022-04-07
Expl use statem

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , I herewith carry out a pull request for issue #148 for iHAMOCC code. Explicit use statements are now fully implemented throughout the iHAMOCC code in order to enhance traceability. I hope it will serve us well for future coding and implementations.


TomasTorsvik:
Looks good to me. I see that there is one file name change (`get_pi_ph.F90` -> `mo_get_pi_ph.F90`) and a few minor changes that are not connected with the use statements, but nothing that should change model output.

I think a standard merge commit should be used to merge this to `master`, in particular to preserve the history of the file name change.

jmaerz:
Hi Tomas, I could re-check again, but before merging the branch with the master (after setting up all the various use statements), I did a check and the results were bit identical on betzy as mentioned in #148.

JorgSchwinger (approved):
This looks really good, I think this is the best solution. Sorry for the many suggestions that didn't work out (and thanks for trying them).

PR #158, opened 2022-05-04
Remove reference to unused cpp flag RESTART_BGC.

### LINKED ISSUE #103: Replace some preprocessor flags with namelist variables in iHAMOCC
I would like to replace some preprocessor flags with namelist variables for iHAMOCC. At the moment preprocessor flags are used to turn on/off some basic functionality (e.g. cfc, natDIC, carbon isotopes), that I think would be better handled by a namelist variable. This will make it possible to change the model behavior without recompiling and creating a new executable. To start with, I want to replace FB_BGC_OCE, CFC, natDIC and CISONEW. 

We have a plan to expand the nitrogen cycle representation in iHAMOCC, and perhaps we would like to include other tracers later on. I believe it will become increasingly difficult to maintain the code if we continue adding preprocessor flags, so I would like to set a new structure before adding more tracers.

TomasTorsvik:
The CPP flag RESTART_BGC is referenced but not used anywhere in BLOM/iHAMOCC. I see that this flag is used in CICE, so perhaps it is a remnant from an earlier micom/blom version with integrated CICE? I suggest to remove the remaining references to this CPP flag.

### DIFF
--- a/cime_config/buildcpp
+++ b/cime_config/buildcpp
@@ -121,7 +121,7 @@ def buildcpp(case):
             if module == "iage":
                 blom_cppdefs = blom_cppdefs + " -DIDLAGE"
             elif module == "ecosys":
-                blom_cppdefs = blom_cppdefs + " -DHAMOCC -DRESTART_BGC -DWLIN"
+                blom_cppdefs = blom_cppdefs + " -DHAMOCC -DWLIN"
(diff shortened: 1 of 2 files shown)


JorgSchwinger (approved):
I agree, this is not used anymore and should be removed

PR #160, opened 2022-05-05
Cpp replace dmsph

### LINKED ISSUE #103: Replace some preprocessor flags with namelist variables in iHAMOCC

### LINKED ISSUE #159: DMSPH preprocessor flag and use of pi_ph
@tjiputra @JorgSchwinger 
We have a CPP flag ``DMSPH`` that determines if we calculate the field ``pi_ph``. This is used in ``ocprod`` to calculate [``dms_ph``](https://github.com/NorESMhub/BLOM/blob/82cc25e9921d791d2d634961291d6ee24bc29c8c/hamocc/ocprod.F90#L455-L458) which in turn is used to calculate ``dmsprod``. At the moment ``pi_ph`` is not used and we set ``dms_ph = 1.``, so everything related to the ``DMSPH`` CPP flag is bypassed.

Should we maintain the option to set ``dms_ph`` from ``pi_ph``? If so, can we replace the ``DMSPH`` CPP flag with a logical option that we can read from a namelist file (.e.g. ``with_dmsph``) and make an if statement to select how to set ``dms_ph``?

TomasTorsvik:
I have made a version that replaces the cpp flag `DMSPH` with the logical variable `with_dmsph`. In the process I have renamed the module `mo_get_pi_ph` to `mo_dmsph` and moved most of the DMSPH-related things into this module, so the only dependencies are in `hamocc_init.F` and in `ocprod.F90`.

The original line for calculating `dms_ph` in `oceprod` read:
`!        dms_ph  = 1+(-log10(hi(i,j,1))-pi_ph(i,j,kplmon))*dms_gamma`
The field `hi` was not defined, but I assume this is imported from `mo_carbch`. Also, the index `kplmon` is not defined in `oceprod`. Looking at other parts of the code, I see that `kplmon` is used as a dummy variable for `date%month`, so I assume that was the intention.

In `mo_dmsph` I import the `date` variable directly from the `mod_time` module, which is somewhat inconsistent with the idea of separating BLOM and iHAMOCC code, but the alternative to send `kplmon` as a dummy variable via `oceprod` doesn't look very attractive. I'm not sure what is the best solution here.

Finally, `pi_ph` is allocated with the standard `kpie,kpje,kpke` dimensions, but presumably the 3rd dimension should be number of months. Does it make sense to change this allocation?


hamocc/hamocc_init.F
```diff
@@ -181,9 +177,10 @@ subroutine hamocc_init(read_rest,rstfnm_hamocc)
 #endif
 
       CALL INI_RIVERINPT(idm,jdm,omask)
-#ifdef DMSPH
-      CALL GET_PI_PH(idm,jdm,kdm,omask)
-#endif
+
+      if(with_dmsph) then
+         CALL GET_PI_PH(idm,jdm,kdm,omask)
```
tjiputra: In the original implementation, we are reading monthly surface PI pH, so would it be better to modify this to:

CALL GET_PI_PH(idm,jdm,12,omask)

hamocc/mo_dmsph.F90
```diff
@@ -0,0 +1,141 @@
...
+    INTEGER ::i,j,l
+  
+    REAL,intent(in) ::omask(kpie,kpje)
+
+    ! define the fields
+    REAL :: pi_ph_in(kpie,kpje,12)
+    INTEGER ncid,ncstat
+
+    ! allocate pi_ph field
+    if(.not. allocated(pi_ph)) call alloc_pi_ph(kpie,kpje,kpke)
```
tjiputra: if(.not. allocated(pi_ph)) call alloc_pi_ph(kpie,kpje,12)

hamocc/mo_dmsph.F90
```diff
@@ -0,0 +1,141 @@
...
+    
+    implicit none
+    integer, intent(in) :: kpie,kpje,kpke
+    integer             :: errstat
+
+    IF (mnproc.eq.1) THEN
+       WRITE(io_stdo_bgc,*)'Memory allocation for variable pi_ph ...'
+       WRITE(io_stdo_bgc,*)'First dimension    : ',kpie
+       WRITE(io_stdo_bgc,*)'Second dimension   : ',kpje
+       WRITE(io_stdo_bgc,*)'Third dimension    : ',kpke
```
tjiputra: WRITE(io_stdo_bgc,*)'Third dimension    : ',12

tjiputra (commented):
As the original implementation, we only read climatology monthly surface pH, so the dimension should be revised to (kpie,kpje,12). 

Later, we should probably modify the dimension to (kpie,kpje,kpke,12) as the dms production does not only occur in the first model layer.

hamocc/hamocc_init.F
TomasTorsvik: I don't think we need to send the time information from hamocc_init.F to mo_dmsph.F90, unless we decide to make this a namelist variable. For now I think we can keep it simple and assume monthly data record for the surface PI pH input file:

`CALL GET_PI_PH(idm,jdm,omask)`

hamocc/mo_dmsph.F90
TomasTorsvik: I have defined a new parameter variable `pi_ph_record = 12`, so we don't need to pass this as an argument to the subroutine.

hamocc/mo_dmsph.F90
TomasTorsvik: I have replaced `kpke` with `pi_ph_record`.

cime_config/buildnml
```diff
@@ -1346,6 +1348,8 @@ cat >>! $RUNDIR/ocn_in$inststr << EOF
 ! NDEPFILE    : File name (incl. full path) for atmopheric N-deposition data
 ! INIXXX      : Initial condition file for iHAMOCC, where XXX=DIC, ALK, PO4,
 !               OXY, NO3, SIL, D13C, and D14C
+! WITH_DMSPH  : Logical switch to activate DMS calculation as function of pH
+! PI_PH_PATH  : File name (incl. full path) for pH input data
 &BGCNML
```
JorgSchwinger, line 1353: Since this is the full file name the variable name should be "PI_PH_FILE" or maybe "PI_PH_CLIM"

hamocc/mo_dmsph.F90
```diff
@@ -0,0 +1,146 @@
...
+    ! allocate pi_ph field
+    if(.not. allocated(pi_ph)) call alloc_pi_ph(kpie,kpje)
+
+    !
+    ! Open netCDF data file
+    !
+    IF(mnproc==1) THEN
+       ncstat = NF90_OPEN(trim(pi_ph_path)//'MONTHLY_PI_PH.nc',        &
+     &                   NF90_NOWRITE, ncid)
+       write(io_stdo_bgc,*) 'HAMOCC: opening MONTHLY_PI_PH file'
```
JorgSchwinger: Would be good to not have a fixed file name. The filename should at least indicate the grid, probably also the model version (or run) plus start/end dates from which the climatology is made. The user (or the scripting system) should be able to select the file.

JorgSchwinger (commented):
I think, if we agree that the idea of separating BLOM and iHAMOCC as far as possible, is in principle good, then we should not introduce a new dependency on the BLOM date structure. Why not doing it analogously to the n-deposition, where ini_ndep is called in hamocc_ini (reading in the data for all 12 month), then in hamocc_step there is a local array, that is first passed to get_ndep (getting the data for the current month), and then passed on to hamocc4bcm.

jmaerz:
Hi all, I agree that continuously getting iHAMOCC further separated from BLOM is the way to go. However, I think, it would be good to work on individual tasks (here: cpp flags and code structure/making things more flexible) in separate github issues to keep code changes fine grained and easy to track at least when it comes to these technical code issues. But I guess, it would be good to discuss/communicate further about this briefly at some point. 
Having just briefly discussed with Tomas: did you (@JorgSchwinger) mean to enable different names for the input file (so simply put a variable/string pointer instead of a fixed name) or to write a new function/module to retrieve e.g. grid information from the file name?

hamocc/mo_dmsph.F90
TomasTorsvik: I agree, in principle, but I would like to get this to work first with a version that is close to the original.

TomasTorsvik:
I made mo_dmsph independent of mod_time, keeping the separation between BLOM and iHAMOCC intact. My idea with this PR is to recover the functionality that was available in  the earlier version, but removing the DMSPH CPP flag. I would still like to limit this PR to only do this one task, and make a separate issue and PR to change the data input structure according to the n-deposition format. We have a similar structure for the SWACLIMFILE / mo_clim_swa used for computation of bromoform production, so we may want to have a more general discussion on data input structure before implementing these changes.

cime_config/buildnml
TomasTorsvik: I have kept the `PI_PH_FILE` variable name, but specify in the comment that the file name is set to `MONTHLY_PI_PH.nc`, so only the path and not the file name should be specified.

hamocc/mo_dmsph.F90
```diff
@@ -0,0 +1,145 @@
...
+    use mo_carbch, only: hi
+
+    implicit none
+
+    integer, intent(in) :: i,j,kplmon
+    real :: dms_ph
+
+    dms_ph  = 1. + (-log10(hi(i,j,1)) - pi_ph(i,j,kplmon))*dms_gamma
+
+  end function get_dmsph
```
JorgSchwinger: I this one line should be moved back to ocprod once pi_ph is passed to ocprod

JorgSchwinger (commented):
I see that it is good to test this close to the original code as a first step, but it would be very easy to make the structure of this consistent with the other input data (ndep,fedp, rivinput). I would be very easy to achieve as pointed out before. 

-Filenames should be read via namelist not just the path, this is important.
-Read the data file in hamocc_ini, get the data for a specific month in hamopcc_step and pass it on to hamocc4bcm

This doesn't require large changes to this PR. Of course if you prefer this can be done separately.

TomasTorsvik:
@JorgSchwinger OK, I see your point. The current commit works with the old structure,and I plan to keep all these commits when I merge with `master`, so the code evolution will be possible to follow. I will make some more commits in this PR to change the structure to the `ndep` version. Then we can figure out if we need something more flexible at a later stage.

hamocc/mo_dmsph.F90
TomasTorsvik: I'm not sure why we would need to pass `pi_ph` to `ocprod`. Currently I set `pi_ph` as `private`, and I think all the references to this field can be internal to `mo_dmsph`.

TomasTorsvik: A problem with passing pi_ph through the argument list is that it may not be allocated, and probably we don't want to allocate memory for pi_ph unless we actually use it.

JorgSchwinger: Yes, that is true, that would be the price to pay. Regarding memory I wouldn't see this as a big issue since it is a 2d field. So  if `with_phdms=.false.`, we would pass a 2d-field set to 0.0 by `get_ph_clim` (or whatever the name is)

TomasTorsvik:
@tjiputra @JorgSchwinger 
I have changed the routines so that `mo_read_pi_ph` is only referenced outside `hamocc4bcm`, and pass `pi_ph` as an argument through `hamocc4bcm -> ocprod`. If there are no further comments I will merge this PR later today.

hamocc/mo_control_bgc.F90
```diff
@@ -62,6 +62,7 @@ MODULE mo_control_bgc
 ! Logical switches
       LOGICAL, SAVE :: do_ndep=.true.    ! apply n-deposition   (set via namelist)
       LOGICAL, SAVE :: do_rivinpt=.true. ! apply riverine input (set via namelist)
+      logical :: with_dmsph = .false.    ! apply DMS with pH dependence (set via namelist)
```
JorgSchwinger: would suggest to add "save" here, too

hamocc/mo_read_pi_ph.F90
```diff
@@ -0,0 +1,189 @@
...
+       WRITE(io_stdo_bgc,*)'Memory allocation for variable pi_ph ...'
+       WRITE(io_stdo_bgc,*)'First dimension    : ',kpie
+       WRITE(io_stdo_bgc,*)'Second dimension   : ',kpje
+    ENDIF
+
+    ALLOCATE (pi_ph(kpie,kpje),stat=errstat)
+    if(errstat.ne.0) stop 'not enough memory pi_ph'
+    pi_ph(:,:) = 0.0
+
+  end subroutine alloc_pi_ph
```
JorgSchwinger, line 158: I find this unnecessarily complicated, the other fields (dust, ndep) are just local variables in hamocc_step, they are created/removed automatically when hamocc_step is called/finalized.

JorgSchwinger (approved):
Thanks Tomas, I have left two minor comments, but feel free to merge, this can also fixed or left as it is later.

hamocc/mo_read_pi_ph.F90
TomasTorsvik: Think I will leave it like this for now. Jerry mentioned that he might introduce a 3D PI_pH at some point in the future. If so, I guess we might need to think again about this structure.

PR #162, opened 2022-05-12
Feature hamocc sediment spinup

JorgSchwinger:
@jmaerz I am not able to request a review from you. Maybe you can have a look nevertheless (and probably you need to be added by someone to something that you can be requested as reviewer?).


JorgSchwinger:
I have tested for bit-identical results for the NOINYOC compset on the T62_tn21 grid (OMIP1 simulation) and a fully coupled NorESM2 1%-CO2 simulation.

By the way, the previous BLOM version that is in the master currently wouldn't compile for fully coupled compsets (variable "atm" wasn't included in use statement in hamocc4bcm) - the automated testing doesn't seem to catch such errors?

jmaerz:
Hi @JorgSchwinger , I'll try to have a look into this soon (being a bit time constrained). With regards to the compilation error: thanks for catching this. I suspect, the automated testing might not have included the right pre-processor flag(s) which would have enabled to catch this error, but I haven't looked into this, thus far. For other cases, NorESM was compiling.

TomasTorsvik:
`inventory_bgc.F90` is included in the Files changed list, but only has whites pace changes. Did something get lost in the merging? Otherwise I think this file can be checked out from `master` and removed from this PR.

TomasTorsvik:
The `save` attributes in `mo_control_bgc.F90` are redundant, so this only has a stylistic function. Maybe it can be useful to see the `save` attribute if you run a grep command on the source code, but otherwise I don't see a reason for doing it this way (as a general code style).

hamocc/hamocc4bcm.F90
```diff
@@ -173,7 +175,7 @@ SUBROUTINE HAMOCC4BCM(kpie,kpje,kpke,kbnd,kplyear,kplmon,kplday,kldtday,&
 #endif
 
 #ifdef BROMO
-!$OMP PARALLEL DO
+!$OMP PARALLEL DO PRIVATE(i)
```
TomasTorsvik, line 182: Separate openMP bugfix?

hamocc/hamocc4bcm.F90
```diff
@@ -353,7 +377,7 @@ SUBROUTINE HAMOCC4BCM(kpie,kpje,kpke,kbnd,kplyear,kplmon,kplday,kldtday,&
 !--------------------------------------------------------------------
 ! Pass bromoform flux. Convert unit from kmol CHBr3/m^2 to kg/m^2/s.
 ! Negative values to the atmosphere
-!$OMP PARALLEL DO
+!$OMP PARALLEL DO PRIVATE(i)
```
TomasTorsvik, line 398: Separate openMP bugfix?

TomasTorsvik:
The openMP bug fixes seem to be unrelated to the sediment spinup. Would it make sense to do this in a separate commit / PR? Probably this should also be included as a bug fix to the `release-1.2` branch.

hamocc/powach.F90
```diff
@@ -105,6 +106,9 @@ subroutine powach(kpie,kpje,kpke,kbnd,prho,omask,psao)
 ! needed for boundary layer vertilation in fast sediment routine
 
   real :: bolven(kpie)
+  
+! Set array for saving diffusive sediment-water-column fluxes to zero
+  sedfluxo(:,:,:) = 0.0
```
TomasTorsvik, line 111: Here you set `sedfluxo` to zero every time `powach` is called, but in `dipowa` this is written as something should be accumulated.

hamocc/dipowa.F90
```diff
@@ -208,9 +208,9 @@ subroutine dipowa(kpie,kpje,kpke,omask)
                 &  ( sedb1(i,l,iv) - tredsy(i,l,3) * powtra(i,j,l+1,iv) )      &
                 &  / tredsy(i,l,2)
 
-           ! used in inventory_bgc/maschk (diagnostics)
+           ! diffusive fluxes (positive downward)
            sedfluxo(i,j,iv) = sedfluxo(i,j,iv)                                 &
-                &  + ocetra(i,j,kbo(i,j),iv) - aprior
+                &  -(ocetra(i,j,kbo(i,j),iv) - aprior)* bolay(i,j)
```
TomasTorsvik, line 213: Should this be without accumulation in `sedfluxo`?

TomasTorsvik (approved):
Thanks, Jörg, I think this looks very good!  I have left some comments that you may want to address, but probably it can also be merged as is and we can deal with these things afterwards.

hamocc/dipowa.F90
JorgSchwinger: I think this is correct as it is: In `powach` `sedfluxo` is set to zero when entering the sediment. Then it is receiving a value for silicate and oxygen (but not the others) in `powach`. After that `dipowa` is called, where sedfluxo is updated in a loop. In order to avoid overwriting the values for oxygen and silicate, it is written as above.

hamocc/hamocc4bcm.F90
JorgSchwinger: I agree, I should have done this in a separate commit, but now I don't have the time to disentangle this so I'll leave it as it is.

JorgSchwinger:
> 

Regarding the save attributes in modules: It's not true that they are redundant. The Fortran standard only guarantees that the values of module variables are "saved" if a variable is referenced in the compiled main program (same as for COMMON blocks). 

I don't think that any compiler would NOT save the values of module variables that are not referenced in main, but strictly speaking this is necessary (as far as I know, maybe my knowledge is outdated).

PR #167, opened 2022-05-18
Set return value for bromoform fluxes also if the scheme is not activated

JorgSchwinger:


TomasTorsvik (approved):
Agree, looks reasonable.

PR #168, opened 2022-05-18
Revise modules for handling of external inputs (N-dep, rinvers, iron)…

JorgSchwinger:
…, split into modules for reading the data and modules that apply the fluxes in core hamocc


JorgSchwinger:
Following our discussions last week, I wanted to keep the momentum and do the split into reading and applying input-data. In my opinion this is an improvement, as it makes the code structure clearer. I have also moved the definition of the river-indices to mo_param1, such that all indice-definitions are collected there. The "apply"-modules really only encapsulate the corresponding subroutines. We could/should also apply the iron deposition outside `ocprod` using a dedicated module (as it is now, rivers and n-deposition come after ocprod, while iron is applied at the beginning of `ocprod` - that doesn't make much sense, but changing this would break bit-for-bit reproducibility). @jmaerz: I still can't add you as reviewer, but maybe you could have a look.

hamocc/mo_apply_rivin.F90
```diff
@@ -0,0 +1,162 @@
...
+  rivinflx(:,:,:) = 0.0
+
+  if (.not. do_rivinpt) return
+  
+!$OMP PARALLEL DO PRIVATE(i,k,fdt,volij)
+  DO j=1,kpje
+  DO i=1,kpie
+    IF(omask(i,j).GT.0.5) THEN
+
+      fdt   = dtb/365.
```
jmaerz: Hi Jörg, we could do this outside the loop, since it is constant - but we can do this later as well.

hamocc/mo_apply_rivin.F90
```diff
@@ -0,0 +1,162 @@
...
+                                                                    + rivin(i,j,irdin)*fdt/volij  &
+                                                                    + rivin(i,j,irdip)*fdt/volij
+      ocetra(i,j,1:kmle,ialkali)    = ocetra(i,j,1:kmle,ialkali)    + rivin(i,j,iralk)*fdt/volij
+#ifdef natDIC
+      ocetra(i,j,1:kmle,inatsco212) = ocetra(i,j,1:kmle,inatsco212) + rivin(i,j,iralk)*fdt/volij  &
+                                                                    + rivin(i,j,irdin)*fdt/volij  &
+                                                                    + rivin(i,j,irdip)*fdt/volij
+      ocetra(i,j,1:kmle,inatalkali) = ocetra(i,j,1:kmle,inatalkali) + rivin(i,j,iralk)*fdt/volij
+#endif
+      ocetra(i,j,1:kmle,iiron)      = ocetra(i,j,1:kmle,iiron)      + rivin(i,j,iriron)*fdt/volij*0.01
```
jmaerz: With the fully new restructuring, I feel it would be nice to replace the explicit value of 0.01 by a parameter value (also in line 148), but we can do this later as well.

jmaerz:
Hi Jörg, thanks for your initiative! I was going through your changes and left two minor comments that we can also work out later (since they do not necessarily belong to the modularization). With regards to the iron application, I would suggest to open the module files, while (for now?) keeping the order of application (which then, at some point, could be easily changed by re-locating one line) - potentially placing the apply_iron (or similar) already in hamocc4bgm to make ocprod a bit cleaner. That would enable us to still be compatible with the CMIP6 model version, while already having the modular structure in place. What do you think? Wrt to me as reviewer, Tomas sent an email, but it seems that thus far, I haven't been added to NorESMhub.

JorgSchwinger:
Thanks Joeran, all agreed. I have pushed new changes.

TomasTorsvik (approved):
Good initiative, Jörg. As you say, we should probably do the "apply" steps in a more consistent way, but it's good to make the changes in incremental steps (especially if something breaks backward compatibility).

PR #171, opened 2022-06-02
Fix the compset name for the SSP5-3.4 scenario to conform to the name…

JorgSchwinger:
… already used in other NorESM2 components

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -96,7 +96,7 @@
       <value compset="SSP370REFGHGLOWNTCF_CAM60%NORESM.*_BLOM%ECO">ssp370</value>
       <value compset="SSP434_CAM60%NORESM.*_BLOM%ECO"  >ssp434</value>
       <value compset="SSP460_CAM60%NORESM.*_BLOM%ECO"  >ssp460</value>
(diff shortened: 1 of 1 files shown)


JorgSchwinger:
This is necessary since in CLM the name SSP534 (without "OS") is already in use. To make this work also the other components need to be updated.

JorgSchwinger:
Probably good to cherry-pick this into the release branch?

TomasTorsvik (approved):
Is this a change that needs to be implemented across several NorESM components? If so, it would make sense to raise this as an issue in the NorESM repository, and coordinate for a new bug-fix release.

JorgSchwinger:
Well, it is generally not a bug fix. The SSP5-3.4 scenario was run by Ali and has been published to ESGF, but the compset definition has not been included into a NorESM2 release yet.

So yes, I have more changes for the other components, but I'm not sure which branch etc. they should go into.

matsbn:
Agree that this could be cherry-picked into release branches and then create v.1.1.1 and/or v1.2.1 release tags. Maybe the update to release branches should wait until compset changes for other components are in place to prepare for a coordinated NorESM bug fix release (although not really a bug).

matsbn (approved):
Seems fine to merge into master.

PR #172, opened 2022-06-08
Update meson & bugfix in accfields

TomasTorsvik:
These updates are needed for building BLOM/iHAMOCC with meson after the latest code changes. The meson update and "sedbypass" CPP flags are separate issues, but I include them in the same PR since meson will not build with iHAMOCC without both fixes.


JorgSchwinger (approved):
Looks good, thanks for spotting this!

PR #174, opened 2022-06-22
Fix timestep inventory

jmaerz:
This small fix provides the preprocessor flag `HAMOCC_DEBUG` via `xmlchange` that enables to debug iHAMOCC processes on a time step basis, i.e. the mass conservation of the inventory after individual processes. In addition, the time step based inventory calculation now accounts for particulate fluxes to the sediments.


TomasTorsvik:
This looks fine to me, but I would like to wait with this PR until we have merged in changes for the hybrid coordinates.

TomasTorsvik:
@jmaerz I think we can merge this PR now. Is that OK with you, or do you have any updates you want to include first?

jmaerz:
@TomasTorsvik I don't have any further updates, so from my side, this can be merged.

PR #175, opened 2022-06-27
Ensure checkout of CVMix tag v0.98-beta

matsbn:
Ensure checkout of CVMix tag v0.98-beta for both git submodule and CESM checkout_externals functionality. Also the CVMix URL in .gitmodules has been changed from https to ssh. The code has been checked with BLOM_VCOORD = cntiso_hybrid and it reproduces compset N1850frc2 at resolution f09_tn14 on machine Betzy.


milicak (approved):
Looks good to me

PR #176, opened 2022-06-27
Hybrid vertical coordinate

matsbn:
Support for a generalized vertical coordinate has been added, involving high-order reconstruction, regridding and remapping methods, adaption of physical parameterizations and use of KPP, as implemented in the CVMix package, for the estimation of vertical diffusivities.

With the generalized vertical coordinate framework, a hybrid coordinate with pressure coordinates towards the surface and continuous isopycnal below is currently supported. The previous vertical coordinate, combining a bulk surface mixed layer with isopycnic layers below, is still fully supported.

Using the N1850frc2 NorESM compset at resolution f09_tn14, the code has been tested on machine betzy.sigma2.no to bit-reproduce current NorESMhub/BLOM/master and correctly restart.


JorgSchwinger (approved):
Of course I can't really review a pull request of this size, but I'm sure this is all very well thought through and tested. So I would suggest to merge it and take it from there.

milicak (approved):

PR #179, opened 2022-07-19
Hamocc hybrid coord2

TomasTorsvik:
For hybrid cooridinates, define depth of surface mixed layer in HAMOCC by the ocean boundary layer (OBL) variable provided by CVmix. Default value kml(i,j) = 2 used for isopycnic variables for consistency with the bulk mixed layer formulation.

### DIFF
--- a/hamocc/cyano.F90
+++ b/hamocc/cyano.F90
@@ -66,7 +66,7 @@ SUBROUTINE CYANO(kpie,kpje,kpke,kbnd,pddpo,omask,ptho)
   use mo_param1_bgc, only: ialkali,iano3,igasnit,iphosph,ioxygen
   use mo_vgrid,      only: kmle
 #ifdef natDIC
-      use mo_param1_bgc, only: inatalkali
+  use mo_param1_bgc, only: inatalkali
 #endif
 
   implicit none
@@ -88,42 +88,42 @@ SUBROUTINE CYANO(kpie,kpje,kpke,kbnd,pddpo,omask,ptho)
 ! it is assumed here that this process is limited to the mixed layer
 !
   DO j=1,kpje
-     DO i=1,kpie
-        IF(omask(i,j).gt.0.5) THEN
-           DO k=1,kmle(i,j)
-              IF(ocetra(i,j,k,iano3).LT.(rnit*ocetra(i,j,k,iphosph))) THEN
+  DO i=1,kpie
+  IF(omask(i,j).gt.0.5) THEN
+     DO k=1,kmle(i,j)
+     IF(ocetra(i,j,k,iano3).LT.(rnit*ocetra(i,j,k,iphosph))) THEN
 
-                 oldocetra = ocetra(i,j,k,iano3)
-                 ttemp = min(40.,max(-3.,ptho(i,j,k)))
+        oldocetra = ocetra(i,j,k,iano3)
+        ttemp = min(40.,max(-3.,ptho(i,j,k)))
 
 ! Temperature dependence of nitrogen fixation, Kriest and Oschlies 2015.
-                 nfixtfac = MAX(0.0,tf2*ttemp*ttemp + tf1*ttemp + tf0)/tff
+        nfixtfac = MAX(0.0,tf2*ttemp*ttemp + tf1*ttemp + tf0)/tff
 
-                 ocetra(i,j,k,iano3)=ocetra(i,j,k,iano3)*(1-bluefix*nfixtfac)  &
-                      &    + bluefix*nfixtfac*rnit*ocetra(i,j,k,iphosph)
+        ocetra(i,j,k,iano3)=ocetra(i,j,k,iano3)*(1-bluefix*nfixtfac)            &
+             &    + bluefix*nfixtfac*rnit*ocetra(i,j,k,iphosph)
 
-                 dano3=ocetra(i,j,k,iano3)-oldocetra
+        dano3=ocetra(i,j,k,iano3)-oldocetra
 
-                 ocetra(i,j,k,igasnit)=ocetra(i,j,k,igasnit)-dano3*(1./2.)
+        ocetra(i,j,k,igasnit)=ocetra(i,j,k,igasnit)-dano3*(1./2.)
 
 ! Note: to fix one mole N2 requires: N2+H2O+y*O2 = 2* HNO3 <-> y=2.5 mole O2.
 ! I.e., to release one mole HNO3 = H+ + NO3- requires 1.25 mole O2
-                 ocetra(i,j,k,ioxygen)=ocetra(i,j,k,ioxygen)-dano3*1.25
+        ocetra(i,j,k,ioxygen)=ocetra(i,j,k,ioxygen)-dano3*1.25
 
 ! Nitrogen fixation followed by remineralisation and nitrification decreases
 ! alkalinity by 1 mole per mole nitrogen fixed (Wolf-Gladrow et al. 2007)
-                 ocetra(i,j,k,ialkali)=ocetra(i,j,k,ialkali)-dano3
+        ocetra(i,j,k,ialkali)=ocetra(i,j,k,ialkali)-dano3
 #ifdef natDIC
-                 ocetra(i,j,k,inatalkali)=ocetra(i,j,k,inatalkali)-dano3
+        ocetra(i,j,k,inatalkali)=ocetra(i,j,k,inatalkali)-dano3
 #endif
 
-                 intnfix(i,j) = intnfix(i,j) +                                 &
-                      &         (ocetra(i,j,k,iano3)-oldocetra)*pddpo(i,j,k)
+        intnfix(i,j) = intnfix(i,j) +                                           &
+             &         (ocetra(i,j,k,iano3)-oldocetra)*pddpo(i,j,k)
 
-              ENDIF
-           ENDDO
-        ENDIF
+     ENDIF
      ENDDO
+  ENDIF
+  ENDDO
   ENDDO
 
 END SUBROUTINE CYANO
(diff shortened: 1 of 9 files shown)


phy/mod_difest.F
```diff
@@ -1161,7 +1162,7 @@ subroutine difest_vertical_hyb(m,n,mm,nn,k1m,k1n)
      .                               cellHeight,OBLdepth(i,j))
 
              ! gets index of the level and interface above hbl
-             kOBL = CVMix_kpp_compute_kOBL_depth(iFaceHeight,
+             kOBL(i,j) = CVMix_kpp_compute_kOBL_depth(iFaceHeight,
```
jmaerz: I am a bit uncertain, but `CVMix_kpp_compute_kopb_depth()` i) seem to return a real value and ii) depending on which index one wants, it potentially requires to subtract 1 (not sure about the vertical grid layout in BLOM) iii) why calculating it again - and not using hOBL? - from CVMix:
```
! !DESCRIPTION:
!  Computes the index of the level and interface above OBL\_depth. The index is
!  stored as a real number, and the integer index can be solved for in the
!  following way:\\
!    \verb|kt| = index of cell center above OBL\_depth = \verb|nint(kOBL_depth)-1|
!    \verb|kw| = index of interface above OBL\_depth = \verb|floor(kOBL_depth)|
```
In any case, to be clean, I guess it would be good to explicitly convert from real to integer for the k-index. This applies again below.

TomasTorsvik: Thanks @jmaerz ! 

@matsbn @milicak I would like to hear from you about this issue.

I did not look into the CVMix routine itself when I made these code changes, but I agree that this looks suspicious. For iHAMOCC we probably want the index of the cell center above `OBL_depth`, not the index of the interface. 

`hOBL` is computed twice, first by the subroutine `CVMix_kpp_compute_OBL_depth` and later by function `CVMix_kpp_compute_kOBL_depth`. The function `CVMix_kpp_compute_kOBL_depth` is used internally in `CVMix_kpp_compute_OBL_depth`, so there should be no need to do this again. Then `kOBL` is computed (twice) by the function `CVMix_kpp_compute_kOBL_depth`.

`kOBL` is used in the calculation of buoyancy flux acting on the OBL, so in this case the index may need to reference the interface. If so, we may want to define a new index `OBL_cell_center = nint(hOBL) - 1` for use in iHAMOCC.

TomasTorsvik:
Will update this PR depending on feedback for PR #182

jmaerz:
Thanks for keeping that going!

hamocc/cyano.F90
```diff
@@ -61,74 +61,69 @@ SUBROUTINE CYANO(kpie,kpje,kpke,kbnd,pddpo,omask,ptho)
...
-      ENDDO
-
+                 intnfix(i,j) = intnfix(i,j) +                                 &
+                      &         (ocetra(i,j,k,iano3)-oldocetra)*pddpo(i,j,k)
 
+              ENDIF
+           ENDDO
+        ENDIF
+     ENDDO
+  ENDDO
```
JorgSchwinger, line 127: Maybe not the most important aspect, but I would favor keeping the traditional HAMOCC-style indentation for spatial indices... There is only very few things that were good in HAMOCC regarding code style, but this was one of them. In my opinion

`do k=1,kpke`
`do j=1,kpje`
`do i=1,kpie`

is much preferable over 

`do k=1,kpke`
`___do j=1,kpje`
`______do i=1,kpie`

in terms of readability.

JorgSchwinger (commented):
I haven't dived into the kOBL-issue, but from the HAMOCC side this looks fine to me.

hamocc/cyano.F90
jmaerz: I agree with @JorgSchwinger that the HAMOCC part looks ok. Wrt the loops: I guess, we discussed this already a bit... - I was always puzzled with the non-indentation (left me loosing track, where the loops start and end), but I also don't have a really strong opinion about it either and could live with this exception, if needed/wished (as long in other cases indentation is allowed).

TomasTorsvik: I re-set the indentation pattern to the traditional HAMOCC-style in cyano.F90. In terms of readability, I think both styles are OK. My code editor auto-indents all loops, and in this case I changed the loop order since `k=1,kmle(i,j)` has to be the inner loop.

JorgSchwinger: Thanks Tomas! 

Just to explain a bit more what I mean: In most cases in HAMOCC the i,j-loops (and to a lesser degree also the k-loops) are just loops over spatial indices. The logic is NOT "there is a loop over i where something is done, and in the coarse of doing this there is then also a loop over j", but in almost all cases the logic is "there is an operation done in loops over all horizontal indices". In the first case it would makes sense to indent, but in the second case it is one logical block, where it doesn't really make sense to indent, in my opinion. In principle, these loops could be replaced by modern Fortran ":" notation (not that I think we should do this anytime soon!).

Not that I have a very strong opinion about this, but as mentioned before, this is one of the very few things where I think historical HAMOCC style was good and made sense.

jmaerz: Hi @JorgSchwinger , thanks for the detailed explanation and please forgive me (and remind me) when I am unpurposely deviate from that style (due to old habits). I'll try to adjust my developments accordingly.

TomasTorsvik:
Hi,

I have revived this PR after changing the dependency between BLOM and iHAMOCC, so now I'm porting the real value surface mixed layer depth hOBL, corresponding to the fractional depth in the index space, instead of the integer index value kOBL.

@matsbn , @milicak 
There is only a minor change in `mod_difest.F` where I have moved the hOBL variable initiation from the `difest_vertical_hyb` to the main module initiation section, and made it `public` so that it is accessible to iHAMOCC. This should not change anything substantial in the physics part.

@jmaerz , @JorgSchwinger 
The iHAMOCC implementation is mostly as before, only that now the `kmle` variable is defined from `hOBL` instead of `kOBL`. I had a problem with the default value `spval = 1.e33`, which doesn't fit a regular integer type, hence I included a test to exclude `hOBL` instances with this value before attempting to do `nint(hOBL)`, and I set `kmle` to the default value 2 if this test fails. I'm not sure this is the best solution to this problem, please let me know if you have suggestions how to improve this.

matsbn:
In routine inivar_difest, you could initialise hOBL(:,:) with the following code block at the end of the routine: 
```
c$OMP PARALLEL DO PRIVATE(l,i)
      do j=1,jj
        do l=1,isp(j)
        do i=max(1,ifp(j,l)),min(ii,ilp(j,l))
          hOBL(i,j)=3.
        enddo
        enddo
      enddo
c$OMP END PARALLEL DO
```
This should give kmle = 2 always at ocean grid cells, except when vcoord_type_tag == cntiso_hybrid. Maybe then you do not have to access the mod_vcoord variables in iHAMOCC for now.

jmaerz:
Hi @TomasTorsvik , just to understand the problem with the `spval` value: do I see it right that the moment, the whole water column is mixed, you experience the problem that `hOBL` is still equal to `spval`? - or when does the problem happen? Is `spval` needed for initialization for `hOBL` or could you also use e.g. 9999 as initialization (expecting to never reach more than 9998 layers in BLOM in the future). In any case, we **don't** want to end up with `kmle > number of layers`.

As minor thing: are the line breaks (the re-formatting) in `mo_apply_rivin` needed?

Apart from that the changes look fine to me.

TomasTorsvik:
> As minor thing: are the line breaks (the re-formatting) in `mo_apply_rivin` needed?

No, I changed it back to the the original form.

TomasTorsvik:
> Hi @TomasTorsvik , just to understand the problem with the `spval` value: do I see it right that the moment, the whole water column is mixed, you experience the problem that `hOBL` is still equal to `spval`? - or when does the problem happen? Is `spval` needed for initialization for `hOBL` or could you also use e.g. 9999 as initialization (expecting to never reach more than 9998 layers in BLOM in the future). In any case, we **don't** want to end up with `kmle > number of layers`.

Hi @jmaerz ,
the update of `OBLdepth` and `hOBL` in `mod_difest.F` is only done for "wet" cells, so the `spaval` value remained for land cells. If I had done the looping differently, i.e.
```
   do j=1,jj
      do l=1,isp(j)
         do i=max(1,ifp(j,l)),min(ii,ilp(j,l))
            kmle(i,j) = nint(hOBL(i,j))-1
         enddo
      enddo
   enddo
```
the loop should not encounter any `spval` values. However, if for some reason an `spval` value was encountered, there is the potential for having `kmle > number of layers`, as you say.

I have changed the initiation for `hOBL` so that it gets a default value of 3. (defined by `hOBL_static`), so that `kmle` gets a default value of 2. In this way we don't need to introduce an explicit dependence on `vcoord_type_tag` in iHAMOCC, at least for now.

jmaerz:
Ok, I see, so it was more of a theoretical problem and you haven't encountered it really. Good to know about it and thanks for the information! From my side, you've my approval to merge.

jmaerz (approved):
Approved.

matsbn (approved):
It should be OK now. I would prefer that hOBL is first set to spval everywhere, also including the halo, then initialised to hOBL_static only for wet points. The purpose of first initialising to a very large value, spval, is that the code would typically crash if using the variable where one shouldn't (e.g at land points). This can be corrected later.

JorgSchwinger (approved):

TomasTorsvik:
> It should be OK now. I would prefer that hOBL is first set to spval everywhere, also including the halo, then initialised to hOBL_static only for wet points. The purpose of first initialising to a very large value, spval, is that the code would typically crash if using the variable where one shouldn't (e.g at land points). This can be corrected later.

Thanks @matsbn , I see that iHAMOCC is referencing the wet points indexing at some places, but I'm not sure if this is done consistently. I therefore prefer to merge this version of the PR, and make a separate issue on the use of wet points indexing in iHAMOCC.

PR #181, opened 2022-08-15
Fix initialization of wdust for dust settling

jmaerz:
The settling velocity of 'free' dust `wdust` seem to have not been properly initialized when running the model without the `-D AGG` flag. Further, some surface input flux could happen, since `wdust`, correctly initialized, would always be positive. Hence, I introduced `wdustd` (similar to `wpocd`, etc.). This bugfix can affect the sediment budget, thus the burrying and sediment remineralization rates.

### DIFF
--- a/hamocc/beleg_parm.F90
+++ b/hamocc/beleg_parm.F90
@@ -193,13 +193,6 @@
       wmax  = 60.*dtb       !m/d   maximum sinking speed
       wlin  = 60./2400.*dtb !m/d/m constant describing incr. with depth, r/a=1.0
 #endif
-      ! Dust sinking speed (individual clay particle)
-      dustd1 = 0.0001 !cm = 1 um, boundary between clay and silt
-      dustd2=dustd1*dustd1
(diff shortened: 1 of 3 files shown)


TomasTorsvik:
We might want to place this bug fix in a new development branch instead of the `master` branch.

jmaerz:
> 

Yes, we can do so. I just wanted to prepare the fix and make you aware of the issue.

jmaerz:
Argh, I just realized, I was wrong wrt initialization. The issue with the surface flux might be an issue though, which I need to check. Sorry for the confusion.

jmaerz:
I reverted changes related to the initialization. However, the fix for the surface flux of free dust remains, and with that the potential change of sediment-related processes.

JorgSchwinger:
Thanks for spotting this, looks good. I agree with Tomas that we shouldn't merge this into master right now. We should discuss this very soon, and make a plan for when to start merging answer changing changes (as raised by Tomas in #169).

JorgSchwinger (approved):

PR #182, opened 2022-08-19
Bugfix kOBL

TomasTorsvik:
@milicak @matsbn 
Hi,
the assignment of kOBL looks strange to me (see #179). The code change suggested here is according to code description in CVMix, assuming that kOBL should represent the index of the cell center above the OBL depth (this is what we want to use in iHAMOCC). If this was not your intention, we should perhaps define a new index variable according to `kt` below, that can be imported in iHAMOCC.

> DESCRIPTION:
> Computes the index of the level and interface above OBL\_depth. The index is
> stored as a real number, and the integer index can be solved for in the
> following way:\\
>   \verb|kt| = index of cell center above OBL\_depth = \verb|nint(kOBL_depth)-1|
>   \verb|kw| = index of interface above OBL\_depth = \verb|floor(kOBL_depth)|

### DIFF
--- a/phy/mod_difest.F
+++ b/phy/mod_difest.F
@@ -1205,10 +1205,6 @@ c --- ------ make sure it is in metrics if stability depends on rho
      .              kk,                                ! (in) Number of levels in array shape
      .              CVMix_kpp_params_user=KPP_params ) ! KPP parameters
 
-             ! gets index of the level and interface above hbl
-             kOBL = CVMix_kpp_compute_kOBL_depth(iFaceHeight,
-     .                               cellHeight,OBLdepth(i,j))
-
 c ---- ccc -------
              ! convert m2/s to cm2/s
              Kv_kpp = Kv_kpp*1e4
--- a/phy/mod_difest.F
+++ b/phy/mod_difest.F
@@ -1161,8 +1161,8 @@ c --- ------ centers.
      .                               cellHeight,OBLdepth(i,j))
 
              ! gets index of the level and interface above hbl
-             kOBL = CVMix_kpp_compute_kOBL_depth(iFaceHeight,
-     .                               cellHeight,OBLdepth(i,j))
+             kOBL = int(hOBL(i,j))    ! index of interface above OBL depth
+
 c --- ------ Diapycnal mixing when local stability is weak
 c --- ------ convection routine based on N2 not rho
 c --- ------ make sure it is in metrics if stability depends on rho


matsbn:
I think kOBL as used in mod_difest.F is supposed to be the index of the interface (not cell center) above the OBLdepth. Then line 1164 should be kOBL = int(hOBL(i,j)), which seems consistent with the previous implicit truncation when converting from real to integer (better to do the conversion explicitly).

Then in the argument of CVMix_coeffs_conv, I wonder if kOBL is correct to use. This is similar as in MOM6, but the routine cvmix_coeffs_conv_wrap in cvmix_convection.F90 seems to use a different index (similar to nint(hOBL(i,j))+1). I hope @milicak can comment on this.

Would it be an idea to pass hOBL(:,:) to iHAMOCC? Then iHAMOCC can decide on the most appropriate definition (nint(hOBL(i,j))-1 : cell center above OBLdepth; int(hOBL(i,j)) : interface above OBLdepth).

milicak:
I agree with @matsbn -1 doesn't make sense to me. we shoudl change it as kOBL = int(hOBL(i,j)).
However, what he suggested is also perfect solution, iHAMOCC can decide what to do with kOBL rather than changing the BLOM code with -1.

TomasTorsvik:
Hi @matsbn , @matsbn 
thanks for your comments. I think for the physics you may want to use **floor** instead of **nint**, to make sure you always get the interface above hOBL. Since I have this PR already open I have suggested an implementation here (see latest update).

I still need to pass hOBL(i,j) as a field to iHAMOCC, but I can make this change in a separate PR.

TomasTorsvik:
Sorry, **int** and **floor** should behave the same way in this case, but **nint** could have a different output.

matsbn:
I would prefer kOBL = int(hOBL(i,j)) over kOBL = floor(hOBL(i,j)). They should behave identical for the positive real values in question here, but kOBL = int(hOBL(i,j)) is consistent with the CVMix comment on how to obtain the interface index (also, although hardly important, int might be slightly more efficient compared to floor).

TomasTorsvik:
> I would prefer kOBL = int(hOBL(i,j)) over kOBL = floor(hOBL(i,j)). They should behave identical for the positive real values in question here, but kOBL = int(hOBL(i,j)) is consistent with the CVMix comment on how to obtain the interface index (also, although hardly important, int might be slightly more efficient compared to floor).

Done

matsbn (approved):
Looks good!

PR #185, opened 2022-08-22
Fix wdust

jmaerz:
This is identical to pull request #181 - but now target branch is `feature-hamocc_beyond-CMIP6`.


TomasTorsvik (approved):
Since this was reviewed already for PR #181 I assume this is still OK.

PR #187, opened 2022-08-29
Add framwork for applying ocean alkalinization scenarios

JorgSchwinger:
This PR adds two modules to iHAMOCC that read and apply an alkalinity flux to the surface layer. Future addition of more alkalinization scenarios can be done easily by modifications in mo_read_oafx.F90.

There is currently no configuration through the NorESM xml-files available. For now, this is only for expert users who can set ´do_oalk´ to true in user_nl_blom. If ´do_oalk´ is set to false (the default) the model produces bit-identical results compared to the CMIP6 version.

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -207,11 +207,11 @@ if ($BLOM_N_DEPOSITION == TRUE) then
     set NDEPFNAME = ndep_201501-210012-${BLOM_NDEP_SCENARIO}_tnx1v4_20191112.nc
   else if( $BLOM_NDEP_SCENARIO == UNSET ) then
     set DO_NDEP   = .false.
-    set NDEPFNAME = ""
+    set NDEPFNAME = "''"
   endif
 else
   set DO_NDEP   = .false.
-  set NDEPFNAME = ""
+  set NDEPFNAME = "''"
 endif
 if ($HAMOCC_SEDSPINUP == TRUE) then
   set DO_SEDSPINUP = .true.
@@ -230,8 +230,13 @@ if ($HAMOCC_VSLS == TRUE && $OCN_GRID != tnx1v4) then
   echo "$0 ERROR: HAMOCC_VSLS == TRUE not possible with this grid resolution (no swa-climatology available) "
   exit -1
 endif
+# For the following options, there are currently no switches in Case-XML files.
+# These options can be activated by expert users via user namelist.
+set DO_OALK = .false.
+set OALKSCEN = "''"
+set OALKFILE = "''"
 set WITH_DMSPH = .false.
-set PI_PH_FILE = ""
+set PI_PH_FILE = "''"
 
 # set DIAPHY defaults 
 set GLB_FNAMETAG = "'hd','hm','hy'"  
@@ -1448,6 +1453,9 @@ cat >>! $RUNDIR/ocn_in$inststr << EOF
   RIVINFILE    = $RIVINFILE 
   DO_NDEP      = $DO_NDEP
   NDEPFILE     = $NDEPFILE
+  DO_OALK      = $DO_OALK
+  OALKSCEN     = $OALKSCEN
+  OALKFILE     = $OALKFILE
   DO_SEDSPINUP = $DO_SEDSPINUP
   SEDSPIN_YR_S = $SEDSPIN_YR_S
   SEDSPIN_YR_E = $SEDSPIN_YR_E
@@ -1853,6 +1861,11 @@ EOF
   if ($BLOM_N_DEPOSITION == TRUE) then
 cat >> $CASEBUILD/blom.input_data_list << EOF
 n_deposition_file = `echo $NDEPFILE | tr -d '"' | tr -d "'"`
+EOF
+  endif
+  if ($OALKFILE != "''") then
+cat >> $CASEBUILD/blom.input_data_list << EOF
+oafx_file = `echo $OAFILE | tr -d '"' | tr -d "'"`
 EOF
   endif
   if ($HAMOCC_VSLS == TRUE) then
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -1865,7 +1865,7 @@ EOF
   endif
   if ($OALKFILE != "''") then
 cat >> $CASEBUILD/blom.input_data_list << EOF
-oafx_file = `echo $OAFILE | tr -d '"' | tr -d "'"`
+oafx_file = `echo $OALKFILE | tr -d '"' | tr -d "'"`
 EOF
   endif
   if ($HAMOCC_VSLS == TRUE) then
--- a/hamocc/hamocc4bcm.F90
+++ b/hamocc/hamocc4bcm.F90
@@ -19,7 +19,7 @@
 
       SUBROUTINE HAMOCC4BCM(kpie,kpje,kpke,kbnd,kplyear,kplmon,kplday,kldtday,&
                             pdlxp,pdlyp,pddpo,prho,pglat,omask,               &
-                            dust,rivin,ndep,pi_ph,                            &
+                            dust,rivin,ndep,oafx,pi_ph,                            &
                             pfswr,psicomo,ppao,pfu10,ptho,psao,               &
                             patmco2,pflxco2,pflxdms,patmbromo,pflxbromo)
 !******************************************************************************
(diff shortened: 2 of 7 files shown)


cime_config/buildnml
```diff
@@ -1853,6 +1861,11 @@ EOF
   if ($BLOM_N_DEPOSITION == TRUE) then
 cat >> $CASEBUILD/blom.input_data_list << EOF
 n_deposition_file = `echo $NDEPFILE | tr -d '"' | tr -d "'"`
+EOF
+  endif
+  if ($OALKFILE != "''") then
+cat >> $CASEBUILD/blom.input_data_list << EOF
+oafx_file = `echo $OAFILE | tr -d '"' | tr -d "'"`
```
TomasTorsvik:
```suggestion
oafx_file = `echo $OALKFILE | tr -d '"' | tr -d "'"`
```
I'm guessing this is a typo

hamocc/mo_read_oafx.F90
```diff
@@ -0,0 +1,277 @@
...
+!******************************************************************************
+  implicit none
+
+  private
+  public :: ini_read_oafx,get_oafx,oalkscen,oalkfile
+
+  real,allocatable, save :: oalkflx(:,:)
+   
+  character(len=128), save :: oalkscen=''
+  character(len=512), save :: oalkfile=''
```
TomasTorsvik, line 70: Looks like there is no routine for reading in alkalinity flux from a file. Have you thought to implement such a routine?

hamocc/mo_read_oafx.F90
```diff
@@ -0,0 +1,277 @@
...
+    oafx(:,:) = oalkflx(:,:)
+
+  else
+    
+    write(io_stdo_bgc,*) ''
+    write(io_stdo_bgc,*) 'get_oafx: invalid alkalinization scenario... '
+    call xchalt('(get_oafx)')
+    stop '(get_oafx)' 
+      
+  endif
```
TomasTorsvik, line 282: In the current implementation the simulation will stop if the `oalkscen` is not a valid string. Would you have a separate scenario label in case the if `oalkfile` is used for input, or do you have some other idea in mind?

TomasTorsvik (commented):
I think it could be useful to give feedback that the current implementation does not support a file input for alkalinity flux (assuming that I have not missed something).

Also, how to deal with the case if a user provides both an `oalkscen` and an `oalkfile` input, which one will take precedence? One could set a specific `oalkscen` string when using the input data file, then the user would have to set both of these.

JorgSchwinger:
Thanks for the feedback, Tomas. 

This is really only the framework for adding alkalinization scenarios. Some will need a dedicated input file, some (as the two simple schemes already there) will not need a file. Since we now always specify the full path name of input files (and HAMOCC does not know the base path to the input data directory) it is difficult to avoid the namelist input of a filename. Whether the file name is correct for a given scenario is (for now) up to the user. Later (if we want to make this available to a broader user community) this can be done through env_run.xml and the scripting system such that for a selected scenario the correct file is chosen.

hamocc/mo_read_oafx.F90
JorgSchwinger: A valid scenario label always needs to be provided. Whether a file has to be read depends on the given scenario. So if a scenario requires a file, this needs to be checked in the respective section.

hamocc/mo_read_oafx.F90
JorgSchwinger: Yes, not yet. See below.

cime_config/buildnml
JorgSchwinger: Yes, thanks for spotting this. Will correct.

TomasTorsvik (approved):

hamocc/mo_apply_oafx.F90
```diff
@@ -0,0 +1,105 @@
...
+  if (.not. do_oalk) return 
+
+  ! alkalinization in topmost layer 
+  do j=1,kpje
+  do i=1,kpie
+    if (omask(i,j).gt.0.5) then
+      ocetra(i,j,1,ialkali)=ocetra(i,j,1,ialkali)+oafx(i,j)*dtb/365./pddpo(i,j,1)
+#ifdef natDIC
+      ocetra(i,j,1,inatalkali)=ocetra(i,j,1,inatalkali)+oafx(i,j)*dtb/365./pddpo(i,j,1)
+#endif
```
tjiputra: I can't imagine an experiment where we'd want to see how the natural CO2 flux affected by oalk. Natural DIC/ALK tracers shouldn't be affected by anthropogenic perturbations, except for climate change. Maybe you have something else in mind.

hamocc/mo_read_oafx.F90
```diff
@@ -0,0 +1,277 @@
...
+      else
+        addalk_tot = addalk_0p56
+      endif
+    
+      ! Calculate alkalinity flux (kmol m^2 yr-1) to be applied
+      avflx = addalk_tot/ztotarea*Pmol2kmol
+      if(mnproc.eq.1) then
+        write(io_stdo_bgc,*)' '
+        write(io_stdo_bgc,*)' applying alkalinity flux of ', avflx, ' kmol m-2 yr-1'
+        write(io_stdo_bgc,*)'             over an area of ', ztotarea , ' m-2'
```
tjiputra: m2

tjiputra (approved):

hamocc/mo_apply_oafx.F90
JorgSchwinger: Thanks Jerry, you are right. One could think of a tracer that is affected by alkalinization but not by rising CO2, but I agree that the "natural" tracers should be reserved for the unperturbed state. I will remove this #natDIC section.

hamocc/mo_read_oafx.F90
```diff
@@ -0,0 +1,277 @@
...
+   
+      allocate(oalkflx(kpie,kpje),stat=errstat)
+      if(errstat.ne.0) stop 'not enough memory oalkflx'
+      oalkflx(:,:) = 0.0
+
+      ! Calculate total ocean area 
+      ztmp1(:,:)=0.0
+      do j=1,kpje
+      do i=1,kpie
+        if( omask(i,j).gt.0.5 .and. pglat(i,j)<70.0 .and. pglat(i,j)>-60.0 ) then
```
jmaerz: Maybe a stupid question, but why do you limit the ocean region here and below? - if it is a matter of the scenario (and the application area) I would still prefer not hard-coded areas - rather determined by some parameters (that one can change easier and consistently).

hamocc/mo_apply_oafx.F90
```diff
@@ -0,0 +1,102 @@
...
+  ! local variables 
+  integer :: i,j
+
+  if (.not. do_oalk) return 
+
+  ! alkalinization in topmost layer 
+  do j=1,kpje
+  do i=1,kpie
+    if (omask(i,j).gt.0.5) then
+      ocetra(i,j,1,ialkali)=ocetra(i,j,1,ialkali)+oafx(i,j)*dtb/365./pddpo(i,j,1)
```
jmaerz, line 92: Maybe more of a general question: do we have a variable for the year length (in days)? - if so, it could be used here.

hamocc/hamocc4bcm.F90
```diff
@@ -66,6 +66,7 @@ SUBROUTINE HAMOCC4BCM(kpie,kpje,kpke,kbnd,kplyear,kplmon,kplday,kldtday,&
 !  *REAL*    *dust*       - dust deposition flux [kg/m2/month].
 !  *REAL*    *rivin*      - riverine input [kmol m-2 yr-2].
 !  *REAL*    *ndep*       - nitrogen deposition [kmol m-2 yr-2].
+!  *REAL*    *oaflx*      - alkalinity flux from alkalinization [kmol m-2 yr-2]
```
jmaerz: Maybe I miss something, but typos in the rivin, ndep, oaflx units (yr-1 ?!)?

jmaerz (approved):
Approved from my side.

hamocc/mo_read_oafx.F90
JorgSchwinger: That's the CDRMIP protocol (assuming that the ocean is mostly ice-free between 60S and 70N). I have changed this to be parameters in the module header.

hamocc/hamocc4bcm.F90
JorgSchwinger: Corrected

hamocc/mo_apply_oafx.F90
JorgSchwinger: No, we don't have that. We don't even have a module where such parameters would fit in naturally  (both mo_control and mo_param1 are not really the place for general purpose parameters). Should we make a new on (e.g. mo_param0)?

jmaerz: I guess, I was also thinking about leap years (which might have been defined in BLOM)  and/or other setups for more crazy applications (longer/shorter days, years, ...). Technically, it would be good to have such module, however, I see for time-related parameters the need to keep these manageable via the driving model (aka BLOM for the most time). Other than that hamocc_init is currently the place, where I would store/define it. But nevermind, it was a really general question.

JorgSchwinger: I agree, with respect to calendar related things we would need to pass this from BLOM to HAMOCC via hamocc_ini (although I'm not sure whether we would ever run NorESM or BLOM on a calendar other than 'noleap'). I'm closing and merging this now - thanks for the thorough reviews!

PR #189, opened 2022-09-02
Lon-lat variable sediment porosity

jmaerz:
This feature branch enables to read a sediment porosity file as input for the sediment compartment of iHAMOCC. iHAMOCC then considers the spatially variable porosity through the scaling of the sediment pore water diffusion coefficients and in sediment processes. This feature can thus far only be switched on via the namelist switch `l_sed_por` for expert users (current default is `.false.`). 

The reading of a sediment porosity file is thus far untested (and will produce results different from CMIP6), while the current default version with `l_sed_por=.false.` produces bit identical results compared to the former master branch. I am working on the input file for this feature, but wanted to give you already the possibility to review the code changes.


jmaerz:
Ok, I just realize that the commit comes with all changes done in master over the past days (as compared to this development branch). Maybe it would be good to first merge the master into the development branch?

TomasTorsvik:
> Ok, I just realize that the commit comes with all changes done in master over the past days (as compared to this development branch). Maybe it would be good to first merge the master into the development branch?

Yes, probably that is a good idea, although this will create an additional merge commit on the development branch. Don't know if there is any better alternative though.

UPDATE: I created a PR for merging `master` to `feature-hamocc_beyond-CMIP6` #190

JorgSchwinger:
I'll wait with my review until #190 has been merged...

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , I briefly discussed with @TomasTorsvik about the history issue and tested out, whether an initial merge with the `feature-hamocc_beyond-CMIP6` branch would solve the history issue (**it does**). Given our common agreement on still merging features into master that do not break backward compatibility, I am now moving the pull request to the `master` branch. I will check again with regards to the loop order (thanks @JorgSchwinger).

hamocc/mo_sedmnt.F90
```diff
@@ -70,13 +70,19 @@ MODULE mo_sedmnt
       REAL, save :: dzs(ksp)    = 0.0
       REAL, save :: seddzi(ksp) = 0.0
       REAL, save :: seddw(ks)   = 0.0
-      REAL, save :: porsol(ks)  = 0.0
-      REAL, save :: porwah(ks)  = 0.0
-      REAL, save :: porwat(ks)  = 0.0
+      !REAL, save :: porsol(ks)  = 0.0
+      !REAL, save :: porwah(ks)  = 0.0
+      !REAL, save :: porwat(ks)  = 0.0
```
TomasTorsvik: Maybe these lines should be deleted instead of commented out? With the modified code theses lines can not be re-enabled by un-commenting.

hamocc/bodensed.F90
```diff
@@ -79,33 +82,60 @@ subroutine bodensed(kpie,kpje,kpke,pddpo)
...
+     enddo
   enddo
+  
+  sedict = 1.e-9 * dtbgc ! Moecular diffusion coefficient
+  ! Dissolution rate constant of opal (disso) [1/(kmol Si(OH)4/m3)*1/sec]
+  ! THIS NEEDS TO BE CHANGED TO disso=3.e-8! THIS IS ONLY KEPT FOR THE MOMENT
+  ! FOR BACKWARDS COMPATIBILITY
+  !disso=3.e-8  ! (2011-01-04) EMR
+  !disso=1.e-6 ! test vom 03.03.04 half live sil ca. 20.000 yr 
+  disso_sil = 1.e-6*dtbgc
```
TomasTorsvik:
```suggestion
  ! Dissolution rate constant of opal (disso) [1/(kmol Si(OH)4/m3)*1/sec]
  ! THIS NEEDS TO BE CHANGED TO disso=3.e-8! THIS IS ONLY KEPT FOR THE MOMENT
  ! FOR BACKWARDS COMPATIBILITY
  !disso_sil = 3.e-8*dtbgc  ! (2011-01-04) EMR
  !disso_sil = 1.e-6*dtbgc  ! test vom 03.03.04 half live sil ca. 20.000 yr 
  disso_sil = 1.e-6*dtbgc
```
Maybe a bit pedantic, but I think it is best if the commented lines also reflect the code change.

hamocc/bodensed.F90
```diff
@@ -79,33 +82,60 @@ subroutine bodensed(kpie,kpje,kpke,pddpo)
...
+     do j = 1, kpje
+     do i = 1, kpie
+        porsol(i,j,k) = 1. - porwat(i,j,k)
+        if(k >= 2) porwah(i,j,k) = 0.5 * (porwat(i,j,k) + porwat(i,j,k-1))
+        if(k == 1) porwah(i,j,k) = 0.5 * (1. + porwat(i,j,1))
+     enddo
+     enddo
   enddo
+  
+  sedict = 1.e-9 * dtbgc ! Moecular diffusion coefficient
```
TomasTorsvik:
```suggestion
  sedict = 1.e-9 * dtbgc ! Molecular diffusion coefficient
```
typo

hamocc/bodensed.F90
```diff
@@ -79,33 +82,60 @@ subroutine bodensed(kpie,kpje,kpke,pddpo)
...
+  ! THIS NEEDS TO BE CHANGED TO disso=3.e-8! THIS IS ONLY KEPT FOR THE MOMENT
+  ! FOR BACKWARDS COMPATIBILITY
+  !disso=3.e-8  ! (2011-01-04) EMR
+  !disso=1.e-6 ! test vom 03.03.04 half live sil ca. 20.000 yr 
+  disso_sil = 1.e-6*dtbgc
+  ! Silicate saturation concentration is 1 mol/m3
+  silsat    = 0.001
+
+  ! Degradation rate constant of POP (disso) [1/(kmol O2/m3)*1/sec]
+  disso_poc = 0.01 / 86400. * dtbgc  !  disso=3.e-5 was quite high
```
TomasTorsvik, line 139:
```suggestion
  disso_poc = 0.01 / 86400. * dtbgc
```
Is this comment relevant, do we need to keep it around?

hamocc/bodensed.F90
```diff
@@ -79,33 +82,60 @@ subroutine bodensed(kpie,kpje,kpke,pddpo)
...
 
-  sedict = 1.e-9 * dtbgc
+  ! Denitrification rate constant of POP (disso) [1/sec]
+  sed_denit =  0.01/86400. * dtbgc !ik      denit = 1.e-6*dtbgc
```
TomasTorsvik:
```suggestion
  sed_denit =  0.01/86400. * dtbgc
```
Same as above

hamocc/sedshi.F90
```diff
@@ -219,7 +219,7 @@ SUBROUTINE SEDSHI(kpie,kpje,omask)
 
 ! fill the last active layer
         refill=seddef/(buried+1.e-10) 
-        frac = porsol(ks)*seddw(ks) !changed k to ks, ik
+        frac = porsol(i,j,ks)*seddw(ks) !changed k to ks, ik
```
TomasTorsvik:
```suggestion
        frac = porsol(i,j,ks)*seddw(ks)
```
another comment that doesn't seem to add much info

TomasTorsvik (approved):
This looks fine to me. I wonder if we should use the opportunity to clean up some of the comments that seem outdated.

hamocc/mo_sedmnt.F90
jmaerz: Right. I missed to clean that. I'll do this.

hamocc/bodensed.F90
jmaerz: Correct, I can do this.

hamocc/bodensed.F90
jmaerz: I can't tell - maybe @JorgSchwinger made some experiences that allow for better judgement?

hamocc/bodensed.F90
jmaerz: @JorgSchwinger - any reason to keep the comment?

hamocc/sedshi.F90
jmaerz: I can delete it, sure.

hamocc/mo_sedmnt.F90
jmaerz: Done.

hamocc/bodensed.F90
jmaerz: Done.

hamocc/bodensed.F90
jmaerz: Done.

cime_config/buildnml
```diff
@@ -237,6 +237,9 @@ set OALKSCEN = "''"
 set OALKFILE = "''"
 set WITH_DMSPH = .false.
 set PI_PH_FILE = "''"
+set L_SED_POR = .false.
```
JorgSchwinger: Maybe a more suggestive name would be preferable? For example L_SEDPOR_2D to indicate that this activates a 2d-spatially variable sediment porosity?

cime_config/buildnml
```diff
@@ -1470,6 +1479,8 @@ cat >>! $RUNDIR/ocn_in$inststr << EOF
   INID14C      = $INID14C
   WITH_DMSPH   = $WITH_DMSPH
   PI_PH_FILE   = $PI_PH_FILE
+  L_SED_POR    = $L_SED_POR
+  SEDPORFILE   = $SEDPORFILE
 /
```
JorgSchwinger, line 1485: GitHub won't let me add a comment in an untouched part of the code, but further below at line 1840 there is a file `blom.input_data_list` created. If L_SED_POR is activated and SEDPORFILE specified, the filename should be added to this file (used by the CESM scripting system to check at submission whether all files are available)

cime_config/buildnml
jmaerz: I agree. Thanks for the suggestion! Maybe `l_3Dvarsedpor` (variable sediment porosity)?!

cime_config/buildnml
jmaerz: I see. Thanks for the explanation! I'll add that.

hamocc/bodensed.F90
```diff
@@ -79,33 +82,60 @@ subroutine bodensed(kpie,kpje,kpke,pddpo)
...
-  porwat(7) = 0.73
-  porwat(8) = 0.7
-  porwat(9) = 0.68
-  porwat(10) = 0.66
-  porwat(11) = 0.64
-  porwat(12) = 0.62
+  ! this initialization can be done later via reading a porosity map
+  if (l_sed_por)then
+   ! lon-lat variable sediment porosity from input file
+   call apply_sedpor(kpie,kpje,ks,omask,sed_por)
```
JorgSchwinger: I find it a bit overkill to define an extra module for ~10 lines of code (essentially porwat=sedpor) that could be just placed directly here.

jmaerz: :-) - agree - I was debating with myself - I'll change it.

hamocc/mo_read_sedpor.F90
```diff
@@ -0,0 +1,130 @@
...
+
+public :: ini_read_sedpor,sedporfile,sed_por
+
+character(len=512),save :: sedporfile = ''
+real,save,allocatable   :: sed_por(:,:,:) ! holds input data from nc-file 
+
+
+contains
+
+subroutine ini_read_sedpor(kpie,kpje,ks,omask)
```
JorgSchwinger: Actually, there is no need to allocate the array sed_por permanently: The array is only needed to transfer the data from this routine to bodensed (both called from hamocc_ini). So one could define sed_por as a local (automatic) array in hamocc_ini, pass it first to read_sedpor and then to bodensed. On return from hamocc_ini, sed_por will be automatically deallocated. I would then also rename this routine to `read_sedpor` because then there is nothing that is really initialised (this would be more consistent with the fact that this is a one time read operation, not something that is used later on).

On a side note, one could integrate `bodensed` into `mo_sedmnt` and rename it to `ini_sedmnt`, which would make more sense than having it as a stand-alone subroutine. Maybe something for later.

hamocc/bodensed.F90
JorgSchwinger: No, I don't think so, too.

jmaerz: Removed.

cime_config/buildnml
jmaerz: Done

cime_config/buildnml
jmaerz: Done

JorgSchwinger (approved):

hamocc/mo_read_sedpor.F90
jmaerz: Hi @JorgSchwinger , I am doing the first paragraph. With respect to the second part: I like the idea, was shortly about to do it already before. I would like to do this as another pull request, though, to minimize conflicts with my own developments. I'll move the subroutine soon.

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , I adjusted the code wrt to Jörgs comments and merged with current `master` to resolve upcoming conflicts (due to previous changes to the `powtra -> ocetra` mapping). The only thing I haven't changed thus far is the comment from Tomas: https://github.com/NorESMhub/BLOM/pull/189#discussion_r962732336 which just concerns a comment in the code. Do you want to re-review the pull request or shall I merge into `master`?

TomasTorsvik:
> The only thing I haven't changed thus far is the comment from Tomas: [#189 (comment)](https://github.com/NorESMhub/BLOM/pull/189#discussion_r962732336) which just concerns a comment in the code.

The comment is not essential. From my point of view you can merge it as it is into `master`.

JorgSchwinger:
Yes, I agree, please go ahead

PR #192, opened 2022-09-06
Dynamic mapping of pore water tracers to ocean tracers

### LINKED ISSUE #191: Recode mapping between pore water and ocean tracers
Hi @JorgSchwinger and @TomasTorsvik , in the course of the implementation for the nitrogen cycle, I need to introduce new pore water tracers for NH<sub>4</sub> and NO<sub>2</sub>. Thus far, most pore water tracer indices are matching the ocean tracer indices. Through the use of pre-processor flags, other implemented tracers (solid and dissolved ocean tracers, etc.), etc. this scheme is no longer sustainable and I would like to introduce a mapping between pore water tracers and ocean tracers. This primarily affects `mo_param1_bgc.F90` and `dipowa.F90`.

My current thought is to implement something along the lines (in `mo_param1_bgc.F90`, sort of pseudo-code):
```
int map_por2octra(-1:npowtra)
map_por2octra(ipowaox) = ioxygen
```
and then in `dipowa.F90` using the mapping vector instead of the direct index for the sediment tracer, roughly:
Instead of `iv_oc = iv` using `iv_oc = map_por2octra(iv)` (for most of the tracers that are directly handled via the tredsy).

I will implement this and test this in the master branch (since it shouldn't change the response). If you have a better suggestion how to handle the mapping, please let me know (asap, since I need this part of code soon in the nitrogen cycle). In the course of this, I will also modify (re-order) the sediment tracer index part in `mo_param1_bgc.F90` and make its style similar to the ocean tracer index declaration (and initialization).

jmaerz:
As announced in #191, I modified the way how we map the pore water tracers to the ocean tracers during diffusive exchange between sediment and ocean. I left the (potential) bug untouched, which will be fixed in a separate step. I tested the new implementation against former `master` and results were bit-identical.

### DIFF
--- a/hamocc/dipowa.F90
+++ b/hamocc/dipowa.F90
@@ -198,14 +198,15 @@ subroutine dipowa(kpie,kpje,kpke,omask,lspin)
 
            ! diffusive fluxes (positive downward)
            sedfluxo(i,j,iv) = sedfluxo(i,j,iv)                                 &
-                &  -(ocetra(i,j,kbo(i,j),iv) - aprior)* bolay(i,j)
+                &  -(ocetra(i,j,kbo(i,j),iv_oc) - aprior)* bolay(i,j)
 #ifdef natDIC
-           if (iv==isco212) ocetra(i,j,kbo(i,j),inatsco212) =                  &
+           ! workaround as long as natDIC is not implemented throughout the sediment module 
+           if (iv_oc==isco212) ocetra(i,j,kbo(i,j),inatsco212) =               &
                 &  ocetra(i,j,kbo(i,j),inatsco212) +                           &
-                &  ocetra(i,j,kbo(i,j),iv) - aprior
-           if (iv==ialkali) ocetra(i,j,kbo(i,j),inatalkali) =                  &
+                &  ocetra(i,j,kbo(i,j),isco212) - aprior
+           if (iv_oc==ialkali) ocetra(i,j,kbo(i,j),inatalkali) =               &
                 &  ocetra(i,j,kbo(i,j),inatalkali) +                           &
-                &  ocetra(i,j,kbo(i,j),iv) - aprior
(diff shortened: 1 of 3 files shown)


TomasTorsvik (approved):
Looks fine to me.

hamocc/dipowa.F90
```diff
@@ -190,16 +186,8 @@ subroutine dipowa(kpie,kpje,kpke,omask,lspin)
...
   do iv = 1, npowtra
-     iv_oc = iv
-#ifdef cisonew
-     if (iv == ipowc13) iv_oc=isco213
-     if (iv == ipowc14) iv_oc=isco214
-#endif
+     iv_oc = map_por2octra(iv)
      do i = 1,kpie
         l = 0
         if (omask(i,j) > 0.5) then
```
JorgSchwinger, line 193: If the correspondence between pore water and ocean tracers is not guaranteed, iv needs to be replaced by iv_oc in all ocetra occurances below. Also inside the #ifdef natDIC block

JorgSchwinger (commented):
Looks good to me, but should be thoroughly tested for bit-for-bit reproducibility!

hamocc/dipowa.F90
jmaerz: See https://github.com/NorESMhub/BLOM/issues/191#issuecomment-1238443457 - since it seems as it doesn't affect CMIP backwards compatibility at all, I would then suggest to do it in one goal and exchange the indices. The code as is was tested against the former master, but maybe it's then worth to test it also against master `cisonew` switched on.

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , I checked the latest commit of this pull request against the master and master switched on for natDIC. In both cases, results are identical. So, if you agree, I would merge with master.

JorgSchwinger (approved):

PR #198, opened 2022-09-26
Fixing variable sediment porosity - field initialization in case of `sedbypass=true`

### LINKED ISSUE #197: Fix required for spatially variable porosity for case `sedbypass=true`
Hi @TomasTorsvik and @JorgSchwinger , unfortunately, I overlooked the case  `sedbypass=true` when implementing the spatially variable porosity. Thus far, we ONLY allocate memory for `porsol`, `porwah`, `porwat`, `solfu` and the diffusion coefficients, when `sedbypass=false`. However, during the initialization process, `bodensed.F90` is called irrespective of this and we are trying to fill un-allocated fields, when `sedbypass=true`. Formerly, these fields were 1D and always allocated, while we actually don't need them really, when `sedbypass=true`. 

So, (quick) question: shall we always allocate the 2D and 3D fields or do you feel it is better to only fill them in `bodensed.F90` during the initialization, when needed (meaning: we ignore filling them, when `sedbypass=true` via pre-processor flags in `bodensed.F90`). I see the trade-off between memory usage versus readability here... - which is why I am asking.

jmaerz:
Dear @JorgSchwinger and @TomasTorsvik , as noted in #197, the initialization of 2D and 3D fields in `bodensed.F90` raised an error in case of `sedbypass=true`. As discussed in #197 (and before), I moved the initialization of sediment parameters (some of them also used in the water column in case of aggregation schemes) into `mo_sedmnt.F90` and initialize the 2D and 3D fields only in case we're using the sediment routines. Closes #197


TomasTorsvik (approved):
Thanks @jmaerz , this looks fine to me.

I see one more reference to BODENSED in the header of dipowa.F90, line 43, that you may have missed
```
!     constant diffusion coefficient : 1.e-9 set in BODENSED.
```

jmaerz:
Thanks @TomasTorsvik ! I fixed it.

hamocc/mo_sedmnt.F90
```diff
@@ -97,16 +105,185 @@ MODULE mo_sedmnt
...
+      enddo
+      enddo
+      enddo
+      zcoeflo(:,:,ks) = 0.0                    ! diffusion coefficient for bottom sediment layer
       
+      if (mnproc == 1) then
+        write(io_stdo_bgc,*)  'Pore water diffusion coefficients in sediment initialized'
+      endif
+
+ END SUBROUTINE ini_sedmnt_fields
```
JorgSchwinger: Wouldn't a more specific name be better, e.g. `ini_sedmnt_por` or something like this? `ini_sedmnt_fields` makes me wonder (if I were new to the code) what the difference between this routine and `alloc_mem_sedmnt` is.

JorgSchwinger (approved):
Looks good, thank you. Have you tested whether this gives bit-identical results?

hamocc/mo_sedmnt.F90
jmaerz: Hi @JorgSchwinger , mmh, ok, I would have hoped that the difference between declaration, allocation and initialization should be clear - not sure, how to best name it then, though, since it's about pore water space, solid volume fraction and diffusion coefficients in the sediment (ini_sedmnt_vars also sounds odd, which would follow the beleg_vars.F90 naming). I'll think about it. I haven't tested it after the re-shuffling (since nothing else was changed), but I can/will set up a run on betzy.

JorgSchwinger: Ok, I agree the difference between 'ini' and 'alloc' shoud be clear (allthough the former could include the latter). I just thought that the routine is dealing with initializing sediment porosity (and related/derived quantities) so '_por'  or similar instead of '_fields' would be more specific. But not so important after all.

jmaerz: Ok, I renamed it to `_por` and provide a bit more information what will be initialized.

jmaerz: Hi @TomasTorsvik and @JorgSchwinger , I now checked the feature branch against the `master` - in case of `sedbypass=true`, the latest `master` didn't went through - so there the bug caused issues (sorry again) - so don't use sha: 88d6208d053733a87dec9546be6a2287f8626941 in cases for `sedbypass=true` - I re-ran the `master` with sha: 3e4eb0f4543175dce3b92a070ea454cdfec1f645 (the commit before introducing the variable porosity) for testing purposes and major nutrients and alkalinity are identical in both cases (`sebypass=true` and `sedbypass=false`). So I'll merge the branch.

PR #199, opened 2022-09-27
Support for NUOPC driver

matsbn:
BLOM can now be coupled to CESM using the NUOPC driver. Also, wave forcing fields have been declared and wave-ocean coupling is supported when using the NUOPC driver.

Technical functioning of the NUOPC interface has been verified with cesm2_3_alpha08a. There is no change of results in BLOM standalone configurations or when the MCT driver is used.


phy/mod_difest.F
```diff
@@ -283,6 +285,15 @@ subroutine init_difest
 c    .          lnoDGat1=.true.                 ,
 c    .          CVMix_kpp_params_user=KPP_params )
 c
+c$OMP PARALLEL DO PRIVATE(l,i)
+      do j=1,jj
+        do l=1,isp(j)
+        do i=max(1,ifp(j,l)),min(ii,ilp(j,l))
+          OBLdepth(i,j)=10.
```
TomasTorsvik, line 292: Why is this needed in init_difest? I see a reference to OBLdepth in the `blom_accflds` routine in `mod_nuopc_methods`. I'm slightly worried that `hOBL` is not initialized in a similar way (see recent changes in PR #179 ), but it is not clear to me if this can have implications for iHAMOCC.

TomasTorsvik: OK, I just saw your latest comment to PR #179. I will include an inital value `hOBL(i,j) = 3.` in a similar way as for OBLdepth.

TomasTorsvik (approved):
Hi @matsbn ,
I should probably have a better grasp of NUOPC to make a full review of this PR, but at least I don't see anything obvious that looks strange to me.

phy/mod_difest.F
matsbn: OBLdepth is passed to the coupler with NUOPC, while it was not with MCT. OBLdepth is needed in case the wave component is active. The export coupling fields have to be sent before the model has started integrating, so before difest_vertical* is called. Therefore it is necessary to give OBLdepth a sensible value in the case vcoord_type_tag == cntiso_hybrid .

JorgSchwinger (approved):
Looks fine to me, although I can't really say much more than this.

matsbn:
Thanks for approving, @TomasTorsvik and @JorgSchwinger . This is a change that the BLOM devs have been informed about, and I thought it as useful to inform the iHAMOCC devs through a review request since it changes the folder structure under ```drivers```. As mention in the PR, it should not impact current application of BLOM in NorESM. I'll proceed to merge.

PR #201, opened 2022-11-08
Fix porosity read

jmaerz:
In preparation of the new NorESM release, I fixed the reading of the variable sediment porosity so that it is now fully functional. As written in #189 , the reading was thus far untested and has been now tested and applied in my private N-cycle branches (from which I cherry-picked the fix). For non-variable sediment porosity (the default), nothing changes.


JorgSchwinger (approved):
Looks good to me.

TomasTorsvik (approved):
Thanks! I'm fine with this code change.

PR #202, opened 2022-11-18
pCO2, Piston velocity and solubility output

### LINKED ISSUE #31: Add output for actually used atmopheric pCO2 and pCO2 water vapour correction
Atmospheric CO2 mixing ratio in the gas exchange formulation is modified by sea level pressure and a water vapour correction in the gas exchange formulation, but these are not in the output, so the true Delta pCO2 used cannot be traced.

jmaerz:
Hi Tomas, hi Jörg, 

I added output for pCO2 under moist air assumption (which can actually make quite some difference - up to ca. O(10ppm)), pure piston velocity, and solubility under dry and moist air assumption. There are two things to note, though:
1. the former `kwco2` now really only holds the Piston velocity (not times solubility anymore) - Piston times solubility (moist air, as before) is available via `kwco2kh` now. Does that change need some changes in any cmorization packages (which I am currently not fully aware of)?
2. I am uncertain, why `kwco2sol(i,j) = kwco2*Kh*1e-6`  (WHY 1e-6???, carchm l. 533) From my understanding, `kwco2` has units of m/s and `Kh` mol/kg/atm - so to which units were/is it converted to? Should it be per µatm?

Another thing might be, if we want to keep track of the air-pressure correction `rpp0` used for calculating CO2 fluxes (which affects the gas saturation).

Closes #31

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -424,7 +424,11 @@ set SRF_EXPORT    = '0, 2, 2'
 set SRF_EXPOSI    = '0, 2, 2' 
 set SRF_EXPOCA    = '0, 2, 2' 
 set SRF_KWCO2     = '0, 2, 2' 
+set SRF_KWCO2KHD  = '0, 2, 2' 
+set SRF_CO2KHD    = '0, 2, 2' 
+set SRF_CO2KH     = '0, 2, 2' 
 set SRF_PCO2      = '0, 2, 2' 
+set SRF_PCO2M     = '0, 2, 2' 
 set SRF_CO2FXD    = '4, 2, 2' 
 set SRF_CO2FXU    = '4, 2, 2' 
 set SRF_OXFLUX    = '0, 2, 2' 
@@ -1578,7 +1582,11 @@ cat >>! $RUNDIR/ocn_in$inststr << EOF
 !   EXPOCA         - Ca export production (epcalc100) [mol Ca m-2 s-1]
 !   EXPOSI         - Si export production (epsi100) [mol Si m-2 s-1]
 !   PCO2           - Surface PCO2 (spco2) [uatm]
-!   KWCO2          - kwco2 x solubility
+!   PCO2M          - Surface PCO2 under moist air assumption [uatm]
+!   KWCO2          - kwco2 
+!   KWCO2KHD       - kwco2 x solubility (dry air)
+!   CO2KHD         - khd CO2 solubility under dry air assumption 
+!   CO2KH          - kh CO2 solubility under moist air assumption 
 !   CO2FXD         - Downward CO2 flux (co2fxd) [kg C m-2 s-1]
 !   CO2FXU         - Upward CO2 flux (co2fxu) [kg C m-2 s-1]
 !   NIFLUX         - Nitrogen flux (fgn2) [mol N2 m-2 s-1]
@@ -1658,7 +1666,11 @@ cat >>! $RUNDIR/ocn_in$inststr << EOF
   SRF_EXPOSI    = $SRF_EXPOSI
   SRF_EXPOCA    = $SRF_EXPOCA
   SRF_KWCO2     = $SRF_KWCO2
+  SRF_KWCO2KHD  = $SRF_KWCO2KHD
+  SRF_CO2KHD    = $SRF_CO2KHD
+  SRF_CO2KH     = $SRF_CO2KH
   SRF_PCO2      = $SRF_PCO2
+  SRF_PCO2M     = $SRF_PCO2M
   SRF_CO2FXD    = $SRF_CO2FXD
   SRF_CO2FXU    = $SRF_CO2FXU
   SRF_OXFLUX    = $SRF_OXFLUX
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -1583,10 +1583,10 @@ cat >>! $RUNDIR/ocn_in$inststr << EOF
 !   EXPOSI         - Si export production (epsi100) [mol Si m-2 s-1]
 !   PCO2           - Surface PCO2 (spco2) [uatm]
 !   PCO2M          - Surface PCO2 under moist air assumption [uatm]
-!   KWCO2          - kwco2 
-!   KWCO2KHM       - kwco2 x solubility (moist air)
-!   CO2KH          - khd CO2 solubility under dry air assumption 
-!   CO2KHM         - kh CO2 solubility under moist air assumption 
+!   KWCO2          - Piston velocity (kwco2)  [m s-1]
+!   KWCO2KHM       - Piston velocity times solubility (kwco2*kh; moist air) [m s-1 mol kg-1 uatm-1]
+!   CO2KH          - CO2 solubility under dry air assumption (khd) [mol kg-1 atm-1] 
+!   CO2KHM         - CO2 solubility under moist air assumption (kh) [mol kg-1 atm-1] 
 !   CO2FXD         - Downward CO2 flux (co2fxd) [kg C m-2 s-1]
 !   CO2FXU         - Upward CO2 flux (co2fxu) [kg C m-2 s-1]
 !   NIFLUX         - Nitrogen flux (fgn2) [mol N2 m-2 s-1]
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -424,9 +424,9 @@ set SRF_EXPORT    = '0, 2, 2'
 set SRF_EXPOSI    = '0, 2, 2' 
 set SRF_EXPOCA    = '0, 2, 2' 
 set SRF_KWCO2     = '0, 2, 2' 
-set SRF_KWCO2KH   = '0, 2, 2' 
-set SRF_CO2KHD    = '0, 2, 2' 
+set SRF_KWCO2KHM  = '0, 2, 2' 
 set SRF_CO2KH     = '0, 2, 2' 
+set SRF_CO2KHM    = '0, 2, 2' 
 set SRF_PCO2      = '0, 2, 2' 
 set SRF_PCO2M     = '0, 2, 2' 
 set SRF_CO2FXD    = '4, 2, 2' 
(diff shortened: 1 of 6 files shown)


hamocc/carchm.F90
```diff
@@ -180,7 +181,11 @@ SUBROUTINE CARCHM(kpie,kpje,kpke,kbnd,                                  &
        co214fxd (:,:)=0.
        co214fxu (:,:)=0.
 #endif
-       pco2d    (:,:)=0. 
+       pco2d    (:,:)=0.
+       pco2m    (:,:)=0.
+       kwco2d    (:,:)=0.
```
TomasTorsvik:
```suggestion
       kwco2d   (:,:)=0.
```

hamocc/ncout_hamocc.F90
```diff
@@ -1119,6 +1133,7 @@ subroutine hamoccvardef(iogrp,timeunits,calendar,cmpflg)
        &   nctime,ncfcls,ncedef,ncdefvar3d,ndouble
 
   use mo_bgcmean, only: srf_kwco2,srf_pco2,srf_dmsflux,srf_co2fxd,              &
+       &   srf_kwco2kh,srf_co2khd,srf_co2kh,srf_pco2m,                         &
```
TomasTorsvik:
```suggestion
       &   srf_kwco2kh,srf_co2khd,srf_co2kh,srf_pco2m,                          &
```

hamocc/ncout_hamocc.F90
```diff
@@ -1195,9 +1210,17 @@ subroutine hamoccvardef(iogrp,timeunits,calendar,cmpflg)
...
+       &   'kwco2','CO2 piston velocity',' ','m s-1',0)
+  call ncdefvar3d(SRF_KWCO2KH(iogrp),cmpflg,'p',                                  &
+       &   'kwco2kh','CO2 piston velocity times solubility (moist air)',' ','m s-1 mol kg-1 atm-1',0)
+  call ncdefvar3d(SRF_CO2KHD(iogrp),cmpflg,'p',                                  &
+       &   'co2khd','CO2 solubility (dry air)',' ','mol kg-1 atm-1',0)
+  call ncdefvar3d(SRF_CO2KH(iogrp),cmpflg,'p',                                  &
+       &   'co2kh','CO2 solubility (moist air)',' ','mol kg-1 atm-1',0)
   call ncdefvar3d(SRF_PCO2(iogrp),cmpflg,'p',                                   &
        &   'pco2','Surface PCO2',' ','uatm',0)
+  call ncdefvar3d(SRF_PCO2M(iogrp),cmpflg,'p',                                   &
```
TomasTorsvik:
```suggestion
  call ncdefvar3d(SRF_PCO2M(iogrp),cmpflg,'p',                                  &
```

hamocc/ncout_hamocc.F90
```diff
@@ -1195,9 +1210,17 @@ subroutine hamoccvardef(iogrp,timeunits,calendar,cmpflg)
   call ncattr('bounds','depth_bnds')
   call ncdefvar('depth_bnds','bounds depth',ndouble,8)
   call ncdefvar3d(SRF_KWCO2(iogrp),cmpflg,'p',                                  &
-       &   'kwco2',' ',' ',' ',0)
+       &   'kwco2','CO2 piston velocity',' ','m s-1',0)
+  call ncdefvar3d(SRF_KWCO2KH(iogrp),cmpflg,'p',                                  &
+       &   'kwco2kh','CO2 piston velocity times solubility (moist air)',' ','m s-1 mol kg-1 atm-1',0)
+  call ncdefvar3d(SRF_CO2KHD(iogrp),cmpflg,'p',                                  &
```
TomasTorsvik:
```suggestion
  call ncdefvar3d(SRF_CO2KHD(iogrp),cmpflg,'p',                                 &
```

hamocc/ncout_hamocc.F90
```diff
@@ -1195,9 +1210,17 @@ subroutine hamoccvardef(iogrp,timeunits,calendar,cmpflg)
   call ncattr('bounds','depth_bnds')
   call ncdefvar('depth_bnds','bounds depth',ndouble,8)
   call ncdefvar3d(SRF_KWCO2(iogrp),cmpflg,'p',                                  &
-       &   'kwco2',' ',' ',' ',0)
+       &   'kwco2','CO2 piston velocity',' ','m s-1',0)
+  call ncdefvar3d(SRF_KWCO2KH(iogrp),cmpflg,'p',                                  &
```
TomasTorsvik:
```suggestion
  call ncdefvar3d(SRF_KWCO2KH(iogrp),cmpflg,'p',                                &
```

TomasTorsvik (approved):
Thanks, Jöran! This looks fine to me. I have a few pedantic comments about code alignment, if you could take a moment to look at it.

cime_config/buildnml
```diff
@@ -1578,7 +1582,11 @@ cat >>! $RUNDIR/ocn_in$inststr << EOF
 !   EXPOCA         - Ca export production (epcalc100) [mol Ca m-2 s-1]
 !   EXPOSI         - Si export production (epsi100) [mol Si m-2 s-1]
 !   PCO2           - Surface PCO2 (spco2) [uatm]
-!   KWCO2          - kwco2 x solubility
+!   PCO2M          - Surface PCO2 under moist air assumption [uatm]
+!   KWCO2          - kwco2 
+!   KWCO2KH        - kwco2 x solubility (moist air)
+!   CO2KHD         - khd CO2 solubility under dry air assumption 
+!   CO2KH          - kh CO2 solubility under moist air assumption 
 !   CO2FXD         - Downward CO2 flux (co2fxd) [kg C m-2 s-1]
```
JorgSchwinger, line 1590: Regarding the naming of of these variables, I find it confusing that sometimes a "D" for dry is appended, and sometimes a "M" for moist, and sometimes nothing. I would suggest to use only "M" to indicate the moist air assumption:

!   PCO2M            - Surface PCO2 under moist air assumption [uatm]
!   KWCO2            - kwco2 
!   KWCO2KHM    - kwco2 x solubility (moist air)
!   CO2KH             - khd CO2 solubility under dry air assumption 
!   CO2KHM          - kh CO2 solubility under moist air assumption 

I see that the idea was to use names that are close to the values in the code, but in the namelist, it should make sense for users that have no insights in the code. To resolve this one could also rename kh -> khm and khd -> kh in carchm, but not sure if this would make sense

jmaerz: I can do this. One question though was/is the issue with previous names and potential confusion about it during subsequent cmorization steps. Any opinion/insights/knowledge about this? I tried to keep the former names, if possible.

JorgSchwinger: What is important is that PCO2 is not changed (since this was/is of course used in cmorization). Everything else is not CMIP6 output. So the only potential confusion is that KWCO2 will change its meaning, but this was actually very confusing before, so this is an improvement (not backwards-compatible though).

TomasTorsvik: I think this should still be fine, but we should remember to mention the change in KWCO2 in the CHANGELOG (for both BLOM/iHAMOCC-1.3.0 and NorESM-2.0.6)

JorgSchwinger:
And yes, the 1e-6 (or 1e6) converts atm to microatm, I think.

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , I tried to adopt your suggestions with the last commit.

hamocc/carchm.F90
```diff
@@ -518,13 +523,17 @@ SUBROUTINE CARCHM(kpie,kpje,kpke,kbnd,                                  &
...
+       pco2m(i,j) = cu * 1.e6 / Kh
 #ifdef natDIC
        natpco2d(i,j) = natcu * 1.e6 / Khd
 #endif
 
 ! Save product of piston velocity and solubility for output
-       kwco2sol(i,j) = kwco2*Kh*1e-6
-
+       kwco2sol(i,j) = kwco2*Kh*1e-6 !m/s mol/kg/muatm 
+       kwco2d(i,j)   = kwco2 ! m/s (incl. ice fraction!)
```
JorgSchwinger, line 534: The naming is also a bit confusing here, because the piston velocity is not changed by dry/moist assumptions. Since kwco2 is already taken a different name as "piston" or "vp" or so might be an alternative. Sorry quite picky, can go without changing this.

cime_config/buildnml
```diff
@@ -1578,7 +1582,11 @@ cat >>! $RUNDIR/ocn_in$inststr << EOF
 !   EXPOCA         - Ca export production (epcalc100) [mol Ca m-2 s-1]
 !   EXPOSI         - Si export production (epsi100) [mol Si m-2 s-1]
 !   PCO2           - Surface PCO2 (spco2) [uatm]
-!   KWCO2          - kwco2 x solubility
+!   PCO2M          - Surface PCO2 under moist air assumption [uatm]
+!   KWCO2          - kwco2 
+!   KWCO2KHM       - kwco2 x solubility (moist air)
+!   CO2KH          - khd CO2 solubility under dry air assumption
```
JorgSchwinger: Would be nice to stick to the convention used for the other namamelist entries, e.g.

CO2KH     - CO2 solubility under dry air assumption (khd) [mol/kg/atm]

JorgSchwinger (approved):

hamocc/carchm.F90
jmaerz: Ok, I read the `pco2d` formerly as 2D (not dry) as opposed to `pco2` (as local variable) - so extended it to `kwco2d`, since a local variable `kwco2` is already provided ....   - didn't want to rewrite the code.

cime_config/buildnml
jmaerz: I hope the last commit satisfies the conventions. I left the other comment untouched (`kwco2d`) for the mentioned reasons. I hope that's ok.

PR #208, opened 2022-11-25
Bugfix pnetcdf

TomasTorsvik:
Some variables for PnetCDF were not explicitly imported from xc_mod.

PR #209, opened 2022-11-25
Bugfix pnetcdf (#208)

TomasTorsvik:
* Add variables used by PNETCDF to explicit use staements.

* Move implicit none statments

* update explicit use statement for pnetcdf


jmaerz (approved):
Approved.

PR #212, opened 2022-12-02
Fixed dust units and renamed calcium burial to CaCO3 burial

### LINKED ISSUE #211: Units of dust/clay in iHAMOCC output
Hi @JorgSchwinger an @TomasTorsvik , to my understanding, the output units related to clay/dust (`ifdust`/`ssster`) should be kg  (or g, depending on the factor involved) instead of mol (so e.g. kg/m3, ... instead of mol/m3). Any calculation done for dust and clay material seems to be/are done in units of kg/m3 (see e.g. iron-input `mo_apply_fedep`, where dust input is applied or `mo_sedmnt`, where the clay factor `clafa` is calculated). If I am correct, shall we put this into the NorESM2.0.6 version?

jmaerz:
In addition to the clay units, I renamed `Burial calcium` to `Burial CaCO3` which is more appropriate. If need, I can revert this, though.  I also checked the units change against some former output for sediment clay. Closes #211 while the merge with the milestone-branch is still needed afterward.


TomasTorsvik (approved):
Thanks @jmaerz , looks good to me!

JorgSchwinger (approved):
Looks good, thank you for fixing this

PR #215, opened 2022-12-05
Merge from master

TomasTorsvik:
Testing to merge from `master` using rebase on top of `feature-hamocc_beyond_CMIP6`.


hamocc/powach.F90
```diff
@@ -374,7 +374,8 @@ subroutine powach(kpie,kpje,kpke,kbnd,prho,omask,psao,lspin)
            if(powtra(i,j,k,ipowaox) < 3.e-6 .and. powtra(i,j,k,ipowno3) < 3.e-6) then
               posol = denit * sedlay(i,j,k,issso12)         ! remineralization of poc
               umfa = porsol(i,j,k) / porwat(i,j,k)
-              sulf(i,k) = posol*umfa      !this has P units: kmol P/m3 of pore water
```
jmaerz: Hi @TomasTorsvik , the above line
'''
sulf(i,k)=posol*umfa
'''
is the correct line, while the lines 377/378 were deleted. This needs to be reverted. I checked your `merge_from_master` against my own local merge attempt, where I resolved the issues. All other resolved items seem to/should be ok and both our attempts are identical.

jmaerz: Following the suggestion from Jörg, I locally merged `merge_from_master` into the `beyond-CMIP6` branch and compared it with my merge and also with the `master` branch. Results look good to me. So AFTER the the above change, the branch is ready to be merged.

TomasTorsvik: Hi @jmaerz 
Thanks for reviewing. I fixed this, and now powach.F90 is te same as in `feature-hamocc_beyond_CMIP6`.

PR #219, opened 2022-12-12
SI units

milicak:
This is the commit for the SI units into the master.


JorgSchwinger:
Hi all, before merging this into master, wouldn't it be good to make and test the necessary changes for HAMOCC in the feature branch?

matsbn (approved):
I support to merge this now. Default behaviour is identical to current master. I think support for MKS unit system for HAMOCC and the channel setup can be added later and was one of the reasons to support both CGS/MKS unit systems for an interim period. I have several pending code modifications in files this PR modifies, and it would be convenient to introduce them in branches based on updated master.

TomasTorsvik (approved):
I think we can merge this to master now, even though the MKS option does not work with iHAMOCC at the moment. CGS is still the default, so this works as before.

PR #221, opened 2023-01-06
Add option for surface pH output

JorgSchwinger:
Has been tested with and without natDIC option.

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -422,6 +422,7 @@ set SRF_ALKALI    = '4, 2, 2'
 set SRF_SILICA    = '0, 2, 2' 
 set SRF_DIC       = '4, 2, 2' 
 set SRF_PHYTO     = '4, 2, 2' 
+set SRF_PH        = '0, 2, 2' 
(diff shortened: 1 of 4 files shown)


jmaerz (approved):
Looks generally ok to me. The question is generally, if you want to generate the time mean ph, or ph of time mean H+ concentration, though (which can make a -likely subtle- difference). Only minor thing: one could rename 'h+' to 'H+' in the netcdf variable units (as in '-log10(H+)').

TomasTorsvik (approved):
Thanks, @JorgSchwinger , no complaints from my side!

JorgSchwinger:
Thanks, Jöran and Tomas for reviewing this quickly!

@jmaerz I also just got aware that we do not save the average pH, but the log10 of the average H+ concentration. Since this has been done always like this for the 3d-pH field in HAMOCC (at least in Bergen), I implemented it consistent with the 3d fields. I am not sure how large the difference is. If there is a difference, one should have the option for both average H+ and average pH.

I have changed h+ to H+.

PR #222, opened 2023-01-09
Remove unused parameters in wrt* subroutine calls in ncout_hamocc.F90

JorgSchwinger:
Since PNETCDF has been introduced, attributes to netCDF output variables are defined in `hamoccvardef` and not longer during writing the fields to file in the wrt* subroutines defined in mo_bgcmean.F90. I have now removed these no longer used arguments, which avoids confusion and improves readability of `ncout_hamocc` quite a bit.

### DIFF
--- a/hamocc/mo_bgcmean.F90
+++ b/hamocc/mo_bgcmean.F90
@@ -55,7 +55,7 @@
 !**********************************************************************
       use mod_xc,         only: ii,jj,kk,idm,jdm,kdm,nbdy,ifp,isp,ilp,mnproc,ip
       use mod_dia,        only: ddm,depthslev,depthslev_bnds,nstepinday,pbath
-      use mod_nctools,    only:ncpack,nccomp,nccopa,ncwrtr
+      use mod_nctools,    only: ncpack,nccomp,nccopa,ncwrtr
       use netcdf,         only: nf90_fill_double
       use mo_param1_bgc,  only: ks
       use mo_control_bgc, only: get_bgc_namelist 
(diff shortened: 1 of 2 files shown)


jmaerz (approved):
Dear Jörg, looks good to me - I haven't checked each individual factor, though. So I approve. The only thing that is getting a bit puzzling to me is the further detachment of variables and units (as in: the factors set in wrtX is even less clear now... - my first reaction was wishing for a unit comment :-) ). When looking into your code changes, I started to ponder, whether it would be possible to write a wrapper function that combines i) declaration of netcdf variable ii) closing of `iogrp`, iii) the writing of the actual variable to netcdf and iv) re-opening of the `iopgr`. Maybe we can discuss this.

JorgSchwinger:
Hi Jöran, everything is still there - in the `hamoccvardef` subroutine, so you can still find the variables together with the long names and units. This PR removes only duplicate (and unused) stuff.

One could think of combining `ncout_hamocc` and `hamoccvardef` into `mo_bgcmean` (or creating a new module `mo_ncout_hamocc`)

jmaerz:
Hi Jörg, sure, I know. From my side just go ahead as approved. My point is that one sets a factor somewhere and defines the nc variables units somewhere else (while indeed that was the case before, as I now understand better after your changes - I also got confused about the doubling of the information that we provided before - so in that way, it's clearer now). Thanks for the work!

JorgSchwinger:
Mhm yes, I see the point with the units and the conversion factor. I agree that this not nice, although I am not sure if there is a way around it (I think with PNETCDF everything has to be defined before any processor starts writing into the file, while the conversion factor can only be applied at the time of writing).

I'll see if Tomas has something and go ahead with merging this.

TomasTorsvik (approved):
Hi @JorgSchwinger , this loks fine to me. I don't have anything to add beyond what was already commented.

PR #224, opened 2023-01-23
Add linear ramping-up scenario for ocean alkalinisation

TimotheeBrgs:
[mo_read_oafx.F90](https://github.com/NorESMhub/BLOM/blob/master/hamocc/mo_read_oafx.F90) was previously including 2 ocean alkalinisation scenarios with constant alkalinity addition over time. This commit adds another scenario with linear ramping up of alkalinity addition from 0 Pmol alk/yr in year  `ramp_start` to `addalk_ramp` Pmol alkalinity/yr in year `ramp_end`. Then the alkalinity addition is kept constant after `ramp_end`.


hamocc/mo_read_oafx.F90
```diff
@@ -77,13 +82,19 @@ module mo_read_oafx
...
   !
   real, parameter :: addalk_0p14   = 0.14  ! Pmol alkalinity/yr added in the
   real, parameter :: addalk_0p56   = 0.56  ! 'const_0p14' and 'const_0p56' 
                                            ! scenarios
   real, parameter :: cdrmip_latmax =  70.0 ! Min and max latitude where
   real, parameter :: cdrmip_latmin = -60.0 ! alkalinity is added according
                                            ! to the CDRMIP protocol
+  real, parameter :: addalk_ramp   = 0.135 ! Max Pmol alkalinity/yr added
+  integer, parameter :: ramp_start = 2025  ! in 2035 in the 'ramp' scenario,
+  integer, parameter :: ramp_end   = 2035  ! starting at 0 Pmol/yr in 2025.
```
TomasTorsvik: It is a bit arbitrary to have the ramp-up start and end years as hard-coded parameters. Would it be better to read this from a namelist file? On the other hand, the min/max latitude limits are equally arbitrary ...
@JorgSchwinger , @jmaerz , any thoughts about this?

TomasTorsvik (approved):
The code contribution looks fine to me, so I will approve it. The only question I have is whether or not the parameters can be read in a more sensible way.

TomasTorsvik:
@TimotheeBrgs 
Hi, I made branch in my personal fork based on this PR with a suggestion for how to read the parameters from a namelist file
https://github.com/TomasTorsvik/BLOM-TTfork/tree/OA_rampscen_read
Would this soltion work for you?

Here is the commit diff:
https://github.com/TomasTorsvik/BLOM-TTfork/commit/69f30fef51c9f43145a77b8d4c95c0b3d0a87999

jmaerz:
Hi @TimotheeBrgs , I agree with @TomasTorsvik that it would be good to discontinue hard-coding individual cases and make the structure more flexible to user demands (we touched upon it in #187). @TomasTorsvik , thanks for the alternative solution! One could even further simplify the routine `get_oafx` and only write it in the form of the ramping (the moment that `ramp_start` == `ramp_end`, the whole procedure reduces to the formerly implemented case, I believe) - but maybe this reduces a bit the readability (so more meant as an option). To keep the information on the other two use cases of alkalinization, it would be good to automatically adjust the `ADDALK_CONST` parameter in the `buildnml` according to the scenario - so everything could be handled via the nml. Further, even though Fortran should do it internally, an explicit casting from `int` to `float` would be nice for the `oafx` calculation (or at least `365.` instead of `365`).

TomasTorsvik:
> Here is the commit diff: [TomasTorsvik@69f30fe](https://github.com/TomasTorsvik/BLOM-TTfork/commit/69f30fef51c9f43145a77b8d4c95c0b3d0a87999)

Forgot to re-label the namelist entry ...
https://github.com/TomasTorsvik/BLOM-TTfork/commit/9725deb3533fc9a0358ea1e5f46c5df763c63ae5

hamocc/mo_read_oafx.F90
```diff
@@ -268,10 +286,28 @@ subroutine get_oafx(kpie,kpje,kplyear,kplmon,omask,oafx)
...
+    if(kplyear.lt.ramp_start ) then
+      oafx(:,:) = 0.0
+    elseif(kplyear.ge.ramp_end ) then
+      oafx(:,:) = oalkflx(:,:)
+    else
+      current_day = (kplyear-ramp_start)*365.+nday_of_year
+      oafx(:,:) = oalkflx(:,:) * current_day / ((ramp_end-ramp_start)*365.)
+    endif
+
+    if(mnproc.eq.138 ) then
```
jmaerz: Is there any reason to write out at 138 and not at 1? I would suggest to stay with 1, so `mnproc.eq.1`

hamocc/mo_read_oafx.F90
```diff
@@ -66,24 +69,32 @@ module mo_read_oafx
...
-  real, parameter :: addalk_0p14   = 0.14  ! Pmol alkalinity/yr added in the
-  real, parameter :: addalk_0p56   = 0.56  ! 'const_0p14' and 'const_0p56' 
-                                           ! scenarios
-  real, parameter :: cdrmip_latmax =  70.0 ! Min and max latitude where
-  real, parameter :: cdrmip_latmin = -60.0 ! alkalinity is added according
-                                           ! to the CDRMIP protocol
+  real, protected :: addalk    = 0.56  ! Pmol alkalinity/yr added in the
+                                           ! scenarios. Read from namelist file
+                                           ! to overwrite default value.
+  real, protected :: cdrmip_latmax =  70.0 ! Min and max latitude where
```
TomasTorsvik:
```suggestion
  real, protected :: addalk        = 0.56  ! Pmol alkalinity/yr added in the
                                           ! scenarios. Read from namelist file
                                           ! to overwrite default value.
  real, protected :: cdrmip_latmax =  70.0 ! Min and max latitude where
```

TomasTorsvik: Adjust code allignment

cime_config/buildnml
```diff
@@ -237,6 +237,11 @@ endif
 set DO_OALK = .false.
 set OALKSCEN = "''"
 set OALKFILE = "''"
+set ADDALK    = 0.135
+set CDRMIP_LATMAX = 70.0
+set CDRMIP_LATMIN = -60.0
+set RAMP_START    = 2025
+set RAMP_END      = 2035
```
TomasTorsvik:
```suggestion
set BGCOAFX_DO_OALK = .false.
set BGCOAFX_OALKSCEN = "''"
set BGCOAFX_OALKFILE = "''"
set BGCOAFX_ADDALK    = 0.135
set BGCOAFX_CDRMIP_LATMAX = 70.0
set BGCOAFX_CDRMIP_LATMIN = -60.0
set BGCOAFX_RAMP_START    = 2025
set BGCOAFX_RAMP_END      = 2035
```

TimotheeBrgs:
Hi all, thank you for reviewing this, the suggestions and the proposed commit.

From @TomasTorsvik's commit, I simplified the code by merging the 2 constant scenarios into a single `const` scenario, knowing that the flux can now be defined externally. I use now `365.` instead of `365` to explicitly cast from `int` to `float` as suggested. I suggest keeping the distinction between `const` and `ramp` scenarios for the readibility. All ocean alkalinization parameters can now be defined in the namelist user_nl_blom. I rearranged the buidnml in order to get all oafx-related parameters in the BGCOAFX section.

There is a mention of `addalk_const` as a comment on line 97 of @TomasTorsvik's [example](https://github.com/TomasTorsvik/BLOM-TTfork/blob/69f30fef51c9f43145a77b8d4c95c0b3d0a87999/hamocc/mo_read_oafx.F90) , this variable does not exist and might have created confusion with the `ADDALK_TOT` of the original version of the code. I renamed `ADDALK_TOT` by just `ADDALK` to improve readability.

hamocc/mo_read_oafx.F90
TimotheeBrgs: This should not be here, sorry. It was for debugging purpose. I was using proc 138 because proc 1 does not include ocean grid cells so I was not able to verify that the code was working. I will remove it.

cime_config/buildnml
TomasTorsvik: We haven't done it before, but I wonder if it would make sense to introduce a naming convention for these variables on the form: `NAMELIST`_`VARIABLE`
to make it clear where the variable belongs. Not sure if this is the best way, but I think it can be useful.

cime_config/buildnml
```diff
@@ -1495,6 +1497,28 @@ cat >>! $RUNDIR/ocn_in$inststr << EOF
...
+! RAMP_END      : End year for 'ramp' scenario
+&BGCOAFX
+  DO_OALK       = $DO_OALK
+  OALKSCEN      = $OALKSCEN
+  OALKFILE      = $OALKFILE
+  ADDALK        = $ADDALK
+  CDRMIP_LATMAX = $CDRMIP_LATMAX
+  CDRMIP_LATMIN = $CDRMIP_LATMIN
+  RAMP_START    = $RAMP_START
+  RAMP_END      = $RAMP_END
```
TomasTorsvik:
```suggestion
  DO_OALK       = $BGCOAFX_DO_OALK
  OALKSCEN      = $BGCOAFX_OALKSCEN
  OALKFILE      = $BGCOAFX_OALKFILE
  ADDALK        = $BGCOAFX_ADDALK
  CDRMIP_LATMAX = $BGCOAFX_CDRMIP_LATMAX
  CDRMIP_LATMIN = $BGCOAFX_CDRMIP_LATMIN
  RAMP_START    = $BGCOAFX_RAMP_START
  RAMP_END      = $BGCOAFX_RAMP_END
```

TomasTorsvik: See previous comment.

hamocc/mo_read_oafx.F90
```diff
@@ -268,10 +286,28 @@ subroutine get_oafx(kpie,kpje,kplyear,kplmon,omask,oafx)
...
+  ! Scenario of ramping-up fluxes
+  !--------------------------------
+  elseif(trim(oalkscen)=='ramp' ) then
+
+    if(kplyear.lt.ramp_start ) then
+      oafx(:,:) = 0.0
+    elseif(kplyear.ge.ramp_end ) then
+      oafx(:,:) = oalkflx(:,:)
+    else
+      current_day = (kplyear-ramp_start)*365.+nday_of_year
```
jmaerz: Fortran compilers will probably handle this, but first you declare `current_day` as `int` and in l. 303 you write now 365. (real number) instead of 365 (integer).
So, I guess, I would either declare it as real above or remove the `.`, while casting all the stuff to float in the next line (where you can the use `365.`).

TomasTorsvik (approved):
I suggest some minor modifications. Accept if you agree, it should be included in the PR automatically.

hamocc/mo_read_oafx.F90
TimotheeBrgs: Hi @jmaerz, good point thanks. It will be clearer despite having the compiler handling it.
if I leave `current_day` declared as `int`, just to keep it intuitive, are you fine with the below?
```
      current_day = (kplyear-ramp_start)*365+nday_of_year
      oafx(:,:) = oalkflx(:,:) * current_day / ((ramp_end-ramp_start)*365.)
```
I hesitate using the Fortran intrinsic procedure `FLOAT(foo)`.

jmaerz: I am fine with this (casting will be done internally). Out of curiosity: any reason not to use `float()`?

TimotheeBrgs: No specific reason. That was my doubt. If FLOAT is used, I presume that there is no need to use a `.` to 365 knowing that it will be explicitely casted.

Let's fully agree :) Are you fine with this one?

`oafx(:,:) = oalkflx(:,:) * FLOAT( current_day / ((ramp_end-ramp_start)*365) )`

jmaerz: Erh, now I am getting uncertain how Fortran will/would handle this - first making an integer division and then turning the truncated integer value into a float? Just stick with  
```oafx(:,:) = oalkflx(:,:) * current_day / ((ramp_end-ramp_start)*365.)```
which should conform to Fortran standards. My former intention was to apply float to the integer values individually. But just keep it simple.

jmaerz (approved):
Just go ahead from side. Approved.

PR #225, opened 2023-01-25
Import get_bgc_namelist only in subroutine where it is needed.

TomasTorsvik:
I think it makes sense to move the module import statement to the subroutine where `get_bgc_namelist` is used. Not sure why it was placed in the main body of `mo_bgcmean` in the first place.

### DIFF
--- a/hamocc/mo_bgcmean.F90
+++ b/hamocc/mo_bgcmean.F90
@@ -58,7 +58,6 @@
       use mod_nctools,    only: ncpack,nccomp,nccopa,ncwrtr
       use netcdf,         only: nf90_fill_double
       use mo_param1_bgc,  only: ks
-      use mo_control_bgc, only: get_bgc_namelist 
 
       IMPLICIT NONE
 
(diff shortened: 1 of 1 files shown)


jmaerz (approved):
Hi @TomasTorsvik , from my side, it's fine. Admittedly, I would be tempted to disregard this fine-grained import in ncout_hamocc.F90, once we make it to a module, so that the import is getting less lengthy and cumbersome. But this is another discussion. So go ahead. Approved.

jmaerz:
Maybe one question, though. Any idea, why the testing is with this change now failing with a time-out (and not earlier, asking wrt the intel compiler)?

TomasTorsvik:
> Maybe one question, though. Any idea, why the testing is with this change now failing with a time-out (and not earlier, asking wrt the intel compiler)?

Don't know why this happens. The macos check has been failing for a while, but the others usually work. The timeout (at 900s) happens when running the fuk95 test case, and the other Intel check runs this case in 366s. I re-run the check to see if it was just a one-off event.

jmaerz:
I did already the same ;-) - and it also failed.

JorgSchwinger (approved):
Looks fine to me.

jmaerz:
Hah, seems like the test-case now went through and the time might be just on the edge of the set time limit for the test run.

PR #229, opened 2023-01-26
fix missing '

jmaerz:
@TomasTorsvik and the same for the `beyond_CMIP6 branch`


JorgSchwinger:
Thanks for fixing this, and sorry for the trouble!

TomasTorsvik (approved):
Thanks for fixing!

PR #230, opened 2023-01-27
Merge master into feature-hamocc_beyond-CMIP6 branch

jmaerz:
@TomasTorsvik & @JorgSchwinger: I am trying to merge the recent `master` changes, incl. SI units on the physics side into the `feature-hamocc_beyond-CMIP6` branch. It is resulting in merging conflicts and I am first trying to resolve them online. If that doesn't work, I'll move to a local approach to resolve this. The latter might include merging the two branches locally and trying to push directly to NorESMhub (don't know, if that's possible for me, though) in order to 'synchronize' the master and the beyond-CMIP6 branch properly again - this isn't the case at the moment due to formerly resolved merge conflicts via a temporary merging branch. Direct merging thus eases i) future merging and ii) merging the `beyond-CMIP6` branch back to master (which we should discuss/think about).


jmaerz:
Closed because of dangerous behavior of github - for details see closed pull request #231 (basically, after resolving the conflicts, changes would be written to master - and not, as expected to the feature branch in this case).

PR #233, opened 2023-01-30
SIunits4hamocc

### LINKED ISSUE #223: iHAMOCC: convert to SI units (CGS -> MKS)
BLOM is introducing SI units (#219). We need to make adjustments in iHAMOCC accordingly.

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , I drafted a first change wrt new SI units introduced lately in BLOM in response to #223 . I will have another look into it, but thus far, it seems as if the changes can be handled rather simple in `mo_intfcblom.F90`. The first commit with message `TESTING ONLY: ...` in this branch hasn't changed anything, but allows to write stuff out in a 1D water column, how it will/would be changed (was done for testing purposes). I corrected the units for `prho`, since it seems as if it was always calculated as `g/cm3` and not `kg/m3` (and also seem to be used as such in the sediment routines - at least it was labeled there differently from `hamocc4bgm`).   

I also would suggest that on the physics side, some comments on state variables should be updated (e.g. for units of `dp` in `mod_state.F90`)


jmaerz:
Dear @matsbn , since we discussed the SI units for iHAMOCC yesterday, I would like to bring this drafted pull request to your attention, which you might want to follow or comment on.

JorgSchwinger (commented):
To me this looks good. Should be tested by running a few hundred years (NOINYOC tn21 compset would be sufficient) to see if we have overlooked something.

TomasTorsvik:
> To me this looks good. Should be tested by running a few hundred years (NOINYOC tn21 compset would be sufficient) to see if we have overlooked something.

I will do some test runs with NOINYOC-tn21 based on this PR.

TomasTorsvik:
I have done some NOINYOC-tn21 runs over the weekend:
A: 100 year noresm2.0.6 (reference run)
B: 10 year SIunits4hamocc - CGS
C: 100 year SIunits4hamocc - MKS

A and B were bit-identical afte 10 year tun, so I didn't continue the B run. A and C were different, and I have used the standard diagnostics toolbox to compare the output after 100y.

BLOM_DIAG : http://ns2980k.web.sigma2.no/tomast/diagnostics/NOINYOC_T62_tn21_noresm2.0.6_10y_20230210T1812/BLOM_DIAG/yrs1to100-NOINYOC_T62_tn21_noresm-SIunits4hamocc-MKS_10y_20230210T1826/

HAMOCC_DIAG : http://ns2980k.web.sigma2.no/tomast/diagnostics/NOINYOC_T62_tn21_noresm2.0.6_10y_20230210T1812/HAMOCC_DIAG/yrs1to100-NOINYOC_T62_tn21_noresm-SIunits4hamocc-MKS_10y_20230210T1826/

From the physics we see that the differences in the output fields are fairly small, but the evolution of the model run differs a bit, e.g. time series for maximum AMOC. Hence, it is not unexpected that we see some discrepancies also in BGC time series for global fluxes and averages, in particular for surface DIC and alkalinity. DMS flux also shows some notable differences.

None of the differences are orders of magnitude in range, so there is nothing obviously wrong with the cgs2mks conversion. However, the differences are not negligible, so I'm not entirely sure everything is in order.

matsbn:
Some comments/suggestions:

In line 324 of mo_intfcblom.F90 I would consider to change

        else if(dp(i,j,kn) <  1.0e-3*M_mks2cgs/L_mks2cgs) then

to

        else if(dp(i,j,kn) <  1.0e-3*P_mks2cgs) then

This should end up the same, but P_mks2cgs has correct unit, while M_mks2cgs/L_mks2cgs does not.

How badly will iHAMOCC fail if "bgc_rho" is kept at the selected MKS or CGS unit, and not being forced to be CGS? Within mo_intfcblom.F90 the application of rho0 for rescaling seems to cancel, and this also seems to partly hold for other applications of "bgc_rho", but I only looked a few places. For example in beleg_vars.F90 the ocean tracer is divided by rho so that should be fine with consistent unit of rho. Then in the same routine there is an expression like "ocetra(i,j,k,ibromo)= 1.e-14/prho(i,j,k)" where again a dimensional constant (1.e-14) will cause problems. There could of course be many of those (we certainly had many in BLOM), but it seems somewhat inconsistent to enforce CGS unit of density if MKS is used.

jmaerz:
Dear @matsbn , first of all: thanks for looking into it and for making the suggestions for change. I'll take a look and change it.
For the second part of your comment: I am not entirely sure, if I understand it right. Thus far, the current status is to use `rho` in iHAMOCC always in CGS units (as in g/cm3) due to historical reasons and convert `rho` from BLOM MKS to iHAMOCC CGS via `mo_intfcblom.F90`. Inside iHAMOCC, all/most tracers are then handled in terms of kmol/m3 (and mixed units for used physical variables, which is admittedly not nice).

JorgSchwinger:
A difference of almost 0.5 PgCy-1 for PP is more than what I would expect. AMOC and drake passage transport are not systematically different, so I'm not so convinced that this can be explained by differences in uppwelling etc.

I have not time right now to look closer into this, but my conclusion from this test run is that we should check this further.

TomasTorsvik:
@jmaerz , @JorgSchwinger 

image

It seems there may be something fishy already at the startup, as there is a difference of 0.2 PgCy-1 for PP at year 1.

jmaerz:
Many thanks! - that hopefully helps a lot to further trace it down.

JorgSchwinger:
Yes, thank you! 

Looking at the corresponding plot for nutrients (which are the tracers that are initialized) everything seems ok (no visible difference). Looks more like something in the forcing (light?) is off.

jmaerz:
Maybe just as a side note - I am not sure, in how far this is relevant, but the #245 is still a persisting issue (and was a bit holding me back from following up on the SIunits-issue) - I don't know, why the calculations for `dens` and `thkns` (and `dpth`) go wrong in `mod_inicon.F` when writing it out for the `init.profile`.

jmaerz:
I was setting up two runs - one `mks` and one `cgs` with the `SIunits4hamocc` branch (NOINYOC - T62_tn21) - just for a month and was running in `debug` mode - so writing out information on inventories after individual iHAMOCC processes. Right after startup, the inventories differ (left cgs, right mks):
cgs-left-verus-mks-right-at-startup

After one day of run time (daily mean otput), difference in phosphate concentrations (and also other tracers) show a strange pattern... (**Note the different scales - MKS minus CGS**)
**At surface:**
mks-minus-cgs-after-1d

**At 100m:**
mks-minus-cgs-after-1d-100m

.... and after one month (still daily mean output - also note again the different scale as compared to the first day):
**At surface:**
mks-minus-cgs-after-1month

**At 100m:**
mks-minus-cgs-after-1month-100m

**At surface - day 1 - percent difference:**
percent-difference-1d

On the physics side, there are some similar surface pattern - but not as rigid as in iHAMOCC - and also increasing differences in time - locally pretty high percentage difference after one month particularly in frontal areas for sst. 

**Percentage difference first day daily averaging:** 
percent-difference-1d-physics-sst

percent-difference-1d-physics-sss

JorgSchwinger:
Hi Jöran, thanks for doing these runs! Maybe this is something with the freshwater fluxes? It looks like around large rivers there is a larger error. Also, in the Southern Ocean, there is some freshwater input due to the melting of the initial ice cover. Possibly the virtual fluxes for the tracer fields have not been correctly updated? I can't remember where to look for it, but I can find it tomorrow when I'm back in office.

jmaerz:
Hi Jörg, for today, two observations wrt calculation of virtual fluxes: the calculations differ in `./ben02/thermf_ben02.F90` (`trflxc=0`, l. 699) versus `./cesm/thermf_cesm.F` (`trflxc=(-tottrsf)/area`, l. 417) and I am a bit uncertain how to interpret the multiplication with `L_mks2cgs` for the `trflx` calculation in these routines (few lines below). But both observations are for now without deeper understanding. The calculated `trflx` is then applied also in `./phy/mod_mxlayr.F` (just for present reference).

jmaerz:
In addition to the physics plots above after 1d of runtime (**1d average, difference in percent**), here the same for end of the first month (also 1d average):
percent-difference-1month-physics-sst
percent-difference-1month-physics-sss
The biggest differences are in the most dynamic regions, where I also would expect to see the biggest effects of round-off errors. @matsbn , I am not entirely sure, how much of a difference one could/should expect from round-off errors due to cgs->mks change.

jmaerz:
@JorgSchwinger , I can merge the latest `master` into this branch so that you can use it for your test?

jmaerz:
Hi @JorgSchwinger , I just updated this branch with the most recent `master` so that you make easily make use of it.

jmaerz:
Technically, we can merge this branch into master and afterward, you can put in your fixes on top @JorgSchwinger , but up to you.

TomasTorsvik:
I support pushing this to `master`, and go on with further fixes from there.

JorgSchwinger:
I think it's fine merging this into master. 

I will probably (hopefully) come up with a couple of fixes afterwards. But running the model in default cgs configuration isn't affected, so this could be merged.

JorgSchwinger (approved):

TomasTorsvik (approved):

PR #234, opened 2023-01-31
Quick fix on buildnml after incomplete shell variable renaming from PR 224

TimotheeBrgs:
A proposed commit has been suggested in [PR 224](https://github.com/NorESMhub/BLOM/pull/224) to include a new naming convention for shell variables in buildnml. I had approved the commit but there was 2 missing renaming that I missed. The code compiles correctly with this PR.


TomasTorsvik (approved):
Thanks, Timothee!
Since this is a straight forward bug fix, we can do without a lengthy review :)

PR #235, opened 2023-02-08
Hybrid coordinate enhancements

matsbn:
Hybrid coordinate enhancements that includes: Separate non-local terms for vertical distribution of surface temperature restoring flux, surface salinity restoring flux and brine flux; Gradual change of layer thickness in a transition zone separating interfaces with constant pressure and constant potential density, respectively; Submesoscale eddy-induced transport parameterization for hybrid vertical coordinate.

CI issues on macOS have been fixed.


cime_config/buildnml
```diff
@@ -675,9 +700,11 @@ else if ($OCN_GRID == tnx1v1 || $OCN_GRID == tnx1v3 || $OCN_GRID == tnx1v4) then
     set BACLIN = 3600.
     set BATROP = 60.
     set CWBDTS = .75e-4
-    set CE     = .5
     set NIWGF  = .4
     set SMTFRC = .false.
+    if ($BLOM_VCOORD == isopyc_bulkml) then
+      set CE     = .5
```
JorgSchwinger, line 706: Does this mean that CE is undefined if $BLOM_VCOORD /= isopyc_bulkml? Is that ok?

JorgSchwinger (approved):
Looks good to me, although I can't review any details here.

TomasTorsvik (approved):
I don't see any red flags, so it looks fine to me as well.

cime_config/buildnml
matsbn: CE is first set to a common literature value of 0.06 (Fox-Kemper et al., 2008), then adjusted (quite strongly) if $BLOM_VCOORD == isopyc_bulkml. With hybrid coordinate we found reasonable submesoscale eddy-induced overturning strength with the standard value.

milicak (approved):

PR #239, opened 2023-02-14
Hamocc fix restart file units

### LINKED ISSUE #238: Units in iHAMOCC restart file
@JorgSchwinger and @TomasTorsvik: looking into `aufw_bgc.F90`, I think the units for `ssster` and `bur_clay` should be `kg/m**3` and `kg/m**2`, respectively (both similar to the unit fixes introduced in #213 - I missed the restart file writing in #213). Additionally, `satoxy` should be   `Saturated oxygen` in `mol O2 kg` (currently handled as `xxx..`). I am not sure, if it would break any run cases (and automatic reading of restart files) though.

jmaerz:
Hi @TomasTorsvik  and @JorgSchwinger , I prepared a fix for the units in the restart file as suggested in #238. I noticed that the particulate materials were all in `X/m2`, while I believe it should read `X/m3` and changed this as well. Of course, for the burial layer, it should stay with `X/m2`. Closes #238

### DIFF
--- a/hamocc/aufw_bgc.F90
+++ b/hamocc/aufw_bgc.F90
@@ -730,7 +730,7 @@
      &    rmissing,92,io_stdo_bgc)
 
       CALL NETCDF_DEF_VARDB(ncid,8,'bur_clay',3,ncdimst,ncvarid,        &
-     &    9,'kg/m**2',20,'Burial layer of clay',                      &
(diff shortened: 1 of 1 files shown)


JorgSchwinger (approved):
Looks good, thanks for fixing this!

TomasTorsvik (approved):
Great!

PR #242, opened 2023-02-15
fix string count number

jmaerz:
It seems as I have missed to also adjust the string counter when changing the units in the restart file writing, which currently affects the restart file writing - I am about to test it on betzy now.


jmaerz:
Tested on betzy - restart file successfully written - merged.

PR #243, opened 2023-02-17
Removed OMP in pressure calculation to avoid race condition

jmaerz:
Hi @JorgSchwinger, Tomas and I, we yesterday identified a potential race condition that (among others) causing issues when running BLOM/iHAMOCC in single column mode. Further places in iHAMOCC seem to be in `ocprod` (settling routine - potential race condition due to `k` versus `kdonor`?) and  in `carchm` (to be checked). In this PR, I only addressed the pressure calculation. Currently, single column mode with `master` is only advisable with flag `-D openmp=disabled` to get reproducible results.

### DIFF
--- a/hamocc/mo_intfcblom.F90
+++ b/hamocc/mo_intfcblom.F90
@@ -269,8 +269,6 @@ subroutine blom2hamocc(m,n,mm,nn)
 
 ! --- calculate pressure at interfaces (necesarry since p has
 ! --- not been calculated at restart)
-
-!$OMP PARALLEL DO PRIVATE(k,kn,l,i)
   do k=1,kk
      kn=k+nn
      do j=1,jj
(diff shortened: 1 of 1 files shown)


TomasTorsvik:
I suspect the single column case is more sensitive to these OpenMP thread race issues, since all the threads will try to access the same section of memory at the same time, which is less likely to happen for larger cases. However, the test shows that there is potential for random errors, so we should definitely try to eliminate them.

TomasTorsvik (approved):

JorgSchwinger:
What is a race condition? Anyway I'll approve this, if it helps to get the single column to run.

JorgSchwinger (approved):

jmaerz:
@JorgSchwinger with my own limited understanding: depending on the time the different OpenMP threads finish, the results vary (due to the interdependence between the `k` and `k+1`) - here it means that a split in the `k` loop results in `rp(i,j,k+1)=dp(i,jkn)` since `rp(i,j,k)` can still be 0.  Tomas figured out that on Betzy, someone was setting the number of OMP threads to 1 which prevents from such race conditions (and results should then be identical). I tapped into this issue since I used the single_column setup with OMP.

PR #244, opened 2023-02-17
Fix OPenMP issues in carchm

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , in connection to PR #243, I checked the OMP private statement in `carchm`. Indeed a number of local real values were not included (I am aware that loop indices are private by default, but it doesn't harm to be even more explicit) - probably most importantly also for the Si units: `rrho`. The fix is only related to `carchm`. In combination with the fix in #243, variations in DIC inventories during the first time steps disappear. I hope, no further local real value escaped my investigation. Upon request, I sort in the values in the OMP statement (for now I thought it is easier to see the changes). Issues with the sinking in OCPROD still remain (I guess, the easiest would be to un-OMP the settling to avoid race conditions for `k` versus `kdonor`).

### DIFF
--- a/hamocc/carchm.F90
+++ b/hamocc/carchm.F90
@@ -206,7 +206,7 @@
 !$OMP  ,niflux,n2oflux,dmsflux,omega,supsat,undsa,dissol              &
 #ifdef CFC
 !$OMP  ,sch_11,sch_12,sch_sf,kw_11,kw_12,kw_sf,a_11,a_12,a_sf,flx11   &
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):
Thanks @jmaerz !
it seems that some of these variables were missing in the `private` statement. 
I believe the convention is that the index of the first DO statement inside a OMP PARALLEL DO is always private by defaault, but I can't find the reference now. Still, I don't think there is any harm in explicitly declaring them as private.

JorgSchwinger (approved):

PR #248, opened 2023-03-29
Update config_pes for OMIP-2 simulations on Betzy

TomasTorsvik:
I wanted to run OMIP-2 simulations on Betzy, and found that I could not find a suitable `config_pes` definition for NOIIAJRA for runnig on Betzy. I have made a new definition for running on 512 cores (4 nodes), but I'm not sure it's an optimal setting. Should it be defined for "betzy" only, or for some particular compset (e.g. "DATM") only?

### DIFF
--- a/cime_config/config_pes.xml
+++ b/cime_config/config_pes.xml
@@ -187,6 +187,46 @@
       </pes>
     </mach>
   </grid>
+
+  <!-- 512 pes is minimum setting for normal job size on betzy !-->
+  <grid name="a%TL319.+oi%tnx1v4">
+    <mach name="betzy">
+      <pes pesize="L" compset="_DATM.*_BLOM">
+	      <comment>Large pe-layout with 512 pes in total</comment>
+	      <ntasks>
+	        <ntasks_atm>1</ntasks_atm>
+	        <ntasks_rof>1</ntasks_rof>
+	        <ntasks_ice>154</ntasks_ice>
+	        <ntasks_ocn>354</ntasks_ocn>
+	        <ntasks_cpl>158</ntasks_cpl>
+	        <ntasks_lnd>1</ntasks_lnd>
+	        <ntasks_glc>1</ntasks_glc>
+	        <ntasks_wav>1</ntasks_wav>
+	      </ntasks>
+	      <nthrds>
+	        <nthrds_atm>1</nthrds_atm>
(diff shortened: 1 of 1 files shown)


cime_config/config_pes.xml
```diff
@@ -602,4 +602,42 @@
     </mach>
   </grid>
 
+  <grid name="a%TL319.+oi%tnx1v4">
```
JorgSchwinger: The configuration above (line 153) seems to be fine for fram, whereas on betzy your configuration has less idle processors. But I would suggest to move this configuration up, below the one in line 153. Then it might make sense to specify mach="betzy" here, and maybe compset="_DATM.*_BLOM%ECO" to avoid that this configuration is picked unintendely instead of the one in line 415

TomasTorsvik: Thanks for input, I will set mach="betzy" and compset="_DATM.*_BLOM"
I would like to keep the option to run with and without iHAMOCC. Once in a while it is interesting to check how much extra resources are needed for running the model with BGC.

matsbn:
Sorry for not responding sooner to this PR. Actually, it seems that hardly any processor configuration for TL319+tnx1v4 is utilizing betzy nodes well (e.g for compset _DATM.*_BLOM%ECO, more than 70% of the 5th node is not used), so some overhaul of this probably needed. Specific comments to this PR:
- CICE can be quite performance sensitive to processor count. Did you test the speed of the particular choice of CICE processors?
- CICE does not generally reproduce results with different processor counts. It could be beneficial to get the same ice/ocn results with or without HAMOCC enabled, but then number of CICE processors must be the same.

@monsieuralok has quite some experience in benchmarking this, so maybe he can provide some recommendations.

monsieuralok:
@TomasTorsvik You can merge this pull request as it is for tnx1v4 it would not impact much with CICE only check in CASE folder in env_build.xml that <entry id="CICE_DECOMPTYPE" value="sectrobin"> it exist. Usually, that is the fastest decomposition type especially for tripolar grid.

TomasTorsvik:
Thanks for the feedback!
@monsieuralok I'm not sure I understand what I should look for in `env_build.xml`.  There is a reference to the case build directory, but that is set to `$EXEROOT` by default. Do you mean I should check if the `env_build.xml` file exist in the CASE folder?

PR #251, opened 2023-04-28
fixed pnetcdf usage in CIME

mvertens:
The -DPNETCDF flag was not getting set as part of the build using CIME.
This PR fixes that problem.

### DIFF
--- a/cime_config/buildcpp
+++ b/cime_config/buildcpp
@@ -89,10 +89,14 @@ def buildcpp(case):
     hamocc_ciso = case.get_value("HAMOCC_CISO")
     hamocc_vsls = case.get_value("HAMOCC_VSLS")
     blom_unit = case.get_value("BLOM_UNIT")
+    pio_typename = case.get_value("PIO_TYPENAME", subgroup="OCN")
 
     expect(blom_vcoord != "cntiso_hybrid" or not turbclo, "BLOM_VCOORD == {} and BLOM_TURBULENT_CLOSURE == {} is not a valid combination".format(blom_vcoord, turbclo))
 
     blom_cppdefs = ""
+
+    if pio_typename == "pnetcdf":
+        blom_cppdefs = blom_cppdefs + " -DPNETCDF"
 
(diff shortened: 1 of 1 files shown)


mvertens:
@matsbn - I'd like to have this fix as part of the sandbox that I'm putting together for you, Thomas, Ana and Jens for Making Waves.

matsbn:
Since BLOM master is currently supporting NorESM based on CESM2.1 to CESM2.3 frameworks, I wanted to test this PR with NorESM2 (based on CESM2.1). The modified buildcpp works well in NorESM2, although the fix is redundant in this configuration (-DPNETCDF flag is duplicated when pnetcdf module is loaded).

If I remember correctly, I got the -DPNETCDF flag set correctly with cesm2.3.alpha08a code base by modifying cime/scripts/Tools/Makefile to include something like:

ifdef LIB_PNETCDF
   CPPDEFS += -D_PNETCDF
   SLIBS += -L$(LIB_PNETCDF) -lpnetcdf
   ifeq ($(COMP_NAME), blom)
      CPPDEFS += -DPNETCDF
   endif
endif

Also setting the cmake specific netcdf/pnetcdf settings for Betzy was important. Would you think that is a better place to set CPP flags for external libraries?

mvertens:
Hi Mats,

Thanks for this. I think our messages crossed.

If possible we are trying to keep component specific changes for things
like CPP variables out of CIME to the greatest extent possible.
In the Makefile there are still component specific settings for libraries
and includes. The whole build process in CESM is hoping to evolve to CMAKE
and not be Makefile based the way it is now. Since blom already has a
buildcpp file - it made sense to me to just add this logic in there.
If you feel that we should place it in CIME - I can do that - but it would
be good to chat about how we want to approach changes like this moving
forwards.

Mariana

matsbn:
Thanks for the clarification. I agree we should approach a consistent procedure on how to set library specific settings across components, and I guess it is a bit of a moving target right now. Until we agree on how to go about this, this fix should work for now so I'll merge it.

PR #253, opened 2023-05-15
New mesoscale eddy diffusivity options

matsbn:
These options, by @AleksiNummelin, include 2D structure of diffusivity, anisotropy correction of diffusivity and to use the minimum of planetary and topographic beta to define the Rhines scale.

Default namelist options for mesoscale eddy diffusivity have been preserved, and with these options results are unchanged.

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -178,6 +178,7 @@ set DKTZL                  = 1
 set EITMTH   = "'gm'"
 set EDRITP   = "'large scale'"
 set EDWMTH   = "'smooth'"
+set EDDF2D   = .false.
 set EDSPRS   = .true.
 set EGC      = 0.85
 set EGGAM    = 200.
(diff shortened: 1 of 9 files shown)


AleksiNummelin (approved):
The changes that are part of this PR are tested and documented in a manuscript that is under review for JAMES and available as pre-print from https://doi.org/10.22541/essoar.168394750.04852652/v1

The tested combination of the new flags used in that manuscript are the following:

control 2D:
eddf2d=.true.
edanis=.false.
rhsctp=.false.
redi3d=.false.
edsprs=.false.

2D with topographic Rhines scale:
eddf2d=.true.
edanis=.false.
rhsctp=.true.
redi3d=.false.
edsprs=.false.

2D with topographic Rhines scale and eddy efficiency (edanis):
eddf2d=.true.
edanis=.true.
rhsctp=.true.
redi3d=.false.
edsprs=.false.

Other combination can also be used, although they are not as much tested. However, setting

edanis=.true.
edsprs=.true.

is not reasonable and should be avoided (both act to suppress diffusivity in rather similar manner). At later state a check for a reasonable combination of flags should be added.

PR #256, opened 2023-06-14
work around compile-time error in iHAMOCC when debug mode is activated

### LINKED ISSUE #254: BLOM is not compiling in DEBUG mode
I am getting the following errors when compiling BLOM in DEBUG model without  carbon isotopes 

```
mo_param1_bgc.F90(264): error #5561: Subscript #1 of the array MAP_POR2OCTRA has value -1 which is less than the lower bound of 1
mo_param1_bgc.F90(265): error #5561: Subscript #1 of the array MAP_POR2OCTRA has value -1 which is less than the lower bound of 1
```

I believe that the fix would be to introduce the following (see the <= new lines)
```F90
#ifdef cisonew <=== new
        if(ipowc13 > 0) map_por2octra(ipowc13) = isco213
	if(ipowc14 > 0) map_por2octra(ipowc14) = isco214
#endif <=== new
```

I have verified that the code now compiles in DEBUG mode with this change.

JorgSchwinger:
In debug mode the intel compiler generates an error in mo_param1 (see #254 by @mvertens). Apparently, since indices are defined as parameters, the compile-time array bounds check generate errors for actually unused indices even though the code cannot be reached. Since the intel compiler is our workhorse I suggest to fix this by using pre-processor directives instead of if-statements.

### DIFF
--- a/hamocc/mo_param1_bgc.F90
+++ b/hamocc/mo_param1_bgc.F90
@@ -259,10 +259,10 @@
         map_por2octra(ipown2)  = igasnit 
         map_por2octra(ipowno3) = iano3 
         map_por2octra(ipowasi) = isilica 
-       
-        ! if statements for non-base tracers 
-        if(ipowc13 > 0) map_por2octra(ipowc13) = isco213 
-        if(ipowc14 > 0) map_por2octra(ipowc14) = isco214
+#ifdef cisonew
+        map_por2octra(ipowc13) = isco213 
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):
Looks fine to me :+1:

jmaerz (approved):
Dear @JorgSchwinger , thanks for taking care of this compile time error. I approve.

PR #260, opened 2023-06-28
Merging recent master developments into the beyond-CMIP6 branch to keep it up-to-date 

jmaerz:
Hi, to keep the beyond-CMIP6 branch with recent developments, I would like to merge `master` into  the `beyond-CMIP6` branch.


JorgSchwinger (approved):
Thanks for keeping the beyond-CMIP branch up to date. As always, it is impossible to really review the changes in detail, but I assume that this merge will do the correct thing.

TomasTorsvik (approved):
Thanks @jmaerz ! I think the main change in iHAMOCC is expansion of the alkalinization routine with more scenarios, and that the default value for `addalk` has been changed. I don't see any major problems.

PR #261, opened 2023-07-04
Feature iHAMOCC nml

jmaerz:
Hi @TomasTorsvik  and @JorgSchwinger , I create this draft pull request to follow up on and move the discussion https://github.com/NorESMhub/BLOM/discussions/195 based on the current changes I did for the BGC parameter nml. I restructured the parameter initialization a bit related to the different topics of the parameters.

I currently see a few related questions:
1. How shall we deal with parameters calculated from other (tunable) parameters? - shall we introduce functions or shall we ONLY calculate them after having read the namelist?
2. Shall we continue putting atmospheric concentrations (as external controlling parameters) into the former BGC namelist, where `atm_co2` is still located? - this would enable to recombine the further calculation of dependent concentrations (isotopes) and initializing the atmosphere fields.
3. I thus far only introduced TYPICAL tuning parameters in the new nml. I personally would keep it like this or do you have a different opinion?
4. I would like to delete the line breaks in the write parameter routine - any objections?  
5. I still need to delete the old subroutine that is still present (but not used anymore).

Thus far, I didn't move any of the parameter declarations to get the parameter values under the `protect` statement. I can do this, if you want me to do it. Just let me know.


hamocc/mo_parambgc_ini.F90
```diff
@@ -0,0 +1,1241 @@
...
+#ifdef cisonew
+  use mo_biomod,      only: bifr13,bifr14,c14fac,prei13,prei14,re1312,re14to
+  use mo_carbch,      only: atm_c13, atm_c14,c14_t_half,c14dec
+#endif
+#ifdef natDIC
+  use mo_carbch,      only: atm_co2_nat
+#endif
+
+  implicit none
+
```
JorgSchwinger: A thought from my side: Having the variables that qualify as "parameters" defined in the module header of this module would simplify things a lot. Those variables that are truly "parameters" could be defined as such and assigned a value (that isn't changed), e.g.

  `integer, parameter :: rcar =122`

Parameters that can be modified via namelist could be assigned a default value, e.g.

  `real, save  :: atm_co2 = 284.32`

Then the problem of having to recalculate things after reading of the namelist would be obsolete, and the module would be much simpler (plus it would be much more transparent to the user - the fact that all parameter definition are scattered between different modules now isn't very nice in my opinion.

Of course this is more invasive and much more work (a lot of use statements would need to be changed).

hamocc/mo_parambgc_ini.F90
```diff
@@ -0,0 +1,1241 @@
...
+      atm(i,j,iatmc13)  = atm_c13
+      atm(i,j,iatmc14)  = atm_c14/c14fac
+#endif
+#ifdef BROMO
+      atm(i,j,iatmbromo)= atm_bromo
+#endif
+    ENDDO
+    ENDDO
+  end subroutine
+
```
JorgSchwinger: These 3 subroutines above could be simplified and combined into one routine if we would follow the idea of defining all parameters inside this module (as opposed to importing them)

hamocc/mo_parambgc_ini.F90
```diff
@@ -0,0 +1,1241 @@
...
+    !
+    ! Read the bgcparams namelist for parameter tuning.
+    ! Note that afterward, i) rates need to be adjusted for timestep
+    ! and some depending parameters need re-calculation
+    !
+
+    integer  :: iounit
+
+    namelist /bgcparams/ bkphy,dyphy,bluefix,bkzoo,grazra,spemor,gammap,gammaz,ecan,zinges,epsher,bkopal,rcalc,ropal,              &
+                       & remido,drempoc,dremopal,dremn2o,dremsul,fetune,relaxfe,wpoc,wcal,wopal
```
JorgSchwinger: `wpoc` would only be used directly if the old HAMOCC sinking (constant wpoc) is used. Shouldn't we rather include wmin/wmax/wlin of the WLIN scheme as tuning parameters? (I would even be tempted to remove the constant sinking  completely from HAMOCC, since this produces crap results)

hamocc/mo_parambgc_ini.F90
```diff
@@ -0,0 +1,1241 @@
...
+ 
+  !---------------------------------------------------------------------------------------------------------------------------------
+  subroutine readjust_param()
+    !
+    !  AFTER reading the namelist:
+    ! re-adjust parameters that depend on tuning parameters
+    !
+    perc_diron = fetune * 0.035 * 0.01 / 55.85
+
+  end subroutine
```
JorgSchwinger: If this remains the only parameter to adjust, I think it is better to just have this line in the main routine instead of making a separate subroutine.

hamocc/mo_parambgc_ini.F90
jmaerz: Hi @JorgSchwinger , I -partially- agree - the calculation and field initialization need to be carried out after reading the namelist, while the initial default initialization could be done before.

hamocc/mo_parambgc_ini.F90
jmaerz: I am totally happy to extend the parameter list of the namelist. So far, I saw the new namelist as extendable and only for relevant parameters.

hamocc/mo_parambgc_ini.F90
jmaerz: I am not sure, if I understand, what you consider the main routine here. I guess, my intention was to provide a structure , which one can follow.

hamocc/mo_parambgc_ini.F90
jmaerz: Hi @JorgSchwinger , seeing the upcoming pull request in #264 (and potentially my nitrogen cycle), I currently would opt for a step-wise approach, so first introducing this less invasive step, and after pulling in Marianas changes doing a second step and following your suggestion, potentially even go the step with the `protect` statement, what @TomasTorsvik suggested. Let's discuss this in person.

hamocc/beleg_vars.F90
```diff
@@ -28,7 +28,7 @@ SUBROUTINE BELEG_VARS(kpaufr,kpie,kpje,kpke,kbnd,pddpo,prho,omask,      &
 !  Modified
 !  --------
 !  J.Schwinger,        *NORCE Climate, Bergen*    2020-05-19
-!   -split the original BELEG_BGC in two parts, BELEG_PARM and BELEG_VARS
+!   -split the original BELEG_BGC in two parts, BELEG_PARM (NOWMO_PARAMBGC_INI) and BELEG_VARS
```
TomasTorsvik:
```suggestion
!   -split the original BELEG_BGC in two parts, BELEG_PARM (NOW MO_PARAMBGC_INI) and BELEG_VARS
```
missing space

TomasTorsvik (approved):
Looks fine for me

hamocc/beleg_vars.F90
jmaerz: Done.

hamocc/mo_parambgc_ini.F90
jmaerz: Done. I left the constant sinking, though - can be discussed afterward in an extra PR.

hamocc/mo_parambgc_ini.F90
jmaerz: I would leave it for now and re-think about it, once introducing the `protect` statements after #264 was coming in.

jmaerz:
To briefly summarize the conversation on this PR: To avoid too many merging conflicts, the potential restructuring wrt `protect` statements and pulling all parameters into one module is envisaged in an extra PR, once #264 is merged into `master`.

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , I tested the branch against the current `master` using the `NOINYOC` `T62_tn21` setup and ran the model for 1 year and checked successfully for binary identical results. 

**How to use the new feature (thus far, tested):**
After model setup (`./case.setup`), copy the `./components/blom/cime_config/buildnml` to the `./SourceMods/src.blom/` folder (no further subfolders needed). In this copy of `buildnml`, enter the parameters that you wish to change in the `BGCPARAMS` namelist (for available parameters, see code, potentially add it, if needed - careful with rates and parameters that are required for further calculation of other parameters - make sure that they are converted and applied after reading the namelist - currently, this should be ensured, but with new developments, this needs care-taking). Parameters available through the `user_nl_blom` are still possible to change via an additionally provided user namelist for BLOM (so nothing changed in this regards). 

I will perform the pull once the `boxatm` change #265 has been merged into `master`. Afterward, I will also merge these changes into the `feature-HAMOCC_beyond-CMIP6` branch. Once the `ifdefs` refactoring #264 are merged into `master`, we can proceed with restructuring the parameter declaration and initialization.

cime_config/ocn_in.readme
```diff
@@ -376,7 +376,13 @@
 ! SEDPORFILE  : File name (incl. full path) for sediment porosity
 !
 !===========================================================================
-! NAMELIST FOR ALKALINIZATION SCENARIO
+! NAMELIST BGCPARAMS FOR iHAMOCC-BGC PARAMETERS (DEVELOPERS ONLY)
+! 
+! CONTENTS: EMPTY BY DEFAULT
+!           FOR ADJUSTABLE PARAMETERS, SEE CODE
```
JorgSchwinger, line 382: Once we have settled on a set of parameters, it would be good to list those here

hamocc/hamocc_init.F90
```diff
@@ -48,6 +48,7 @@ subroutine hamocc_init(read_rest,rstfnm_hamocc)
        &                    dtb,dtbgc,io_stdo_bgc,ldtbgc,                       &
        &                    ldtrunbgc,ndtdaybgc,with_dmsph,l_3Dvarsedpor
   use mo_param1_bgc,  only: ks,init_por2octra_mapping
+  use mo_parambgc_ini,only: ini_parambgc
```
JorgSchwinger: Since we plan to have all parameters defined within the module (in the longer run), the name `mo_parambgc_ini` is not really the best, in my opinion. `mo_param_bgc` would be better (but that is too close to `mo_param1_bgc`?). Not sure what to do about it.

hamocc/mo_parambgc_ini.F90
```diff
@@ -0,0 +1,686 @@
...
+    d14cat   = (prei14+alpha14)/(1.-alpha14/1000.)
+    ! calculate atm_c13 and atm_c14
+    atm_c13  = beta13*re1312*atm_co2/(1.+beta13*re1312)
+    d13C_atm = (((atm_c13/(atm_co2-atm_c13))/re1312)-1.)*1000.
+    ! absolute 14c concentration in preindustrial atmosphere
+    atm_c14  = ((d14cat/1000.)+1.)*re14to*atm_co2
+    ! factor for normalizing 14C tracers (~1e-12)
+    c14fac   = atm_c14/atm_co2
+#endif
+  end subroutine
```
JorgSchwinger: A matter of personal taste, but I would prefer to have not too many small subroutines. I would place the calculations done here in other routines: The 3 first lines could go into  `ini_param_atm` and the rest into `ini_fields_atm` (`d13C_atm` is only used to print out the value, so it could go into `write_parambgc`)

hamocc/mo_parambgc_ini.F90
```diff
@@ -0,0 +1,686 @@
...
+    call ini_aggregation()         ! Initialize aggregation module of Iris Kriest (no NML read thus far)
+#endif
+
+    call read_bgcnamelist()        ! read the BGCPARAMS namelist
+    call calc_param_atm()          ! calculate atmospheric parameters after updating parameters via nml
+    call ini_fields_atm(kpie,kpje) ! initialize atmospheric fields with (updated) parameter values 
+    call readjust_param()          ! potentially readjust namlist parameter-dependent parameters
+    call rates_2_timestep()        ! Converting rates from /d... to /dtb
+
+    call write_parambgc()          ! write out used parameters and calculate back rates from /dtb to /d..
```
JorgSchwinger: Wouldn't it make more sense to write out the parameter BEFORE multiplying by `dtb`? If I want to check the settings of parameters in the log file, I relate much more to the value that I have set via namelist than the value multiplied by `dtb`.

JorgSchwinger (approved):

jmaerz:
To briefly summarize (incl. the short-conversation today), before merging: 
- After merging, and once #264 has been merged in, we can i) implement the `protect` statements and ii) potentially slightly restructure `mo_param_bgc` wrt to the number of subroutines. This includes potentially shifting the initialization of the atmosphere field into `beleg_vars`. Further `mo_param1_bgc` could be renamed to `mo_param_tracer` in that process.
- In `write_parambgc`, rates are multiplied with an inverse of `dtbgc` so that the namelist parameter is comparable - not refinding the nml-provided parameter give rise to an intended double-check.
- Wrt to explicitly mentioning the BGC nml parameters: we can discuss this again. I would tend to not state them explicitly.

PR #264, opened 2023-08-01
remove most hamocc #ifdefs in favorite of new logicals

mvertens:
This is a draft PR that is a **prototype** implementation to move away from #ifdefs in BLOM and particularly in HAMOCC to the greatest extent possible. Most modeling efforts have adopted the approach to have the least number of #ifdefs possible.
This is a result of the fact that the code itself is not really tested until all possible #ifdef combinations are invoked. The more #ifdefs there are - the greater is the challenge to do this. By minimizing #ifdefs - the compilation can at least detect if there are inconsistencies in various branches that would always be compiled regardless of if they are actually invoked run time.

The approach I am suggesting is to introduce a new module in hamocc - `mo_ifdefs.F90` that contains logicals that are parameters and that are triggerred by the presence of the #ifdefs. 
```F90
#ifdef BROMO
  logical, parameter :: use_BROMO = .true.
#else
  logical, parameter :: use_BROMO = .false.
#endif
```

In the code itself - the majority of the #ifdefs are then replaced with logicals like (e.g. in aufw_bgc.F90):
```F90
       if (use_BROMO) then
         CALL write_netcdf_var(ncid,'bromo',locetra(1,1,1,ibromo),2*kpke,0)
      end if
```
Note that the work around to deal with ibromo values of -1 (when #BROMO is not defined) is to define loctre as
```F90
      REAL :: locetra(kpie,kpje,2*kpke,-1:nocetra) ! local array for reading
```
But to initialize locetra as 
```F90
      locetra(:,:,:,1:)=trc(1:kpie,1:kpje,:,itrbgc:itrbgc+ntrbgc-1)
```
Multiple compilation errors were found when this was implemented. Below is a brief summary

The following were not defined when the ifdefs were removed - which means they would not have worked if the ifdefs were turned on
```
beleg_vars.F90(254): error #6404: This name does not have a type, and must have an explicit type.   [IPOWC13]
beleg_vars.F90(255): error #6404: This name does not have a type, and must have an explicit type.   [IPOWC14]
beleg_vars.F90(256): error #6404: This name does not have a type, and must have an explicit type.   [ISSSO13]
beleg_vars.F90(257): error #6404: This name does not have a type, and must have an explicit type.   [ISSSO14]
beleg_vars.F90(258): error #6404: This name does not have a type, and must have an explicit type.   [ISSSC13]
beleg_vars.F90(259): error #6404: This name does not have a type, and must have an explicit type.   [ISSSC14]

[...]

PR #265, opened 2023-08-02
Add file mo_boxatm.F90 

JorgSchwinger:
making it possible to run HAMOCC with a well-mixed 1D box atmosphere for air-sea exchange of CO2 (and C-isotopes). There is currently no way to activate this option other than adding the preprocessor flag -DBOXATM in cime_config/buildcpp.

I have tested that it compiles and runs as expected (bit-for-bit if BOXATM is deactivated). Results with BOXATM look reasonable, but are not validated beyond this.

### DIFF
--- a/hamocc/hamocc4bcm.F90
+++ b/hamocc/hamocc4bcm.F90
@@ -314,7 +314,7 @@
 
       ! Update atmospheric pCO2 [ppm]
 #if defined(BOXATM)
-      CALL update_boxatm(kpie,kpje,pdlxp,pdlyp)
+      CALL update_boxatm(kpie,kpje,kpke,pdlxp,pdlyp,pddpo,omask)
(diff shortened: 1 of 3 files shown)


JorgSchwinger:
When testing compiling with C-isotopes and BOXATM I encountered (some of) the same errors as @mvertens in #264. I did not include fixes to this errors in this PR to avoid later conflicts when merging in #264

TomasTorsvik (approved):
@JorgSchwinger - Thanks for including the missing module.

jmaerz (approved):
Hi @JorgSchwinger , also thanks from my side. I think, the file needs to be added to `./hamocc/meson.build` to make it accessible to 1D setups. After this, approved.

mvertens (approved):
Thanks for this PR.

PR #266, opened 2023-08-03
Add capability to run C-isotopes with sediment code switched on

JorgSchwinger:
This is the PR that @jmaerz has mentioned in #264 - it is slightly more than "minor changes to sedshi.F90", so it will most probably create merge conflicts when merging with #264. Up to you, @mvertens to decide what order to follow when merging this, please let me know.

This PR has been tested technically but has not been validated beyond this. The build script will still refuse to run this combination of options, but it would be good to have this code in already now.


mvertens (approved):
I think this PR should go in next.

TomasTorsvik (approved):
Thanks @JorgSchwinger - It's good to get in these changes now, even though it may not be the final version.

jmaerz (approved):
Dear @JorgSchwinger , many thanks for the effort to bring this in! Approved.

PR #273, opened 2023-09-20
Fix potential OMP issues and bug fix in wrapping the burial fields

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , during my coding and checking the SIunits4HAMOCC, I found another potential race condition during OMP parallelization (in `mo_vgrid`) in addition to the known one in the sinking scheme that I now also fixed (see also #243 and #244) - for the sinking in `ocprod.F90`, I understood that `tco,tcn,q` should be private within the `i`, `j` loops (right?), hence I also added those as private in the OMP statement. Further, there was a bug in the wrapper related to burial fields - the `k`-loop was causing to time-smooth the `burial2` field multiple times. Please have a closer look, since this pull request is potentially/likely answer changing (hence the pull request for the `beyond-CMIP6` branch). As a side note: after these fixes, it was possible to run the single column setup **with** OMP and get bit identical results (which wasn't the case before - while for speed reasons, disabling OMP in single column mode is advisable).  

I still have to run a test run on betzy (somehow, I have currently -ongoing- access problems - I contacted sigma2).

### DIFF
--- a/hamocc/dipowa.F90
+++ b/hamocc/dipowa.F90
@@ -84,8 +84,7 @@ subroutine dipowa(kpie,kpje,kpke,omask,lspin)
 !ik needed for boundary layer ventilation in fast sediment routine
   real :: bolven(kpie)                 ! bottom layer ventilation rate
 
-!$OMP PARALLEL DO                            &
-!$OMP&PRIVATE(i,k,iv,l,bolven,tredsy,sedb1,aprior,iv_oc)
+!$OMP PARALLEL DO PRIVATE(i,k,iv,l,bolven,tredsy,sedb1,aprior,iv_oc)
   j_loop: do j=1,kpje
 
 ! calculate bottom ventilation rate for scaling of sediment-water exchange
@@ -206,5 +205,5 @@ subroutine dipowa(kpie,kpje,kpke,omask,lspin)
   endif ! .not. lspin
   
   enddo j_loop
-
+!$OMP END PARALLEL DO
 end subroutine dipowa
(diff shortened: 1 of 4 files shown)


TomasTorsvik (approved):
Thanks, Jöran. I think you are right about the potential race condition (although I suspect it mainly shows up when running single column).

One question, maybe more code style: The "ORDERED" clause force sequential execution of a code block, while the code outside "ORDERED" can run in parallel. If a code block is intended to be executed sequentially, would it not be better to remove the openMP directives all together for this part of the code?

jmaerz:
Hi Tomas, after reading on the OMP ORDERED clause, I would expect that in our case, only the `k`-loop is forced to be in sequential mode, while the lateral (`i`, `j`) loops can still be parallelized - which is exactly the behavior that we would want to have, I would say. The initial ORDERED clause only tells that within the following code block, there will be one part, which should be calculated in ORDERED mode - then, this code part is defined **within** the `k`-loop via `OMP ORDERED`. But correct me, if I am wrong.

But in general, I agree, thus far, the reproducible results on betzy and fram point towards no effect of this potential race condition - while I feel it is good to rule them out.

JorgSchwinger (approved):
Looks fine to me, thanks for fixing this! 

Regarding bit-for-bit: If OMP is not used this shouldn't introduce any changes, should it? So we could also put this fix into master?

JorgSchwinger:
Ah, ok sorry, I see now that it was a real bug, and won't be bit-for-bit.

TomasTorsvik:
@jmaerz - I agree with you about how `OMP ORDERED` works, and that you get parallel performance for the `(i, j)` loops. It's mainly a question if we should dive deeper into the openMP "rabbit hole" or keep things simple. But you got it working now, so I'm happy to keep this solution.

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , thanks for the feedback - I'll perform a short test simulation later today and would expect at least changes in the burial fields, when compared to the `beyond-CMIP6` branch. @TomasTorsvik , where would it be to change the OMP threads settings for betzy - do you remember? - then I could have a test run with changed OMP settings as well (for now out of curiosity). Once the runs look reasonable, I'll perform the PR.

PR #277, opened 2023-10-03
add riverine C-isotope input

monsieuralok:


jmaerz:
Hi @monsieuralok , many thanks for bringing in these changes! - just one thing: all the `sedbypass` flags in `powach` and `sedshi` are not really needed, since `powach` and `sedshi` are only called, when `sedbypass=.false.` (see `hamocc4bgm.F90`).

monsieuralok:
> Hi @monsieuralok , many thanks for bringing in these changes! - just one thing: all the `sedbypass` flags in `powach` and `sedshi` are not really needed, since `powach` and `sedshi` are only called, when `sedbypass=.false.` (see `hamocc4bgm.F90`).

@jmaerz Jerry mentioned that it could be we can have both TRUE also that what we checked today and yesterday. But, you should also check it.

monsieuralok:
> > Hi @monsieuralok , many thanks for bringing in these changes! - just one thing: all the `sedbypass` flags in `powach` and `sedshi` are not really needed, since `powach` and `sedshi` are only called, when `sedbypass=.false.` (see `hamocc4bgm.F90`).
> 
> @jmaerz Jerry mentioned that it could be we can have both TRUE also that what we checked today and yesterday. But, you should also check it.

sorry. I understand what you mentioned. I will update.

JorgSchwinger (approved):
I agree with Jöran that the changes in powach and sedchi are not necessary - these routines are only called when sediment-bypass is disabled. Otherwise I can approve this - thanks for bringing this in!

monsieuralok:
> I agree with Jöran that the changes in powach and sedchi are not necessary - these routines are only called when sediment-bypass is disabled. Otherwise I can approve this - thanks for bringing this in!

@JorgSchwinger @jmaerz It would be then good to pick file mo_apply_rivin.F90 only.  Or should I send it again?

jmaerz:
Hi @monsieuralok , I would suggest to revert the changes for files `sedshi` and `powach` and then push again to the same branch. Either just copy the original files or there seem to be the git way (I just found it on stackoverflow: https://stackoverflow.com/questions/2620775/revert-changes-to-a-file-in-a-commit):
```
git show some_commit_sha1 -- some_file.c | git apply -R
```
(and add + commit + push to the same github branch) Hope that helps. Thanks again for bringing the stuff in!

JorgSchwinger:
Yes would be easiest if you could update this PR, then we can just merge it from here.

monsieuralok:
> Hi @monsieuralok , I would suggest to revert the changes for files `sedshi` and `powach` and then push again to the same branch. Either just copy the original files or there seem to be the git way (I just found it on stackoverflow: https://stackoverflow.com/questions/2620775/revert-changes-to-a-file-in-a-commit):
> 
> ```
> git show some_commit_sha1 -- some_file.c | git apply -R
> ```
> 
> (and add + commit + push to the same github branch) Hope that helps. Thanks again for bringing the stuff in!

@JorgSchwinger @jmaerz I have updated my repo

PR #280, opened 2023-10-13
remove #ifdefs from HAMOCC

mvertens:
This PR is a substitute for #264 (which I would like to close). It does not build upon the ability for CMEPS to calculate DMS and BROMO fluxes. 
The current testing I have done is to compare relative to the latest baseline:
1. ./create_test  --xml-category aux_blom_noresm --xml-machine betzy --xml-compiler intel --gen aux_blom_noresm_13102023 --baseline-root /cluster/shared/noresm/noresm_baselines --compare aux_blom_noresm_27092023 --project nn2345k
The following tests were run:
  ERP_Ld3.T62_tn14.NOINY.betzy_intel  - passed
  ERS_Ld3.T62_tn14.NOINYOC.betzy_intel - passed
  ERS_Ld3.T62_tn14_wtn14nw.NOINY_WW3.betzy_intel - failed baseline comparison
  ERS_Ld3.T62_tn14_wtn14nw.NOINY_WW3.betzy_intel.blom-wavice - failed baseline comparison
  SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel - passed
I think there is something problematic with the current ww3dev code that is leading to failures in the baseline comparison. The same tests without ww3 passed with no problem. I will raise an issue on this - but I don't think this should hold up this PR.
2. Ran two tests using noresm2.0.6 - SMS.T62_tn14.NOINYOC.betzy_intel. The first used used BLOM master and the second tests used this PR for BLOM. The results were bfb after 5 days.
3. Ran two fully coupled tests using noresm2.0.6 - SMS.f19_tn14.N1850esm.betzy_intel. The first used used BLOM master and the second tests used this PR for BLOM. The results were bfb after 5 days.


mvertens:
@JorgSchwinger - I ran a fully coupled test in the noresm2.0.6 code base - as was suggested-  and the results were bfb.

cime_config/buildnml.csh
JorgSchwinger, line 1: This file was removed previously, I believe it shouldn't be re-introduced through this PR?

hamocc/aufr_bgc.F90
```diff
@@ -466,175 +460,158 @@ SUBROUTINE AUFR_BGC(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,               &
...
+         CALL read_netcdf_var(ncid,'powaic',powtra2(1,1,1,ipowaic),2*ks,0,iotype)
+         CALL read_netcdf_var(ncid,'powaal',powtra2(1,1,1,ipowaal),2*ks,0,iotype)
+         CALL read_netcdf_var(ncid,'powaph',powtra2(1,1,1,ipowaph),2*ks,0,iotype)
+         CALL read_netcdf_var(ncid,'powaox',powtra2(1,1,1,ipowaox),2*ks,0,iotype)
+         CALL read_netcdf_var(ncid,'pown2',powtra2(1,1,1,ipown2),2*ks,0,iotype)
+         CALL read_netcdf_var(ncid,'powno3',powtra2(1,1,1,ipowno3),2*ks,0,iotype)
+         CALL read_netcdf_var(ncid,'powasi',powtra2(1,1,1,ipowasi),2*ks,0,iotype)
+         if (use_cisonew) then
+            IF(lread_iso) THEN
+               ! Burial fields for c-isotopes still missing
```
JorgSchwinger: This comment shouldn't be re-introduced, the burial fields for C-isotopes are now in place (`bur_*` below)

hamocc/aufr_bgc.F90
```diff
@@ -466,175 +460,158 @@ SUBROUTINE AUFR_BGC(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,               &
...
-        powtra2(i,j,k,ipowc14)=powtra2(i,j,k,ipowaic)*rco214
-        sedlay2(i,j,k,issso13)=sedlay2(i,j,k,issso12)*rco213*bifr13
-        sedlay2(i,j,k,issso14)=sedlay2(i,j,k,issso12)*rco214*bifr14
-        sedlay2(i,j,k,isssc13)=sedlay2(i,j,k,isssc12)*rco213
-        sedlay2(i,j,k,isssc14)=sedlay2(i,j,k,isssc12)*rco214
-        ENDIF
-      ENDDO
-      ENDDO
-      ENDDO
-
```
JorgSchwinger, line 612: This loop is now missing!!! Needs to be added again

hamocc/aufr_bgc.F90
```diff
@@ -466,175 +460,158 @@ SUBROUTINE AUFR_BGC(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,               &
...
+                     powtra2(i,j,k,ipowc13)=powtra2(i,j,k,ipowaic)*rco213
+                     powtra2(i,j,k,ipowc14)=powtra2(i,j,k,ipowaic)*rco214
+                     sedlay2(i,j,k,issso13)=sedlay2(i,j,k,issso12)*rco213*bifr13
+                     sedlay2(i,j,k,issso14)=sedlay2(i,j,k,issso12)*rco214*bifr14
+                     sedlay2(i,j,k,isssc13)=sedlay2(i,j,k,isssc12)*rco213
+                     sedlay2(i,j,k,isssc14)=sedlay2(i,j,k,isssc12)*rco214
+                  ENDIF
+               ENDDO
+               ENDDO
+               ENDDO
```
JorgSchwinger, line 589: Loop for burial variables should be added here (see comment above)!

hamocc/aufw_bgc.F90
```diff
@@ -164,13 +158,17 @@ SUBROUTINE AUFW_BGC(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,              &
...
+! are passed into the local array locetra; No unit conversion here,
+! tracers in the restart file are written in mol/kg
 !--------------------------------------------------------------------
 !
       testio=0
-      locetra(:,:,:,:)=trc(1:kpie,1:kpje,:,itrbgc:itrbgc+ntrbgc-1)
+!
+! Allocate and initialize local array for writing (locetra)
+!
+      locetra(:,:,:,:) = trc(1:kpie,1:kpje,:,itrbgc:itrbgc+ntrbgc-1)  ! TODO - is this right?
```
JorgSchwinger: This should be correct (the trc array carries 2 time-levels of tracer values in the 3rd dimension, these are both copied into locetra here). 

Not sure though - wasn't it necessary to have locetra as allocatable array? If it works it's fine

hamocc/carchm.F90
```diff
@@ -137,9 +135,11 @@ SUBROUTINE CARCHM(kpie,kpje,kpke,kbnd,                                  &
       REAL,    intent(in) :: pfu10(1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd)
       REAL,    intent(in) :: ptho(1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd,kpke)
       REAL,    intent(in) :: psao(1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd,kpke)
+      REAL,    intent(in) :: pflxdms(1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd)
+      REAL,    intent(in) :: pflxbromo(1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd)
 
       ! Local variables
-      INTEGER :: i,j,k,l,js
+      INTEGER :: i,j,k,l,js,ks
```
JorgSchwinger: ks isn't a local variable, it needs to be imported from `mo_param1_bgc`

hamocc/hamocc_init.F90
```diff
@@ -47,7 +47,7 @@ subroutine hamocc_init(read_rest,rstfnm_hamocc)
        &                    sedspin_yr_s,sedspin_yr_e,sedspin_ncyc,             &
        &                    dtb,dtbgc,io_stdo_bgc,ldtbgc,                       &
        &                    ldtrunbgc,ndtdaybgc,with_dmsph,l_3Dvarsedpor
-  use mo_param1_bgc,  only: ks,init_por2octra_mapping
+  use mo_param1_bgc,  only: ks,init_por2octra_mapping,init_indices
```
JorgSchwinger: `init_indices` doesn't seem to be used?

cime_config/buildnml.csh
mvertens: Done.

hamocc/aufr_bgc.F90
mvertens: Done.

hamocc/aufr_bgc.F90
mvertens: Done. Thanks for catching this. It was not picked up in my git merge.

hamocc/aufw_bgc.F90
mvertens: @JorgSchwinger - locetra works now as a non-allocatable array on the stack.

hamocc/carchm.F90
mvertens: Thanks for catching this. I'm concerned that the testing we have is not really covering all the use cases that we need to have identified these problems. It would be good to add a few tests that would create baselines and would have caught these problems as part of the regression testing. A quick chat to out outline what we want to do moving forwards would be very helpful. I'm happy to implement that.

hamocc/mo_control_bgc.F90
```diff
@@ -16,86 +16,151 @@
...
 
-! File containing namelists
-      CHARACTER(LEN=:), ALLOCATABLE, PROTECTED :: bgc_namelist
+  INTEGER :: ldtbgc                   !  time step number from bgc restart file
+  INTEGER :: ldtrunbgc                !  actual time steps of run.
 
-! Control variables
-      REAL,    save :: dtbgc              !  time step length [sec].
-      REAL,    save :: dtb                !  time step length [days].
-      INTEGER, save :: ndtdaybgc          !  time steps per day.
```
JorgSchwinger, line 49: I think it is actually correct to have save attributes here. Unless a module variable is used in the main program it can go out of scope and the Fortran standard doesn't require that the variable's value is retained (as far as I know).

mvertens: @JorgSchwinger - for module scope my understanding is that save is now on by default.

gold2718: > @JorgSchwinger - for module scope my understanding is that save is now on by default.

This is true. Note that this is a change from the behavior in Fortran 2003 and before. Beginning with Fortran 2008 (and universally true in contemporary compilers to my knowledge), all module-scope variables have the save attribute by default.

trc/mod_tracers.F90
```diff
@@ -100,6 +100,7 @@ subroutine allocate_tracers
 
      ! Number of HAMOCC tracers.
 #ifdef HAMOCC
+     call init_indices()
```
JorgSchwinger, line 103: I would find it preferable if the `init_indices` routine would be called in `hamocc_init`, but I guess this isn't possible because this routine is called first? But we should at least have a comment in `hamocc_init` saying that the initialization of HAMOCC indices is done here.

phy/mod_ifdefs.F90
```diff
@@ -0,0 +1,22 @@
+module mod_ifdefs
```
JorgSchwinger, line 1: I this file needed? BLOM ifdefs aren't touched in this PR?

hamocc/mo_control_bgc.F90
mvertens: @JorgSchwinger - CESM has been leverage this to remove save attributes from module variables in new development versions of the code.

JorgSchwinger (approved):
Thanks a lot for this effort! It's great to get this into HAMOCC! I agree that at least for this very invasive PR a testing with more (or all) of the options activated would be good. I'm happy to have a chat on this.

Maybe also good if @jmaerz or @TomasTorsvik could have a second review. It's a lot of stuff and I might have missed something.

hamocc/mo_control_bgc.F90
JorgSchwinger: Ok, thanks for clarifying this - I see my Fortran knowledge is becoming outdated...

phy/mod_ifdefs.F90
mvertens: You still need this file - since these are the #ifdefs outside of HAMOCC that I have not touched yet.
I would be happy to tackle this - but a lot of the files where these appear are not .F90 files and also not modules.
I'd like to see that happen - but it seemed too invasive for this PR.

trc/mod_tracers.F90
mvertens: Good point. I'll add that statement there.

hamocc/mo_apply_rivin.F90
```diff
@@ -86,16 +86,15 @@ subroutine apply_rivin(kpie,kpje,kpke,pddpo,omask,rivin)
...
                             iano3,iphosph,isilica,isco212,iiron,idoc,idet,      &
                             ialkali,inatsco212,inatalkali
-#ifdef cisonew
+  ! cisonew
   use mo_param1_bgc,  only: idet13,idet14,idoc13,idoc14,isco213,isco214,safediv
-#endif
-
   use mo_vgrid,       only: kmle
   use mo_carbch,      only: ocetra,rivinflx
+  use mo_control_bgc, only: use_natDIC
```
jmaerz, line 97: minor, but could be added to l. 89 above

.github/workflows/ci.yml
TomasTorsvik, line 1: This file is removed in commit 5957d2c25a31705778a6fb7dae06e8d0513139e4
What was the background for this? If it is not working, is there some way to disable CI functionality instead of removing it completely?

trc/restart_trcrd.F90
```diff
@@ -32,8 +32,12 @@ subroutine restart_trcrd(rstfnm_ocn)
 
   logical :: error
   character(len=256) :: rstfnm_ocntrc
-#ifdef HAMOCC
+  ! HAMOCC
   character(len=256) :: rstfnm_hamocc
+#ifdef HAMOCC
+  logical :: use_hamocc = .true.
+#else
+  logical :: use_hamocc = .false.
```
TomasTorsvik, line 40: Would it make sense to move this to `mod_ifdefs.F90`, and make use of `use_hamocc`in both `restart_trcrd.F90` and `restart_wt.F`? Not sure if this is a good idea, if the long term plan is to get rid of `mod_ifdefs.F90`.

hamocc/carchm.F90
```diff
@@ -20,7 +20,8 @@
 
       SUBROUTINE CARCHM(kpie,kpje,kpke,kbnd,                                  &
                         pdlxp,pdlyp,pddpo,prho,pglat,omask,                   &
-                        psicomo,ppao,pfu10,ptho,psao)
+                        psicomo,ppao,pfu10,ptho,psao,                         &
+                        pflxdms,pflxbromo)
 !******************************************************************************
```
jmaerz, line 25: I would suggest to keep the `pflxdms` and `pflxbromo` out of this PR, since they will be included via the nuopc PR, right? At the moment, they aren't used.

hamocc/inventory_bgc.F90
```diff
@@ -350,32 +353,45 @@ SUBROUTINE INVENTORY_BGC(kpie,kpje,kpke,dlxp,dlyp,ddpo,omask,iogrp)
...
-       &  +zocetratot(ian2o)*2                                                 &
-       &  - sndepflux                                                          &
-       &  +zprorca*rnit                                                        &
-#if defined(BOXATM)
-       &  +zatmn2*ppm2con*2
-#else
-
-       &  +sn2flux*2+sn2oflux*2
-#endif
+  if (use_BOXATM) then
```
jmaerz, line 364: Here and also for the other species: Instead of repeating all contributions, I would suggest to write once the general contributions and then, afterward:
```
if (use_boxatm) then
        totalcarbon = totalcarbon + zatmco2*ppm2con
else
        totalcarbon = totalcarbon + co2flux
endif
```
which is easier to maintain, shorter and clearer, what the difference is.

hamocc/mo_param1_bgc.F90
```diff
@@ -24,247 +24,364 @@ MODULE mo_param1_bgc
...
+     !use mo_control_bgc, only: use_cisonew
+      use mo_control_bgc
+
       implicit none
-      
-      INTEGER, PARAMETER :: ks=12,ksp=ks+1    ! ks: nb of sediment layers
+      public
 
+      INTEGER, PARAMETER :: ks=12,ksp=ks+1    ! ks: nb of sediment layers
       REAL,    PARAMETER :: safediv = 1.0e-25 ! added to the denominator of isotopic ratios (avoid div. by zero)
```
jmaerz, line 50: Two things: I would suggest to give at least one comment per group (e.g.: `! Base tracer` etc. Then, is there any reason not to at least add a `protect` statement for the indices values?

hamocc/inventory_bgc.F90
```diff
@@ -101,8 +100,8 @@ SUBROUTINE INVENTORY_BGC(kpie,kpje,kpke,dlxp,dlyp,ddpo,omask,iogrp)
   real :: zsedlayto(nsedtra)       ! Sum : Sediment layer tracers
   real :: zburial(nsedtra)         ! Sum : Sediment burial tracers
   !--- oceanic tracers
-  real :: zocetratot(nocetra)      ! Sum : Ocean tracers
-  real :: zocetratoc(nocetra)      ! Mean concentration of ocean racers
+  real :: zocetratot(-1:nocetra)   ! Sum : Ocean tracers
```
jmaerz: I am a bit surprised to see the `-1` here. Is it really needed? - any tracers to be calculated should be `>0` (and loops also start at 1, where these values are calculated)

hamocc/mo_param_bgc.F90
```diff
@@ -43,73 +43,65 @@ module mo_param_bgc
...
   public :: ini_parambgc
 
   ! Module-wide parameters (used in more than one subroutine)
-#ifndef AGG
-  REAL :: dustd1, dustd2, dustsink
-#endif
-#ifdef cisonew 
-  REAL :: beta13, alpha14, d14cat, d13c_atm  
-#endif
+  REAL :: dustd1, dustd2, dustsink !AGG
```
jmaerz: I am running into compilation problems - the line with dustd1, dustd2, dustsink needs to be deleted. (They are introduced now via `mo_biomod` - it was formerly a bit twisted written.)

hamocc/carchm.F90
mvertens: Good point. I've removed them.

jmaerz:
Out of curiosity: what was/is the purpose of `ci.yml` that has been removed - is this the setup to do the testing on github? - if so, why was it now removed?

hamocc/inventory_bgc.F90
mvertens: Great idea! I'll implement that.

TomasTorsvik (approved):

jmaerz:
@mvertens , @JorgSchwinger , just to let you know, with #282, we updated the `beyond-CMIP6` branch with the `master` - so that this PR can go next into `master`.

hamocc/mo_param_bgc.F90
mvertens: I don't understand this comment. I have no problems compiling.

jmaerz: Hi @mvertens , `dustd1, dustd2, dustsink` are already included by `use mo_biomod, only:` above so that at least on my side, the compiler complains `Symbol ‘dustd1’ at (1) conflicts with symbol from module ‘mo_biomod’, use-associated at (2)` (compiled locally via meson). Similarly, `dustd2` and `dustsink` are also already defined. As written, it was very twisted before (these were only included via `use` when `AGG` was switched on, while then the local variables were required inside the routine for some initialization). Deletion of the line 
```
  REAL :: dustd1, dustd2, dustsink !AGG
```
should do the job.

mvertens:
As an additional test - I compare master versus this PR for a SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel - but where I had the following xml variables set to TRUE:
	HAMOCC_CFC: TRUE
	HAMOCC_CISO: TRUE
	HAMOCC_NATTRC: TRUE
	HAMOCC_SEDBYPASS: TRUE
And the results were bit-for-bit in comparing the blom history files
       *.hd.0001-01-01.nc
        *.hbgcd.0001-01-01.nc
So this gives me more confidence that the tests that @JorgSchwinger was proposing will pass.

mvertens:
Once I reintroduce ci.yml - the test is now failing. In the past I did notice that periodically it would fail and then pass.

trc/restart_trcrd.F90
mvertens: Could we defer deciding this to #281? I'd like to know from @matsbn with always compiling the hamocc code since there are no longer any #ifdefs.

hamocc/mo_param_bgc.F90
mvertens: Yes - you are right. Sorry for not seeing that.

jmaerz: No worries - I am happy to be able to help :-)

mvertens: I've pushed the fix back.

jmaerz (approved):

trc/restart_trcrd.F90
```diff
@@ -86,9 +90,9 @@ subroutine restart_trcrd(rstfnm_ocn)
 !
   call restart_ocntrcrd(rstfnm_ocntrc)
 
-#ifdef HAMOCC
-  call hamocc_init(1,rstfnm_hamocc)
-#endif
+  if (use_HAMOCC) then
+     call hamocc_init(1,rstfnm_hamocc)
+  end if
```
TomasTorsvik, line 95:
```suggestion
#ifdef HAMOCC
     call hamocc_init(1,rstfnm_hamocc)
#endif
```

The CI intel compiler build fails on the call to hamocc_init. When meson builds without the 'ecosys' option it does not compile code in the `hamocc` directory, so then `hamocc_init` is not available. 

It may be safer for now to keep the calls to hamocc funtions/subroutines in #ifdef, when called from outside hamocc.

TomasTorsvik: I tried to update the ci.yml file to include tests with 'ecosys' enabled, but the tests are still failing. Not sure what's going on, but the meson build system doesn't seem to recognize `ecosys=true`.

mvertens: @TomasTorsvik - thanks for looking into this. I agree that it makes sense at this point to keep the #ifdef HAMOCC on.
Are we okay with merging this PR - or do we want to resolve the meson build system problem first?

TomasTorsvik: We can merge this now. I can add the meson build afterwards, I didn't enable it properly with ecosys==true.

JorgSchwinger:
@mvertens, @jmaerz, @TomasTorsvik I just tested this PR, and the results for the C-isotopes are NOT bfb!

I'm investigating this further, but this should be clarified before we move on

mvertens:
@JorgSchwinger - thanks for doing this test. Can you please point me to your rundir and caseroot.
I thought I had turned on isotopes last night and it was bfb - but clearly I'm not doing something right.

TomasTorsvik:
@mvertens , @JorgSchwinger , @jmaerz 
I will open a new issue on this, with reference to PR #280. We will probably need to do a fix in a new PR anyway, so it's not convenient to keep the discussion in an already merged PR.

PR #286, opened 2023-10-17
Fixes to run C-isotopes

JorgSchwinger:
This fixes a model crash when running C-isotopes with the sediment activated. It also removes the error message when C-isotopes are build with sediment activated.


JorgSchwinger:
Note: this doesn't fix the issue of being not bfb with the master before #280

jmaerz (approved):
This was the one-line fix from Alok, right? - the rest is more the optics, it seems.

JorgSchwinger:
Sorry, I should have left out the code-cosmetics, but there is a real fix in line 684: the definition of one variable was duplicated

mvertens (approved):

PR #287, opened 2023-10-18
Fix bug in restart/initialization for C-isotope sediment fields

### LINKED ISSUE #285: NorESM is no longer bfb identical after merging PR-280
Merging PR #280 seems to have caused a change that breaks bfb reproducible results with earlier model versions. This was not an expected change from merging the PR.

- Test case: NOINYOC_T62_tn21
C-isotopes with and without sediments give very different results. 13C with sediments look wrong.

JorgSchwinger:
This PR closes #285 - with this fix I get bfb identical results with C-isotopes and sediment switched on.

### DIFF
--- a/hamocc/aufr_bgc.F90
+++ b/hamocc/aufr_bgc.F90
@@ -103,26 +103,21 @@
 !**************************************************************************
 
       use netcdf,         only: nf90_global,nf90_noerr,nf90_nowrite,nf90_close,nf90_open,nf90_get_att,nf90_inq_varid
-      use mo_carbch,      only: co2star,co3,hi,satoxy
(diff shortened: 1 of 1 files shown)


jmaerz (approved):
Good that you found the source of the error. I guess, we should do some code cosmetics after all merges .... I'll allocate time for the merging latest next week (first the N-cycle into `beyond-CMIP6`, then `master` into `beyond-CMIP6`). After that, this can be merged into `master` and there, we can have a code cosmetics PR+merge :-)

mvertens (approved):
@JorgSchwinger - thanks for tracking this down!

PR #194, opened 2022-09-15
BUG fixes: sediment alkalinity and sediment C14

jmaerz:
Dear @TomasTorsvik and @JorgSchwinger , I herewith open the pull request linked to issue #193. In addition to the described alkalinity issue, it seemed to me that there was an **additional bugfix** for C14 needed - please re-check carefully. This branch hasn't been checked on betzy yet, but compiled successfully locally. I left the rest of the isotopes untouched (didn't introduce `sulf13 ` and `sulf14`, respectively, since it wasn't needed from a bug fixing perspective, while it breakes the consistency between `anaerob` and e.g. `anaerob13`). If you wish to introduce `sul13,14` for more consistency, let me know.  closes #193

PR #290, opened 2023-10-19
Fix bug in powach.F90 C-isotope code

JorgSchwinger:
I put this fix in for the CMIP6 bw-compatible release of BLOM, since this will only affect C-isotopes.


jmaerz (approved):
This commit introduces part of the bugfix (specific to the C-isotopes) carried out in #194 in the `beyond-CMIP6` branch. Since the carbon isotopes don't affect backwards compatibility with the CMIP6 version and for the sake of having a bugfixed isotope code in the yet CMIP6-compatible release version, this gets introduced in `mater` with this bugfix. Approved.

PR #291, opened 2023-10-20
Add capability to run OAE experiments with an OmegaA threshold

JorgSchwinger:


TomasTorsvik (approved):
Looks fine. Initially I got a bit confused about the point of including this threshold option.

JorgSchwinger:
Thank you Tomas, I just wanted to have this in the last CMIP6-bw-compatible release since we are using this in project where we look at CMIP6 scenarios with additional ocean alkalinization.

PR #297, opened 2023-11-01
Move main sediment parameters to mo_param_bgc

### LINKED ISSUE #296: Move sediment parameters into mo_param_bgc
Hi @TomasTorsvik and @JorgSchwinger , I am now working on moving the sediment parameters also into `mo_param_bgc` - once this is ready, Jörg can pretty the code which can then be merged into the `beyond-CMIP6` branch.

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , I moved most sediment parameters to `mo_param_bgc` (with the exceptions of `calfa`, `oplfa`, `orgfa`, `clafa` that are calculated from parameters set in `mo_param_bgc` - if you wish, I can also transfer them). In 1D, it looks bit-identical - but I will also set up runs on betzy to test it there. 
Closes #296

### DIFF
--- a/hamocc/dipowa.F90
+++ b/hamocc/dipowa.F90
@@ -56,8 +56,7 @@ subroutine dipowa(kpie,kpje,kpke,omask,lspin)
 !**********************************************************************
 
   use mo_carbch,     only: ocetra, sedfluxo
-  use mo_sedmnt,     only: powtra,porwat,porwah,seddw,seddzi,zcoefsu,zcoeflo
-  use mo_param_bgc,  only: sedict
+  use mo_sedmnt,     only: powtra,porwat,porwah,seddw,zcoefsu,zcoeflo
   use mo_param1_bgc, only: ks,npowtra,map_por2octra
   use mo_vgrid,      only: kbo,bolay
   ! cisonew
--- a/hamocc/dipowa.F90
+++ b/hamocc/dipowa.F90
@@ -55,8 +55,9 @@ subroutine dipowa(kpie,kpje,kpke,omask,lspin)
 !
 !**********************************************************************
 
-  use mo_carbch,     only: ocetra, sedfluxo 
-  use mo_sedmnt,     only: powtra,porwat,porwah,sedict,seddw,seddzi,zcoefsu,zcoeflo 
+  use mo_carbch,     only: ocetra, sedfluxo
+  use mo_sedmnt,     only: powtra,porwat,porwah,seddw,seddzi,zcoefsu,zcoeflo
+  use mo_param_bgc,  only: sedict
   use mo_param1_bgc, only: ks,npowtra,map_por2octra
   use mo_vgrid,      only: kbo,bolay
   ! cisonew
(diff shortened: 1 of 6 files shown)


jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , I checked the restructured branch against master on betzy (as done for the water column restructuring) and get bit-identical results.

hamocc/dipowa.F90
```diff
@@ -55,8 +55,9 @@ subroutine dipowa(kpie,kpje,kpke,omask,lspin)
 !
 !**********************************************************************
 
-  use mo_carbch,     only: ocetra, sedfluxo 
-  use mo_sedmnt,     only: powtra,porwat,porwah,sedict,seddw,seddzi,zcoefsu,zcoeflo 
+  use mo_carbch,     only: ocetra, sedfluxo
+  use mo_sedmnt,     only: powtra,porwat,porwah,seddw,seddzi,zcoefsu,zcoeflo
+  use mo_param_bgc,  only: sedict
```
TomasTorsvik: I don't see that `seddzi` and `sedict` are used in `dipowa.F90`. Is this correct? If so, they don't need to be imported.

TomasTorsvik (approved):
Looks good to me :+1:

mvertens:
@TomasTorsvik @jmaerz - would it be okay to start the master tagging here?
What should we use to start with? blom_v1.0.0?

hamocc/dipowa.F90
jmaerz: Correct. I changed this.

jmaerz:
Dear @mvertens , @TomasTorsvik and @JorgSchwinger wrt to the tagging: We are almost there to have a still CMIP6 compatible, but clean(er) code version on the iHAMOCC side. After this PR, Joerg wants to take care a bit about the indentations, `use` statements etc in iHAMOCC. Maybe with this cleaned version, a `blom_v1.0.0` tag state would be nice? - but I don't have strong opinions on it. So from my side, either way is fine.

mvertens:
@jmaerz - thanks for the update. I agree that starting with a cleaned up version sounds fine. Can we have a chat about indentations throughout hamocc when you start? I think we can automate a lot of this with emacs and have it standardized throughout the code if you all think that would be useful.

jmaerz:
:-D - this is a question to @JorgSchwinger - he has particular opinions on it (and I don't work with emacs). But I am happy to learn!

JorgSchwinger (approved):
Thank you looks good!

JorgSchwinger:
@mvertens @TomasTorsvik @jmaerz  

Regarding the tags: What would be the relation between these automated tags and our releases? We have currently tagged releases v1.0.0 to v1.3.0 and we are working towards v1.4.0 (the last CMIP6 backwards compatible release). So I would find it confusing to have the automated tag v1.0.0 associated with our release v1.4.0 - is there a reasonable way to synchronize this?

Regarding automated indentations, I am a bit skeptical about it, because I really like to use vertical alignment to improve (in my opinion) code readability. I find automatically indented code often difficult to read because things that belong together logically are displaced. But I might be the only one with this kind of reservation... (I do not use emacs a lot though)

jmaerz:
Then it would be at least logical to start with `blom_v1.4.0`, once you (@JorgSchwinger) have cleaned the code. But maybe we should continue this discussion in #295.

jmaerz:
I copied the discussion points on tagging to #295.

Issue #289, opened 2023-10-18
Merging strategy for new release versions (CMIP6-compatible and INES interim release)


jmaerz:
Hi @JorgSchwinger  and @TomasTorsvik ,

PR related in preparation to https://github.com/NorESMhub/NorESM/issues/466:

Required for **CMIP6-compatible** final version:
- #281 
- #291 
- #290
- potentially #233 
- #294
- #297
- #299 
- #300
- #301
- #303
- #305
- #307
- #308
- #309 (currently expected as the final and last CMIP6-compatible version from the HAMOCC side)
- #310 tagged release of CMIP6 compatible BLOM/HAMOCC

(Upcoming) PR for getting iHAMOCC ready for **NorESM2.1** version (PRs are named, once they are opened):
- #311 Merge `master` into `beyond-CMIP6`  & code cosmetics on `beyond-CMIP6`
- #314 Some minor declaration, etc. issues
- #315 Merge `beyond-CMIP6` (incl. bug-fixes) back into `master` - making `master` ready for milestone NorESM2.1
**From [cca8d11](https://github.com/NorESMhub/BLOM/commit/cca8d118e638d38e14eb860c60cfb49579b87728) onward, `master` is not backward compatible any longer for iHAMOCC** and ready for NorESM2.1.

(Upcoming) PR for getting iHAMOCC ready for **NorESM2.x** version (PRs are named, once they are opened):
- #269 
- potentially add preformed silicate tracer

jmaerz:
**Notes for merging `master` into `beyond-CMIP6`:**
Check fixes formerly carried out: 
- DONE burial
- DONE OMP in sinking
- DONE alkalinity, etc. in sediment

Further changes to apply, once `master` is merged into `beyond-CMIP6`:
- DONE: `wpoc`, `wopal`, etc. as `protected` (via #307) 
- DONE adapt OMP settings in sinking
- #314 `bifr13`, `bifr14`

Changing opal dissolution rate in sediment will be carried out for a later tagged NorESM version

PR #299, opened 2023-11-03
Some cosmetic changes

JorgSchwinger:
I went through the code and prettied a few things up a bit (as I believe). If there is something controversial I'm happy to revert it.

### DIFF
--- a/hamocc/aufr_bgc.F90
+++ b/hamocc/aufr_bgc.F90
@@ -163,18 +163,19 @@
       !
       testio=0
       IF(mnproc==1 .AND. IOTYPE==0) THEN
+      
         ncstat = NF90_OPEN(rstfnm,NF90_NOWRITE, ncid)
         IF ( ncstat .NE. NF90_NOERR ) THEN
-             CALL xchalt('(AUFR: Problem with netCDF1)')
-                    stop '(AUFR: Problem with netCDF1)'
+           CALL xchalt('(AUFR: Problem with netCDF1)')
+                  stop '(AUFR: Problem with netCDF1)'
         ENDIF
         !
         ! Read restart data : date
         !
         ncstat = NF90_GET_ATT(ncid, NF90_GLOBAL,'date', idate)
         IF ( ncstat .NE. NF90_NOERR ) THEN
-          CALL xchalt('(AUFR: Problem reading date of restart file)')
-          stop '(AUFR: Problem reading date of restart file)'
+           CALL xchalt('(AUFR: Problem reading date of restart file)')
+                  stop '(AUFR: Problem reading date of restart file)'
         ENDIF
         restyear  = idate(1)
         restmonth = idate(2)
@@ -202,8 +203,8 @@
 
         ncstat = NFMPI_OPEN(mpicomm,rstfnm,NF_NOWRITE,INFO, ncid)
         IF ( ncstat .NE. NF_NOERR ) THEN
-             CALL xchalt('(AUFR: Problem with netCDF1)')
-                    stop '(AUFR: Problem with netCDF1)'
+           CALL xchalt('(AUFR: Problem with netCDF1)')
+                  stop '(AUFR: Problem with netCDF1)'
         ENDIF
 
         !
@@ -211,8 +212,8 @@
         !
         ncstat = NFMPI_GET_ATT_INT(ncid, NF_GLOBAL,'date', idate)
         IF ( ncstat .NE. NF_NOERR ) THEN
-          CALL xchalt('(AUFR: Problem reading date of restart file)')
-          stop '(AUFR: Problem reading date of restart file)'
+           CALL xchalt('(AUFR: Problem reading date of restart file)')
+                  stop '(AUFR: Problem reading date of restart file)'
         ENDIF
         restyear  = idate(1)
         restmonth = idate(2)
@@ -230,34 +231,26 @@
         WRITE(io_stdo_bgc,*) ' '
         ENDIF
 #endif
-      if(testio .eq. 0) then
-      CALL xchalt('(AUFR: Problem with namelist iotype)')
-                    stop '(AUFR: Problem with namelist iotype)'
-      endif
-      ENDIF
+        if(testio .eq. 0) then
+           CALL xchalt('(AUFR: Problem with namelist iotype)')
+                  stop '(AUFR: Problem with namelist iotype)'
+        endif
+
+      ENDIF ! mnproc==1 .AND. IOTYPE==0
 
       !
       ! Compare with date read from ocean restart file
       !
       IF (mnproc.eq.1) THEN
 
-         IF ( kplyear  .NE. restyear  ) THEN
-            WRITE(io_stdo_bgc,*)                                     &
-     &      'WARNING: restart years in oce/bgc are not the same : '  &
-     &      ,kplyear,'/',restyear,' !!!'
-         ENDIF
+         IF ( kplyear .NE. restyear  ) WRITE(io_stdo_bgc,*)                                      &
+            'WARNING: restart years in oce/bgc are not the same : ', kplyear,'/',restyear,' !!!'
 
-         IF ( kplmon .NE. restmonth ) THEN
-            WRITE(io_stdo_bgc,*)                                     &
-     &      'WARNING: restart months in oce/bgc are not the same : ' &
-     &      ,kplmon,'/',restmonth,' !!!'
-         ENDIF
+         IF ( kplmon  .NE. restmonth ) WRITE(io_stdo_bgc,*)                                      &
+            'WARNING: restart months in oce/bgc are not the same : ',kplmon,'/',restmonth,' !!!'
 
-         IF ( kplday   .NE. restday   ) THEN
-            WRITE(io_stdo_bgc,*)                                     &
-     &      'WARNING: restart days in oce/bgc are not the same : '   &
-     &      ,kplday,'/',restday,' !!!'
-         ENDIF
+         IF ( kplday  .NE. restday   ) WRITE(io_stdo_bgc,*)                                      &
+            'WARNING: restart days in oce/bgc are not the same : ',  kplday,'/',restday,' !!!'
 
       ENDIF
 
@@ -275,11 +268,11 @@
 #endif
          ENDIF
          IF(mnproc==1 .and. .not. lread_cfc) THEN
-         WRITE(io_stdo_bgc,*) ' '
-         WRITE(io_stdo_bgc,*) 'AUFR_BGC info: CFC tracers not in restart file, '
-         WRITE(io_stdo_bgc,*) ' CFCs initialised to zero.'
+            WRITE(io_stdo_bgc,*) ' '
+            WRITE(io_stdo_bgc,*) 'AUFR_BGC info: CFC tracers not in restart file, '
+            WRITE(io_stdo_bgc,*) ' CFCs initialised to zero.'
          ENDIF
-      end if
+      endif
 
       ! Find out whether to restart natural tracers
       if (use_natDIC) then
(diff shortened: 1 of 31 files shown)


hamocc/ncout_hamocc.F90
```diff
@@ -1113,22 +1114,22 @@ subroutine hamoccvardef(iogrp,timeunits,calendar,cmpflg)
...
   if (use_cisonew) then
      call ncdefvar3d(SRF_ATMC13(iogrp),cmpflg,'p',                                 &
           &   'atmc13','Atmospheric 13CO2',' ','ppm',0)
      call ncdefvar3d(SRF_ATMC14(iogrp),cmpflg,'p',                                 &
           &   'atmc14','Atmospheric 14CO2',' ','ppm',0)
-  end if
+  endif
 
   ! --- define 3d layer fields
   call ncdefvar3d(LYR_DP(iogrp),cmpflg,'p',                                     &
```
jmaerz: missed this `&` sign

hamocc/ncout_hamocc.F90
```diff
@@ -1240,11 +1241,11 @@ subroutine hamoccvardef(iogrp,timeunits,calendar,cmpflg)
...
-  end if
+  endif
   if (use_BROMO) then
      call ncdefvar3d(LYR_BROMO(iogrp),cmpflg,'p',                                  &
           &   'bromo','Bromoform',' ','mol CHBr3 m-3',1)
-  end if
+  endif
 
   ! --- define 3d level fields
   call ncdefvar3d(LVL_DIC(iogrp),cmpflg,'p',                                    &
```
jmaerz: and these ones

hamocc/mo_sedmnt.F90
```diff
@@ -151,17 +151,18 @@ SUBROUTINE ini_sedmnt(kpie,kpje,kpke,omask,sed_por)
...
-      !========================================================================
+ !========================================================================
  SUBROUTINE ini_sedmnt_por(kpie,kpje,kpke,omask,sed_por)
-      !
-      ! Initialization of:
-      !   - 3D porosity field (cell center and cell boundaries)
-      !   - solid volume fraction at cell center
-      !   - vertical molecular diffusion coefficients scaled with porosity
-      !
+!
```
jmaerz, line 160: Admittedly, I am not really happy with loosing the indentation here for the comments, since they belong to what comes after, which is also indented. But that might be just my personal matter and if you prefer it like this, I'll adapt.

mvertens: @jmaerz - is your preference to have all comments not indented? Is that the general consensus for hamocc?

jmaerz: No, just the opposite. I would want to have the comments aligned to the code - I was just about to suggest this as a general comment - but it is something to generally agree on.

mvertens: We totally agree. Best free-format practices recommend - 'Indent blocks by 2 characters. Where possible, comments should be indented with the code within a block'.

jmaerz:
Hi @JorgSchwinger , ok, seems as if now, I am becoming a bit picky, sorry for that, but: could we think about aligning comments in general with the respective code? - e.g. in `carchm` , I find it very confusing, where the comment belongs to. etc. - with indenting comments, it becomes clearer, to which code they belong. Further, they visually don't break indentations of code (BTW: this is similarly true and could be done for OMP, while I there understand the reasoning for non-indentation a bit better).

hamocc/mo_sedmnt.F90
JorgSchwinger: I also generally agree - in this specific case, however, I was interpreting the comment as a header of the subroutine and therefore aligned it with the SUBROUTINE statement above

jmaerz:
Another thing would be the small or capital writing of fortran command, but maybe this goes beyond what we should care about right now. For those, I would prefer to keep `MODULE` `CONTAINS` `END MODULE` in capital, while the rest would be fine also in small. Anyway, this probably is too much for now.

hamocc/mo_intfcblom.F90
```diff
@@ -482,17 +483,17 @@ subroutine hamocc2blom(m,n,mm,nn)
...
+        do i=max(1,ifp(j,l)),min(ii,ilp(j,l))              ! time smoothing (analog to tmsmt2.F)
+           sedlay2(i,j,km,:) = wts1*sedlay2(i,j,km,:)   &  ! mid timelevel
+                             + wts2*sedlay2(i,j,kn,:)   &  ! old timelevel
+                             + wts2*sedlay(i,j,k,:)        ! new timelevel
+           powtra2(i,j,km,:) = wts1*powtra2(i,j,km,:)   &
+                             + wts2*powtra2(i,j,kn,:)   &
+                             + wts2*powtra(i,j,k,:)
+           burial2(i,j,m,:)  = wts1*burial2(i,j,m,:)    &
+                             + wts2*burial2(i,j,n,:)    &
+                             + wts2*burial(i,j,:)
```
JorgSchwinger, line 496: @mvertens
this is a good example of what I mean with the vertical alignment. I guess these changes would be reverted by an auto-indent?

hamocc/mo_sedmnt.F90
jmaerz: Right, if it would have been written above the subroutine (which I wouldn't want) I would agree. Admittedly, here and also in other places, I would prefer indentation (there are a few more places, where this holds). But again, I am also fine, if you prefer it differently.

jmaerz:
Looks like we opened a can of worms :-)

jmaerz:
Mmh, since this PR provokes some comments, I would suggest to collect them afterward in a small document (or updating older documents on best practices for iHAMOCC coding style).

hamocc/aufr_bgc.F90
```diff
@@ -163,18 +163,19 @@ SUBROUTINE AUFR_BGC(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,               &
       !
       testio=0
       IF(mnproc==1 .AND. IOTYPE==0) THEN
+      
         ncstat = NF90_OPEN(rstfnm,NF90_NOWRITE, ncid)
         IF ( ncstat .NE. NF90_NOERR ) THEN
-             CALL xchalt('(AUFR: Problem with netCDF1)')
-                    stop '(AUFR: Problem with netCDF1)'
+           CALL xchalt('(AUFR: Problem with netCDF1)')
```
jmaerz, line 169: Admittedly, the indentation of the stop is unclear to me (if it is about aligning the message, you could just shift the message). Go as you like.

JorgSchwinger: Agree, will fix this.

hamocc/hamocc4bcm.F90
```diff
@@ -361,27 +359,26 @@ SUBROUTINE HAMOCC4BCM(kpie,kpje,kpke,kbnd,kplyear,kplmon,kplday,kldtday,&
...
-         end if
+         endif
 
-         !     sediment is shifted once a day (on both time levels!)
+         ! Sediment is shifted once a day (on both time levels!)
          IF(KLDTDAY .EQ. 1 .OR. KLDTDAY .EQ. 2) THEN
             IF (mnproc.eq.1) THEN
-               WRITE(io_stdo_bgc,*)                                           &
-                    &   'Sediment shifting ...'
+               WRITE(io_stdo_bgc,*)' '
```
jmaerz, line 367: why adding a second line?

JorgSchwinger:
I agree, this seems to be a can of worms...

It looks like we will have to discuss things a bit first. I would suggest that we sit together and update the HAMOCC coding guidelines that we already made some time ago (should be also put into the repository).

For now, I would suggest to leave it here (I have one more pending request though) such that we can proceed making the release v1.4.0 and start working towards v1.5.0.

hamocc/ocprod.F90
```diff
@@ -100,7 +101,6 @@ subroutine ocprod(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,omask,ptho,pi_ph)
                             use_BROMO,use_AGG,use_PBGC_OCNP_TIMESTEP,use_FB_BGC_OCE,use_AGG,use_cisonew,use_natDIC,                &
                             use_WLIN,use_sedbypass
   use mo_vgrid,       only: dp_min,dp_min_sink,k0100,k0500,k1000,k2000,k4000,kwrbioz,ptiestu
-  use mod_xc,         only: mnproc
   use mo_biomod,      only: eps3d,bifr13_perm,growth_co2
```
jmaerz: `mo_biomod` could go up (to the first `mo_biomod`)

hamocc/ocprod.F90
```diff
@@ -1078,50 +1076,50 @@ subroutine ocprod(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,omask,ptho,pi_ph)
...
+                                        + ocetra(i,j,kdonor,idet13)*wpocd)/        &
+                                         (pddpo(i,j,k)+wpoc)
+                 ocetra(i,j,k,idet14)  = (ocetra(i,j,k     ,idet14)*pddpo(i,j,k)   &
+                                        + ocetra(i,j,kdonor,idet14)*wpocd)/        &
+                                         (pddpo(i,j,k)+wpoc)
+                 ocetra(i,j,k,icalc13) = (ocetra(i,j,k     ,icalc13)*pddpo(i,j,k)  &
+                                        + ocetra(i,j,kdonor,icalc13)*wcald)/       &
+                                         (pddpo(i,j,k)+wcal)
+                 ocetra(i,j,k,icalc14) = (ocetra(i,j,k     ,icalc14)*pddpo(i,j,k)  &
+                      	                 + ocetra(i,j,kdonor,icalc14)*wcald)/       &
```
jmaerz, line 1095: & sign indent

hamocc/aufw_bgc.F90
```diff
@@ -178,7 +177,7 @@ SUBROUTINE AUFW_BGC(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,              &
          ncstat = NF90_CREATE(rstfnm,NF90_64BIT_OFFSET,ncid)
          IF ( ncstat .NE. NF90_NOERR ) THEN
             call xchalt('(AUFW: Problem with netCDF1)')
-            stop '(AUFW: Problem with netCDF1)'
+                   stop '(AUFW: Problem with netCDF1)'
```
TomasTorsvik: Same indentation as for `aufr_bgc`. Just a reminder to fix it in both files.

hamocc/mo_intfcblom.F90
mvertens: @JorgSchwinger - yes it is. Thanks for the clarification. I agree that these changes would be reverted. 
Not sure if we could automate that. Maybe a meeting to discuss all of these issues would be good.

jmaerz (approved):
Hi @JorgSchwinger , many thanks for the effort to bring the code to a better shape! - I am fine to approve (and do so), but I would suggest that we note down some (further) suggestions that we have, when browsing your changes. We had this already on the bucket list since a while and now the PR gives rise to further thoughts on it. From my side, I came up with a few general thoughts when browsing the changes:

- How much indent per code block (I would also opt for 2-3)
- Some routines/modules still have the +7 F77 indent, which could be changed to a smaller indent
- I agree with your 'indent by meaning approach' when aligning following up lines, which likely would mean no automated indention via emacs.
- Wrt spatial loops: any way is fine, I would say.
- I would be in favor of aligning comments to code blocks (to not break the visual code flow) - similarly, one could think about OMP directives
- small versus capital writing of Fortran commands (as written before, I would prefer capital for module, contains, implicit none, potentially subroutines, the rest I would be fine either way - like integer, float, if else endif constructs)
- Indention of comments on subroutines?

Ok, I guess, this is the collection of mine for now, but others might have more - I would compile this list also in the discussion #298, where we can follow up on it.

hamocc/mo_intfcblom.F90
TomasTorsvik: Yes, at least in my experience, auto-indent will normally indent by a fixed number of spaces for line continuation.

mvertens:
@jmaerz - thanks so much for compiling this list. That's super helpful and I agree should be moved to a discussion. I think we need to consider if we want a tool to go through the code and format it as a first pass and then fix things we don't like. It does not have to be emacs. Another point to consider is that most auto-formatters will not handle f77 code - but deal with free format code. At any rate this is a good discussion to have and come to a consensus on.

TomasTorsvik (approved):
@JorgSchwinger - Thanks! This look fine to me :+1:

TomasTorsvik:
@JorgSchwinger @jmaerz 
Once this is merged, are we ready for a `release-1.4` (i.e. still CMIP6 compatible) from iHAMOCC side?

jmaerz:
@TomasTorsvik , I think, #300 should also come in, right? - I'll update #289

PR #300, opened 2023-11-03
Remove dmspar variable

JorgSchwinger:
The dmspar variable was used to define the parameters of the DMS scheme, but it was copied over to variables dmsp1-dmsp6. This PR simplifies this by directly defining dsmp1-dsmp6 in mo_param_bgc.

### DIFF
--- a/hamocc/mo_param_bgc.F90
+++ b/hamocc/mo_param_bgc.F90
@@ -93,9 +93,6 @@ module mo_param_bgc
   ! Decay parameter for C14, HalfLive = 5700 years
   real, parameter :: c14_t_half = 5700.*365.      ! Half life of 14C [days]
 
-  ! Scaling factor for pH dependency (used if with_dmsph=.true.)
-  real, parameter :: dms_gamma = 0.87
-
   !'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
   ! Atmosphere:
   !'''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''''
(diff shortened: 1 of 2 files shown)


hamocc/mo_param_bgc.F90
```diff
@@ -162,12 +173,13 @@ module mo_param_bgc
   real, protected :: tf0        = -2.7819
   real, protected :: tff        =  0.2395
 
-  ! Initial fractionation during photosynthesis
-  real :: bifr13 = 0.98
-  real :: bifr14
-
-  ! Decay parameter for sco214, HalfLive = 5730 years
-  real, parameter :: c14_t_half = 5700.*365.      ! Half life of 14C [days]
+  ! Set constants for dms scheme following Kloster et al. (2006), Table 1
```
jmaerz: Stupid question: any reason to keep the reverse order of initialization?

hamocc/mo_param_bgc.F90
```diff
@@ -148,6 +155,10 @@ module mo_param_bgc
   real, protected :: bkphy      = 4.e-8           ! kmol/m3 - i.e. 0.04 mmol P/m3 half saturation constant
   real, protected :: dyphy      = 0.004           ! 1/d -mortality rate of phytoplankton
 
+  ! Initial fractionation during photosynthesis
```
jmaerz, line 155: In the comment for this initialization section, I would suggest to add DMS information (i.e. `Phytoplankton parameters (incl. cyanobacteria and -partially microbial- DMS production`) or something similar.

hamocc/mo_param_bgc.F90
```diff
@@ -89,6 +90,12 @@ module mo_param_bgc
   real, parameter :: rdn2o1 = 2*ro2ut - 2.5*rnit  ! moles N2O used for remineralisation of 1 mole P
   real, parameter :: rdn2o2 = 2*ro2ut - 2*rnit    ! moles N2 released  for remineralisation of 1 mole P
 
+  ! Decay parameter for C14, HalfLive = 5700 years
+  real, parameter :: c14_t_half = 5700.*365.      ! Half life of 14C [days]
+
+  ! Scaling factor for pH dependency (used if with_dmsph=.true.)
```
jmaerz: I would still put the DMS part into the DMS section below.... - just to have that all together.

jmaerz (approved):
Hi @JorgSchwinger , thanks for making this part cleaner. I approve while I suggest to follow the comments.

TomasTorsvik (approved):
Thanks, no complaints from my side :+1:

hamocc/mo_param_bgc.F90
JorgSchwinger: No, just laziness - will fix this.

hamocc/mo_param_bgc.F90
JorgSchwinger: I have made an extra header for DMS and BrO schemes

jmaerz:
Hi @JorgSchwinger , I just realized that you could outsource the multiplication with `dtb` from `ocprod` to `mo_param_bgc` as well... - but up to you.

mvertens:
@JorgSchwinger @jmaerz @TomasTorsvik - since we are migrating this code to master - it would be good moving forwards to run the regression test suite and determine if answers change by comparing to baselines and generating baselines. Are only stand-alone tests being run for these updates - or are tests using blom within NorESM being used? And it might be good to add new compsets that might trigger new added features. Should I try to arrange a meeting next week to talk about this?

JorgSchwinger:
> Hi @JorgSchwinger , I just realized that you could outsource the multiplication with `dtb` from `ocprod` to `mo_param_bgc` as well... - but up to you.

Yes, this is not nice, but note that this was no change in comparison to before. Not sure whether to change this now.

JorgSchwinger:
I have tested this in stand-alone getting bit-for-bit identical results. We will need new baselines when we have finished v1.4.0 and move towards v1.5.0, I guess. I'm happy to meet next week.

jmaerz:
As written, up to you - I haven't really had a look into DMS before, but I realized that these are rates, which one could/would want to tune with changing physics, etc. - and just to be in line with what we have thus far (converting rates per s or d to per time step). From my side, just go ahead and merge. I just had the impression that we aimed at a cleaner code in general for this release.

________________________________________

JorgSchwinger:
I might try to fix the dtb issue in a separate PR (it's the same issue for the BrO scheme). Was a bit worried about bfb.

jmaerz:
@JorgSchwinger , all fine - understandable ( I was also crossing fingers when moving all the parameters around). Wrt meeting next week, @mvertens , I would be also fine with it.

PR #301, opened 2023-11-04
Remove variable bolven from sediment code, consistent treatment of dms parameters

JorgSchwinger:


jmaerz (approved):
Looks good to me. Thanks! Just for documentation: `bolven` was originally applied for sediment acceleration (but hardly used) and became obsolete with the new sediment acceleration scheme.

TomasTorsvik (approved):
Looks fine to me :+1:

PR #303, opened 2023-11-06
Minor fixes for configuration and namelist creation

JorgSchwinger:


mvertens (approved):
Thanks for this.

TomasTorsvik (approved):
Thanks, looks good to me :+1: 
Some CI-tests timed out. Will see if I can re-run, but shouldn't be essential.

jmaerz:
Hi @JorgSchwinger , it shouldn't conflict with #305, but maybe it would be good to merge it nevertheless (I took myself out as a reviewer).

JorgSchwinger:
I have put this on hold since it will interfere with the regression testing. There will be a fix needed since the C-isotope code cannot be switched on via user-namelist, but needs to go via env_run.xml.

@mvertens if you can tell me how to fix this, I could include it into this PR. Unless you think it is easier if you fix this after merging this PR?

mvertens:
@JorgSchwinger - 

 in blom/cime_config/testdefs/testmods_dirs/blom/hamocc1/ 
    add a file shell_commands and in that file add the line
./xmlchange HAMOCC_CISO=TRUE
    and in user_nl_blom remove the line
 use_cisonew = .true.

You need to do this for every new xml variable that was introduced for the user_nl_blom variables that are in the user_nl_blom file

You should also do this for in blom/cime_config/testdefs/testmods_dirs/blom/hamocc2/

That should be it. Does that make sense?

JorgSchwinger:
Ok thank you for the instructions @mvertens, done. 

I guess then this PR can be merged?

mvertens:
That would be fine with me.

PR #304, opened 2023-11-06
Test str replacement - convert capitalized to small (and vice versa, if desired)

jmaerz:
Hi @mvertens , @JorgSchwinger and @TomasTorsvik , here my little contribution to the coding style issue. I tried to enforce the style as I would prefer it to have :-) - but it is just a proposition. I constructed a small shell script for sed-based mapping. It;s not perfect, I know, but as a start....

There might be still some things missing, which I can add later.

### DIFF
--- a/hamocc/accfields.F90
+++ b/hamocc/accfields.F90
@@ -90,7 +90,7 @@
       use mo_sedmnt,      only: powtra,sedlay,burial
       use mo_vgrid,       only: dp_min
 
-      IMPLICIT NONE
+      implicit none
 
       integer , intent(in) :: kpie,kpje,kpke
       real    , intent(in) :: pdlxp(kpie,kpje)
--- a/hamocc/accfields.F90
+++ b/hamocc/accfields.F90
@@ -35,13 +35,13 @@
 !**** Parameter list:
 !     ---------------
 !
-!  *INTEGER* *kpie*    - 1st dimension of model grid.
-!  *INTEGER* *kpje*    - 2nd dimension of model grid.
-!  *INTEGER* *kpke*    - 3rd (vertical) dimension of model grid.
-!  *REAL*    *pdlxp*   - size of scalar grid cell (1st dimension) [m].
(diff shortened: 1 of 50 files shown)


hamocc/aufr_bgc.F90
```diff
@@ -162,35 +162,35 @@ SUBROUTINE AUFR_BGC(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,               &
       ! Open netCDF data file
       !
       testio=0
-      IF(mnproc==1 .AND. IOTYPE==0) THEN
+      if (mnproc==1 .AND. IOTYPE==0) then
       
         ncstat = NF90_OPEN(rstfnm,NF90_NOWRITE, ncid)
-        IF ( ncstat .NE. NF90_NOERR ) THEN
-           CALL xchalt('(AUFR: Problem with netCDF1)')
+        if ( ncstat .NE. NF90_NOERR ) then
```
JorgSchwinger, line 168: maybe we could also replace .NE. with  /=, .EQ. with ==, and so forth?

mvertens: @jmaerz - thanks for this. I'm wondering if moving forwards we can think of doing a lot of this with emacs cs without needing new scripts since emacs can recursively go through a set of files and do regular expression conversions. Just another option to think about.

hamocc/accfields.F90
```diff
@@ -90,22 +90,22 @@ SUBROUTINE ACCFIELDS(kpie,kpje,kpke,pdlxp,pdlyp,pddpo,omask)
       use mo_sedmnt,      only: powtra,sedlay,burial
       use mo_vgrid,       only: dp_min
 
-      implicit none
+      IMPLICIT NONE
```
JorgSchwinger: Fine for me to have MODULE, SUBROUTINE, etc. still in capital letters, since they indicate program units, but I don't understand the logic behind having IMPLICIT NONE in capital letters?

PR #308, opened 2023-11-11
Feature code cosmetics2

JorgSchwinger:
This is my last contribution before we make a last CMIP6 backwards-compatible release. It is mostly cosmetics. I also combined three modules with netcdf routines into one (named mo_netcdf_bgcrw).


hamocc/mo_control_bgc.F90
```diff
@@ -18,15 +18,15 @@
 
 module mo_control_bgc
```
jmaerz, line 20: Hi Jörg,
first of all many thanks for your effort!
Would you mind adding comments, in which namelist the logicals can be changed? - I think it would be good for understanding their role and the function for HAMOCC. 

I'll have a more detailed look into all your changes latest on Tuesday afternoon.

TomasTorsvik (approved):
Thanks @JorgSchwinger ! It's nice to have the netcdf routines in a single module.

hamocc/mo_apply_rivin.F90
```diff
@@ -94,46 +97,64 @@ subroutine apply_rivin(kpie,kpje,kpke,pddpo,omask,rivin)
...
-                 ocetra(i,j,1:kmle(i,j),idoc14)/(ocetra(i,j,1:kmle(i,j),idoc)+safediv)*                 &
-                 rivin(i,j,irdoc)*fdt/volij
-
-            ocetra(i,j,1:kmle(i,j),idet13)     = ocetra(i,j,1:kmle(i,j),idet13)     +                   &
-                 ocetra(i,j,1:kmle(i,j),idet13)/(ocetra(i,j,1:kmle(i,j),idet)+safediv)*                 &
-                 rivin(i,j,irdet)*fdt/volij
-            ocetra(i,j,1:kmle(i,j),idet14)     = ocetra(i,j,1:kmle(i,j),idet14)     +                   &
-                 ocetra(i,j,1:kmle(i,j),idet14)/(ocetra(i,j,1:kmle(i,j),idet)+safediv)*                 &
-                 rivin(i,j,irdet)*fdt/volij
+            ocetra(i,j,1:kmle(i,j),isco213) = ocetra(i,j,1:kmle(i,j),isco213) +                    &
```
jmaerz: Hi @JorgSchwinger and @TomasTorsvik , here and throughout the document: when we do line continuation: what is good practice in terms of where to put the math symbol? - thus far, I personally preferred to have it in the next line, as in (pseudo-code):
```
ocetra = ocetra*xxx &
               + yyy
```
to know, what is happening with the stuff in the continuation line. But is there common agreement on this? - at the moment, I see a mixed approach... - not sure, if we should change this, though.

hamocc/mo_control_bgc.F90
```diff
@@ -48,34 +48,34 @@ module mo_control_bgc
...
+  logical           :: l_3Dvarsedpor          = .false. ! apply spatially variable sediment porosity
+  logical           :: do_ndep                = .true.  ! apply n-deposition
+  logical           :: do_rivinpt             = .true.  ! apply riverine input
+  logical           :: do_sedspinup           = .false. ! apply sediment spin-up
+  logical           :: do_oalk                = .false. ! apply ocean alkalinization
+  logical           :: with_dmsph             = .false. ! apply DMS with pH dependence
+  integer           :: sedspin_yr_s           = -1      ! start year for sediment spin-up
+  integer           :: sedspin_yr_e           = -1      ! end   year for sediment spin-up
+  integer           :: sedspin_ncyc           = -1      ! sediment spin-up sub-cycles
   character(len=64) :: ocn_co2_type
```
jmaerz: One could set a comment, which values `ocn_co2_type` can hold (and what it is good for).

hamocc/mo_dipowa.F90
```diff
@@ -143,7 +144,7 @@ subroutine dipowa(kpie,kpje,kpke,omask,lspin)
           l = ks-k
           do i = 1,kpie
             if (omask(i,j) > 0.5) then
-              powtra(i,j,l,iv) = ( sedb1(i,l,iv)                               &
+              powtra(i,j,l,iv) = ( sedb1(i,l,iv)                                                   &
```
jmaerz, line 147: doesn't fit in one line? - but maybe its just a bit over the line - not sure, if we should see this a bit sloppy...

hamocc/mo_hamocc4bcm.F90
```diff
@@ -75,36 +71,36 @@ subroutine hamocc4bcm(kpie,kpje,kpke,kbnd,kplyear,kplmon,kplday,kldtday,&
...
+    real,    intent(in)  :: pdlyp  (kpie,kpje)                              ! size of grid cell (latitudinal) [m].
+    real,    intent(in)  :: pddpo  (kpie,kpje,kpke)                         ! size of grid cell (depth) [m].
+    real,    intent(in)  :: prho   (kpie,kpje,kpke)                         ! density [g/cm^3].
+    real,    intent(in)  :: pglat  (1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd)      ! latitude of grid cells [deg north].
+    real,    intent(in)  :: omask  (kpie,kpje)                              ! land/ocean mask.
+    real,    intent(in)  :: dust   (kpie,kpje)                              ! dust deposition flux [kg/m2/month].
+    real,    intent(in)  :: rivin  (kpie,kpje,nriv)                         ! riverine input [kmol m-2 yr-1].
+    real,    intent(in)  :: ndep   (kpie,kpje)                              ! nitrogen deposition [kmol m-2 yr-1].
+    real,    intent(in)  :: oafx   (kpie,kpje)                              ! alkalinity flux from alkalinization [kmol m-2 yr-1]
+    real,    intent(in)  :: pi_ph  (kpie,kpje)
```
jmaerz: Hi Jörg, can you add a comment, what `pi_ph` is?

hamocc/mo_hamocc4bcm.F90
```diff
@@ -237,7 +233,7 @@ subroutine hamocc4bcm(kpie,kpje,kpke,kbnd,kplyear,kplmon,kplday,kldtday,&
     endif
 
     call carchm(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,prho,pglat,omask,      &
```
jmaerz: linebreak needed?

jmaerz (approved):
Hi @JorgSchwinger , many thanks again to go through the code that carefully to check the previously done automatic indentation, etc. Perceive the most comments that I was doing as options and address them or not (as you like). I guess, the only thing, where I would be in favor of, would be to break the line **before** operators. But we can also keep it as mixed as it is at the moment. Feel free to push it as it is at the  moment.

hamocc/mo_apply_rivin.F90
TomasTorsvik: I don't have a strong opinion on this, but I tend to put the math operator first on a new line and not last on the preceding line.

JorgSchwinger: I will change this here (first in the new line), since it is a mix here, but I won't go through the code.

jmaerz: Sure. Fair enough. As written, these things I feel as rather minor, anyway.

hamocc/mo_dipowa.F90
JorgSchwinger: No, doesn't fit. Since it was like this before, I left it.

hamocc/mo_hamocc4bcm.F90
JorgSchwinger: done

hamocc/mo_hamocc4bcm.F90
JorgSchwinger: fits in a single line indeed!

jmaerz:
Many thanks! - I'll make the "HAMOCC ifdefs" ready the next hour to enable compiling BLOM without iHAMOCC.

PR #309, opened 2023-11-14
Exclude HAMOCC from compilation for cmp-flag ecosys=false

jmaerz:
Hi @TomasTorsvik , @JorgSchwinger and @mvertens , as discussed among some of us, I reverted back Marianas changes wrt to the compilation of HAMOCC in case it is a physics-only setup. So essentially, the HAMOCC code part is only compiled, when it should be compiled, which can save compile time.  

This will (likely) be the last commit for the CMIP6-compatible HAMOCC version (not sure about BLOM, though). **Question:** How shall we handle the tagging #295 (or can/shall this be a separate commit - also since I am currently unaware, if there will be some BLOM commits)?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -68,7 +68,6 @@ subdir('fuk95')
 subdir('channel')
 subdir('single_column')
 subdir('pkgs/')
-subdir('hamocc')
 
 # Handle options and add necessary flags and subfolders with source files
 
@@ -115,6 +114,7 @@ endif
 
 if get_option('ecosys')
   add_project_arguments('-DHAMOCC', language: 'fortran')
+  subdir('hamocc')
 endif
 
 if get_option('levitus2x')
--- a/trc/initrc.F
+++ b/trc/initrc.F
@@ -23,7 +23,9 @@ c --- ------------------------------------------------------------------
 c --- initialization of tracers
 c --- ------------------------------------------------------------------
 c
+#ifdef HAMOCC
       use mo_hamocc_init, only: hamocc_init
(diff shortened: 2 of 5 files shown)


JorgSchwinger (approved):

mvertens (approved):

mvertens:
@jmaerz - good question about the tagging. I think this would be a good time to come to a consensus about how we want to move forwards with this.

jmaerz:
Wrt to tagging, I currently would suggest to leave the discussion to #295 and doing a final PR+commit via adding the required script for github actions (see https://github.com/NorESMhub/BLOM/discussions/295#discussioncomment-7390610) - so that merging this PR can go ahead (which also enables working on the merging of `master` into the `beyond-CMIP6` branch) - even though I am aware that we should come to a consensus in that question. There was also the wish by Mats to postpone tagging until 15th of November (so maybe there is a merge from his side coming up, which would also suggest to perform an initial tag once that his work is merged).

TomasTorsvik (approved):

TomasTorsvik:
For this specific tag I suggest to use `v1.4.0`, to be consistent with previous CMIP6 compatible releases. Then we should probably abandon this tag name convention for something more descriptive.

mvertens:
@TomasTorsvik - your tagging suggesting for this tag sounds good.

JorgSchwinger:
@TomasTorsvik I agree, too. Are you going to prepare the blom `v1.4.0` release?

PR #314, opened 2023-11-20
Properly declare and initialize bifrXX_ini and clean mo_biomod from growth_co2

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , this is the PR to properly declare and initialize the `bifrXX`. In addition, I cleaned `mo_biomod` from `growth_co2` since it was/is only used as local variable in `mo_ocprod`.

After this PR, I believe that the `beyond-CMIP6` branch is ready to be merged into `master`, which eventually breaks CMIP6 backward compatibility of `master` and makes it ready for NorESM2.1. Technically, the tagging #295 can then still be included in `master`...


JorgSchwinger (approved):
Looks fine to me, good to have this done.

PR #316, opened 2023-12-01
Hybrid coordinate enhancements 3

matsbn:
- Modified the transition between isopycnic and constant pressure interfaces to make it less sensitive to bathymetry variations.
- Modified the estimation of neutral slope vector times Brunt-Vaisala frequency in the vicinity of bathymetry to make Eady growth rate estimates more robust.
- Moved BLOM_VCOORD from run_component_blom to build_component_blom group (BLOM_VCOORD is needed for building since it defines the number of vertical layers used).

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -65,8 +65,8 @@
     <type>char</type>
     <valid_values>isopyc_bulkml,cntiso_hybrid</valid_values>
     <default_value>isopyc_bulkml</default_value>
-    <group>run_component_blom</group>
-    <file>env_run.xml</file>
+    <group>build_component_blom</group>
+    <file>env_build.xml</file>
     <desc>Vertical coordinate type of BLOM</desc>
   </entry>
 
--- a/phy/mod_cmnfld_routines.F90
+++ b/phy/mod_cmnfld_routines.F90
@@ -1,5 +1,5 @@
 ! ------------------------------------------------------------------------------
-! Copyright (C) 2015-2022 Mats Bentsen, Mehmet Ilicak
+! Copyright (C) 2015-2023 Mats Bentsen, Mehmet Ilicak
 !
 ! This file is part of BLOM.
 !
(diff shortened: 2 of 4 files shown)


cime_config/config_component.xml
```diff
@@ -65,8 +65,8 @@
     <type>char</type>
     <valid_values>isopyc_bulkml,cntiso_hybrid</valid_values>
     <default_value>isopyc_bulkml</default_value>
-    <group>run_component_blom</group>
-    <file>env_run.xml</file>
+    <group>build_component_blom</group>
```
mvertens, line 68: Thanks for fixing this.

mvertens (approved):

cime_config/config_component.xml
mvertens: Do we do any tests that verify that OMP works - such as that changing the number of threads gives bfb results? 
I have run into some perplexing differences when I ran an MPI case with threads = 1 and with threads = 2. 
I need to do some more testing before raising an issue. But I'm wondering what tests are done to verify the functionality of OMP threading?

JorgSchwinger: Does this mean that hybrid coordinates would not work in release v.1.4.0?

mvertens: It should work - my comment is only that the xml variable BLOM_VCOORD should be in env_build.xml and not in env_run.xml - since it sets dimensions.
That said - I really think we need to expand the test list to include  hybrid coordinates and more testing of OMP and MPI tasks variations.

mvertens:
Good news - I have run a test which successfully passed and checks bfb across different thread counts:
PET_Ld3.T62_tn14.NOINY.betzy_intel 
The test verifies that running with 2 threads versus 1 thread gives bfb results for both the coupler history file and the blom history file
PET_Ld3.T62_tn14.NOINY.betzy_intel.v1401.blom.hd.0001-01-03.nc
I should verify the same with ihammoc.

TomasTorsvik:
All CI tests with intel compiler fails, but don't see anything in this PR that should cause this change. The `macos-latest` fails at run time, but this seems to b failing also on current `master`. Probably we need to spend some time fixing CI, but I don't think it's specific to this PR.

TomasTorsvik (approved):

cime_config/config_component.xml
matsbn: Hybrid coordinates worked in v1.4.0 by modifying BLOM_VCOORD in env_run.xml, but you would have to remember to do it before building.

mvertens:
Also the PET test PET_Ld3.T62_tn14.NOINYOC.betzy_intel.blom-hamocc1 passes.

mvertens:
Note - that I did these tests with dev1.4.0.1 and not this PR.

matsbn:
> All CI tests with intel compiler fails, but don't see anything in this PR that should cause this change. The `macos-latest` fails at run time, but this seems to b failing also on current `master`. Probably we need to spend some time fixing CI, but I don't think it's specific to this PR.

I also saw the failed CI tests. Many of the tests failed with "configure: error: Fortran could not compile .f90 files". It seems unlikely that this PR is causing this and that the issue is on the CI system.

mvertens:
I had the exact same CI failures in my BLOM refactor pushes to my fork. So its definitely not this PR.

mvertens:
Who will be making the v1.5.0 tag? That would be really helpful to have as soon as possible.

TomasTorsvik:
I can make a draft release.

PR #321, opened 2023-12-08
Remove science_support attribute for compsets

### LINKED ISSUE #320: science support attributes for compsets in BLOM for v1.5.0
It looks like there is still the science_support attribute set in the blom config_compsets.xml.
I think this attribute is not being used correctly in BLOM.
First - if that is the case - there should always be a test associated with that. 
Second - there should be released simulations for the science supported configurations.
Do we still want to keep this for the noresm2.1 release?

TomasTorsvik:
BLOM/iHAMOCC compsets will no longer have science_support attributes in connection with the noresm2.1 release.

If this commit is merged, it will be tagged as `v1.5.1` on the `release-1.5` branch.

closes #320

### DIFF
--- a/cime_config/config_compsets.xml
+++ b/cime_config/config_compsets.xml
@@ -40,6 +40,7 @@
     <alias>NOINY_WW3</alias>
     <lname>2000_DATM%NYF_SLND_CICE_BLOM_DROF%NYF_SGLC_WW3DEV</lname>
   </compset>
+
   <compset>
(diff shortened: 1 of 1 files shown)


mvertens (approved):
@TomasTorsvik - thanks so much for doing this!!

gold2718 (approved):
:+1:

JorgSchwinger (approved):

matsbn (approved):

PR #323, opened 2023-12-12
Remove science_support attribute for compsets

### LINKED ISSUE #320: science support attributes for compsets in BLOM for v1.5.0

TomasTorsvik:
The `master` branch has now diverged from CMIP6, so it would make sense to remove the science_support attribute from all compsets until a new scientifically supported NorESM version is created.

Closes #320

### DIFF
--- a/cime_config/config_compsets.xml
+++ b/cime_config/config_compsets.xml
@@ -40,6 +40,7 @@
     <alias>NOINY_WW3</alias>
     <lname>2000_DATM%NYF_SLND_CICE_BLOM_DROF%NYF_SGLC_WW3DEV</lname>
   </compset>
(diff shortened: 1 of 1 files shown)


JorgSchwinger (approved):

TomasTorsvik:
I will go ahead with this PR. If for some reason the science_support attribute is needed on master this change can easily be reverted.

PR #326, opened 2024-01-15
Re-organize config_pes.xml file

### LINKED ISSUE #325: Suggestion to re-structure config_pes.xml file
The config_pes.xml file is a bit messy. I would like to clean it up a bit, and add some more options for running compsets on the "devel" queue on Betzy, i.e. pecount=128. I guess it doesn't make sense to go lower than that?

TomasTorsvik:
Add options for 1-node configurations (128 CPUs) on Betzy.

### DIFF
--- a/cime_config/config_pes.xml
+++ b/cime_config/config_pes.xml
@@ -1,81 +1,129 @@
 <?xml version="1.0"?>
 
+<!-- OMIP-type compsets:
+     On betzy: single node (XS) = 128 CPU cores   ("devel" job size)
+               4 nodes = 512 CPU cores            (minimum "normal" size)
+-->
+
 <config_pes>
+
+  <!-- Configurations for the one degree tripolar (tnx1) grid  !-->
 
   <grid name="a%T62.+oi%tnx1v1">
     <mach name="any">
       <pes pesize="M" compset="any">
-	<comment>none</comment>
-	<ntasks>
-	  <ntasks_atm>96</ntasks_atm> 
-	  <ntasks_rof>96</ntasks_rof> 
-	  <ntasks_ice>96</ntasks_ice> 
-	  <ntasks_ocn>216</ntasks_ocn> 
-	  <ntasks_cpl>96</ntasks_cpl> 
-	  <ntasks_lnd>96</ntasks_lnd>
-	  <ntasks_glc>96</ntasks_glc> 
-	  <ntasks_wav>96</ntasks_wav> 
-	</ntasks>
-	<nthrds>
-	  <nthrds_atm>1</nthrds_atm>                   
-	  <nthrds_lnd>1</nthrds_lnd> 
-	  <nthrds_rof>1</nthrds_rof> 
-	  <nthrds_ice>1</nthrds_ice> 
-	  <nthrds_ocn>1</nthrds_ocn> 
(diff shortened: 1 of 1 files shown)


cime_config/config_pes.xml
```diff
@@ -1,81 +1,129 @@
...
+<!-- OMIP-type compsets:
+     On betzy: single node (XS) = 128 CPU cores   ("devel" job size)
+               4 nodes = 512 CPU cores            (minimum "normal" size)
+-->
+
 <config_pes>
 
+  <!-- Configurations for the one degree tripolar (tnx1) grid  !-->
+
   <grid name="a%T62.+oi%tnx1v1">
```
JorgSchwinger, line 12: I am wondering if we need to support the tnx1v1 and tnx1v3 grids  any longer. As far as I know, they were never used for any production runs(?) @matsbn might have an opinion about this.

cime_config/config_pes.xml
```diff
@@ -478,203 +620,166 @@
...
-	  <rootpe_atm>0</rootpe_atm> 
-	  <rootpe_lnd>1</rootpe_lnd> 
-	  <rootpe_rof>1</rootpe_rof> 
-	  <rootpe_ice>2</rootpe_ice>    
-	  <rootpe_ocn>32</rootpe_ocn>   
-	  <rootpe_glc>1</rootpe_glc> 
-	  <rootpe_wav>1</rootpe_wav> 
-	  <rootpe_cpl>0</rootpe_cpl>                         
-	</rootpe>
+      <pes pesize="S" compset="_DATM.*_BLOM%ECO">
```
JorgSchwinger, line 637: This is confusing: There is on pesize="S" layout with 155 ocean tasks, one pesize="M" layout with 155 ocean tasks further below, and one pesize="M" layout with 272 ocean tasks. It is a bit difficult to see what is new and what has been there before.

JorgSchwinger (commented):
Thank you Tomas. I am not sure how far we want to maintain "old" pe-layouts. In my opinion, on the way forward to NorESM2.5 there is not really the need to keep everything.

cime_config/config_pes.xml
TomasTorsvik: I have never used tnx1v1 and tnx1v3. I did not remove any existing definitions from the file, but I can do it if we all agree.

TomasTorsvik:
At the moment I have kept all the old definitions, just moved some things around, and added two `XS` definitions for `T62_tn14` and `TL319_tn14`. I agree that we could simplify the file, as we are now mostly running on Betzy, where the relevant pe-layouts are 128 CPUs for the "devel" queue and 512 or larger for the "normal" queue.

cime_config/config_pes.xml
TomasTorsvik: These are old pe-layouts, I only added new `XS` layouts.
I agree that it is confusing, but I didn't want to remove something that someone might still be using.

TomasTorsvik:
@JorgSchwinger @jmaerz @matsbn @milicak 
Hi, if there are no objections, I would like to merge this as it makes possible running short tests on the "devel" queue on Betzy.

jmaerz (approved):
Hi @TomasTorsvik , I can't comment much on it from my side. So from my side no objections.

PR #328, opened 2024-01-30
Keep only `user_nl_blom`-changed values for namelist `/bgcparams/`

### LINKED ISSUE #327: Namelist `/bgcparams/` writing via `buildnml`
Dear @mvertens , @JorgSchwinger , @TomasTorsvik ,
with bringing in the code changes done last autumn to my nitrogen cycle branch, I am facing the issue that model parameters set via the `/bgcparams/` namelist cannot be changed via the python-based `buildnml` file (it should be the same issue for current `master`). It requires to modify a few lines in `buildnml` to make that happen (otherwise, it just fails). Once done, any parameter found in `namelist_definition_blom.xml` belonging to `/bgcparams/` will be written to `/bgcparams/`. In my opinion, this clutters information on which parameters have been tuned via `user_nl_blom`. We currently hard-coded default parameters in `mo_param_bgc.F90` by purpose and I wonder, if we should only write parameters to `/bgcparams/` that were found in `user_nl_blom` to keep the entries in `/bgcparams/` as informative as possible and to not having to maintain two places with correct/identical parameter values (i.e. both, `mo_param_bgc.F90` and `namelist_definition_blom.xml`). It should be rather simple to strip off other parameters via the `ocn_in_paramgen.py`. Any opinion about it? Or is there a push to set all parameters via the xml-file (I could see some benefits, but being rather hesitant in this)?

jmaerz:
This is a suggestion on how to tackle the issue on hard-coded versus xml-coded parameters. Only parameters changed through `user_nl_blom` will be kept to write into the nml-group `/bgcparams/`. All other groups are not affected with this approach and it is minimally invasive. I tested with some parameters set to `None` in the `namelist_definition_blom.xml` (to avoid confusions) and it works as well. Model build works as well for ocean only setups and coupled simulations (tested until `./case.build`).

This means that essentially all variables can be defined in the Fortran part (and with `None`-value in the `namelist_definition_blom.xml` and only tuning parameters set via the `user_nl_blom` will be forwarded to `ocn_in`. No further preprocessor-flag is required to distinguish between coupled or standalone runs and in case that at some point we decide that the xml-way is the only way, all infrastructure is already there - only the values need to be once transferred.   

Closes #327

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -260,6 +260,17 @@ def buildnml(case, caseroot, compname):
         # for now don't write out bgcparams - just create an empty namelist
         if "ecosys" in case.get_value("BLOM_TRACER_MODULES"):
              groups.append("bgcparams")
+             """
+             !!! The following line deletes all entries for group 'bgcparams' 
+             !!! that are ONLY present in namelist_definition_blom.xml 
+             !!! and keeps entries provided through user_nl_blom
+             !!! This makes sure that i) only Fortran-hard coded values from mo_param_bgc.F90
+             !!! are used and that ii) user_nl_blom enables tuning these default parameters.
+             !!! Note that an entry in namelist_definition_blom.xml for the tuning parameter 
+             !!! is still required, while its value can be set to None. 
+             !!! This is an interim solution and the line below can be simply deleted, once 
+             !!! xml is used as default for the bgcparams namelist as well.
+             """
              pg_blom.keep_usernml_only('bgcparams')
 
         pg_blom.write_nmlfile(namelist_file, groups)
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -258,10 +258,11 @@ def buildnml(case, caseroot, compname):
             groups.append("config_bgc")
 
         # for now don't write out bgcparams - just create an empty namelist
-        # if "ecosys" in case.get_value("BLOM_TRACER_MODULES"):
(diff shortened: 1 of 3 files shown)


JorgSchwinger (approved):
For now this solution seems good to me. In addition we could remove the default values from the `namelist_definition_blom.xml` (since they are not used) to avoid confusion.

jmaerz:
Done. I also tested this approach via `./preview_namelists` and it still works as expected (both, ocean only and coupled setups), also when all `bgcparams` parameter are set to `None` in the xml-file.

mvertens:
@jmaerz - this is a very nice implementation. However, I still feel that this is not the best way to proceed.  You will need to look in two places to see namelist setting - the code  for namelist values that do not appear and the actual namelist file. In fact you have no idea from looking at the namelist file what the set of namelist parameters are. Ultimately, this will lead in my opinion to a less robust way to carry out experiments. I think that the implementation proposed by @gold2718 would be my choice. However, it is of course up to the iHAMOCC group to decide what they want to do - and if this is the solution it wants to adopt. If this is chosen - I am fine with the implementation.

mvertens (approved):

mvertens:
@jmaerz - I just looked at the PR again and see that you are proposing this as just an interim solution. If this is the case and the original implementation will be ultimately used - I think this approach is fine.  However, longer term I think we should implement the proposed solution suggested by @gold2718.

PR #329, opened 2024-02-08
Write parameter values as global attributes to nc-file

jmaerz:
This is an implementation and PR to write model parameters as global attributes to each individual netcdf file. It enables i) better tracing of used parameter values and ii) can be easier used in scripts to trace parameter changes during tuning processes. The code is tested on  betzy with a 1month run.   

**If there is a wish for it, we can also make it optional by adding a switch.**


hamocc/mo_ncout_hamocc.F90
```diff
@@ -221,6 +221,9 @@ subroutine ncwrt_bgc(iogrp)
     call ncdims('ks',ks)
     call ncdims('bounds',2)
     call ncdims('time',0)
+    do i=1,nentries
+        call ncputr(param4nc(i)%pname,param4nc(i)%pvalue)
```
JorgSchwinger, line 228: For the result in the nc-file, it would be maybe nicer to have `call ncputr('Parameter value for '//param4nc(i)%pname,param4nc(i)%pvalue)` or something similar.

hamocc/mo_param_bgc.F90
```diff
@@ -728,8 +739,43 @@ subroutine write_parambgc()
       write(io_stdo_bgc,*) '*   opaldens     = ',opaldens
       write(io_stdo_bgc,*) '*   calcdens     = ',calcdens
       write(io_stdo_bgc,*) '*   claydens     = ',claydens
+      call add_entry('orgdens',orgdens,'','')
+      call add_entry('calcdens',calcdens,'','')
+      call add_entry('claydens',claydens,'','')
```
JorgSchwinger: A more descriptive name for the routine might be better? E.g. `pinfo_add_entry`

JorgSchwinger (approved):
I'm not sure how useful this will be, but I also see no drawbacks -  so fine for me.

TomasTorsvik (approved):
Thanks Jöran, it should be fine to add parameter output to netcdf files.

hamocc/mo_ncout_hamocc.F90
jmaerz: Hi @JorgSchwinger , this results in ugly formatting of global attributes:
globa-attr
while keeping it simple:
globa-attr-simple
also eases usage in scripts (output of ncdump - h). Hence, I would stay with the current version.

hamocc/mo_param_bgc.F90
jmaerz: Agree, I'll adopt your suggestion (as written initially, I also didn't like the naming).

hamocc/mo_ncout_hamocc.F90
JorgSchwinger: I don't find the 1st one particularly ugly... that's what ncdump always does with blanks. But anyway, the original is ok, too.

jmaerz: If you prefer something like `parameter_xxxx`, we can write this during the call to `pinfo_add_entry`.

PR #330, opened 2024-02-09
Hybrid coordinate enhancements 4

matsbn:

- Corrected extraction of polynomial coefficients of a reconstruction.
- Modified the regridding by interface nudging to better handle discontinuous density reconstruction at interfaces.
- If the system of equations solved for 6th order accurate edge estimates and 5th order accurate slope estimates leads to tridiagonal matrices that are not diagonally dominant, reduce the order of accuracy locally.
- Added flexibility in selecting the order of accuracy for the explicit scheme for edge and slope estimation at boundaries. The maximum order for the left and right boundaries can be specified and the order will be automatically reduced if the available grid cells cannot support the requested order.
- Avoid division by zero in the PQM limiting when the range of source values is zero.
- Added namelist variables for specifying the maximum order of accuracy for the explicit scheme for edge and slope estimation at the upper and lower boundaries.
- Changed default namelist variables for vcoord_type = cntiso_hybrid.
- Added array data structures for range and small values associated with the source data.

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -794,6 +794,26 @@
     <desc>add desc</desc>
   </entry>
 
+  <entry id="upper_bndr_ord">
+    <type>integer</type>
+    <category>vcoord</category>
+    <group>vcoord</group>
+    <values>
+      <value>6</value>
+    </values>
+    <desc>add desc</desc>
+  </entry>
+
+  <entry id="lower_bndr_ord">
+    <type>integer</type>
+    <category>vcoord</category>
+    <group>vcoord</group>
+    <values>
+      <value>4</value>
+    </values>
+    <desc>add desc</desc>
+  </entry>
+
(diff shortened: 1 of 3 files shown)


cime_config/namelist_definition_blom.xml
```diff
@@ -1041,13 +1067,15 @@
     <category>diffusion</category>
     <group>diffusion</group>
     <values>
-      <value>1500.</value>
+      <value >1500.</value>
```
TomasTorsvik:
```suggestion
      <value>1500.</value>
```
Looks like a whitespace error

TomasTorsvik (approved):
The namelist changes look fine to me. No obvious red flags in vcoord and hor3map, but I did not look in details on those routines.

matsbn:
Also tested without obvious issues for a few years with tnx0.125v4 using both PPM and PQM reconstructions. Will merge now.

PR #334, opened 2024-03-22
Fix bug in cgs to mks conversion for tracer field

JorgSchwinger:
This PR fixes a deviation from expected results when switching the `BLOM_UNIT` option to `mks`. With this fix global integrated time series remain close to the previous results, and spatial variation are reduced much. There are some remaining differences, but these are relatively small and not larger (and in the same places) than in physical variables (SST and SSS). As shown below for alkalinity, there were huge differences after 100 years in simulations without this fix (labeled 34), which are greatly reduced in simulations including this fix (labelled 35). 

The treatment of virtual fluxes is still not (and was not) correct, I believe, but I suggest to fix this later, and first establish the expected behavior when moving to mks.

NorDiag_2d_T62_tn21_31-T62_tn21_34_kms_old_Global_lev01_diff_0100_srftalk
NorDiag_2d_T62_tn21_31-T62_tn21_35_kms_new_Global_lev01_diff_0100_srftalk
[NorDiag_gts_T62_tn21_31-T62_tn21_34-T62_tn21_35_mkstest_bgc_0001-0100.pdf](https://github.com/NorESMhub/BLOM/files/14720125/NorDiag_gts_T62_tn21_31-T62_tn21_34-T62_tn21_35_mkstest_bgc_0001-0100.pdf)

### DIFF
--- a/cesm/thermf_cesm.F
+++ b/cesm/thermf_cesm.F
@@ -422,7 +422,8 @@ c$OMP PARALLEL DO PRIVATE(l,i)
         do j=1,jj
           do l=1,isp(j)
           do i=max(1,ifp(j,l)),min(ii,ilp(j,l))
-            trflx(nt,i,j)=-(trflx(nt,i,j)+trflxc)*L_mks2cgs
+            trflx(nt,i,j)=-(trflx(nt,i,j)+trflxc)
+     .                     *(kg2g*(M_mks2cgs/L_mks2cgs**2))
           enddo
           enddo
         enddo


jmaerz (approved):
Hi Jörg, many thanks for looking into this (again)! I agree that we can merge this now and re-check the virtual fluxes afterward. Seems like we were already close in #233 with our analysis ... (and just to link the documentations, #233 made the required changes in the BLOM-iHAMOCC interface to enable switching between `cgs` and `mks` units).

matsbn (approved):
This is great, @JorgSchwinger! Sorry for the late review.

JorgSchwinger:
For the record, I also checked that this results in bit-identical results if cgs is used.

PR #335, opened 2024-04-16
New and updated topographic beta files.

matsbn:
New and updated topographic beta files, prepared by Aleksi Nummelin, are specified for grids tnx2v1, tnx1v4, tnx0.25v4 and tnx0.125v4. The updated topographic beta file for grid tnx1v4 has been tested for one cycle of the OMIP2 protocol. Diagnostics with the previous topographic beta file can be found here:

https://ns1012k.web.sigma2.no/diagnostics/noresm/matsbn/NOIIAJRAOC20TR_TL319_tn14_ppm_notz_20240312/

while diagnostics with the updated topographic beta file can be found here:

https://ns1012k.web.sigma2.no/diagnostics/noresm/matsbn/NOIIAJRAOC20TR_TL319_tn14_topobeta_20240410/

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -1097,7 +1097,10 @@
     <group>diffusion</group>
     <values>
       <value>unset</value>
-      <value ocn_grid="tnx1v4">$DIN_LOC_ROOT/ocn/blom/bndcon/topo_beta_tnx1v4_20170622_python_coarse_nc64bitoffset.nc</value>
+      <value ocn_grid="tnx2v1">$DIN_LOC_ROOT/ocn/blom/bndcon/topo_beta_tnx2v1_20130206_depth_python_SYNBATH_V1.1_3minute_filt_max100km_nc64bitoffset.nc</value>
+      <value ocn_grid="tnx1v4">$DIN_LOC_ROOT/ocn/blom/bndcon/topo_beta_tnx1v4_20170622_depth_python_SYNBATH_V1.1_3minute_filt_max100km_nc64bitoffset.nc</value>
+      <value ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/topo_beta_tnx0.125v4_20221013_depth_python_SYNBATH_V1.1_3minute_filt_max100km_nc64bitoffset.nc</value>
+      <value ocn_grid="tnx0.125v4">$DIN_LOC_ROOT/ocn/blom/bndcon/topo_beta_tnx0.125v4_20221013_depth_python_SYNBATH_V1.1_3minute_filt_max100km_nc64bitoffset.nc</value>
     </values>
     <desc>Name of file containing topographic beta parameter (a)Name of file containing topographic beta parameter</desc>
   </entry>


TomasTorsvik:
I see there are some changes in transports (e.g. Drake Passage)

https://ns1012k.web.sigma2.no/diagnostics/noresm/matsbn/NOIIAJRAOC20TR_TL319_tn14_ppm_notz_20240312/BLOM_DIAG/yrs1694to1713-obs/set1/set1_ann_voltr5_1model.png

https://ns1012k.web.sigma2.no/diagnostics/noresm/matsbn/NOIIAJRAOC20TR_TL319_tn14_topobeta_20240410/BLOM_DIAG/yrs1694to1713-obs/set1/set1_ann_voltr5_1model.png

I don't have any objections to these new files, but it would be useful to provide some more background info about this change, perhaps in the documentation pages?

TomasTorsvik (approved):

AleksiNummelin (approved):
Tomas asked for some more background, and I think that would be good to provide. I can just shortly mention here a couple of things

- These files are based on first regridding the highres bathymetry to 3 minute grid (to save computational time and allow the filtering to work in the first place), calculating deformation radius (WBK approximation) from WOA2023 data on 0.25 degree grid and mapping that to the same grid as bathymetry, and then finally defining a spatial filter which has locally the min of deformation radius and 100 km (i.e. deformation radius capped to 100 km) as a length scale. After this, we calculate the topographic beta and map it to each grid. In addition each file includes the slope angle, both calculated from the highres bathymetry and from the model bathymetry. At the moment in the code we use the model bathymetry based slope angle to limit the effects of the bathymetry on eddy mixing when the flow is not aligned with the bathymetry, but here one could test using the highres bathymetry based slope angle instead. Regridding is done with gcm_filters python package.
- The topography aware eddy parameterization also introduced a few additional parameters in the namelist. In particular, switching EDDF2D = .true., RHSCTP = .true., and  EDANIS = .true. is a tested and recommended setting. However, there is still the option to use EDSPRS = .true. which is part of the old formulation. However, it is not recommended to use EDSPRS = .true. with RHSCTP = .true. and/or EDSPRS = .true. In fact, the formulations behind EDANIS and EDSPRS are quite similar, but it is really the EDANIS that is designed to be used with RHSCTP. Also, RHSCTP is meant to be used in the 2D formulation i.e. EDDF2D = .true. and if topography aware parameterization with 3D structure is desired, one should use REDI3D = .true. rather than EDDF2D = .false. although this is not well tested and should be worked on in the future.

The main reference for the eddy parameterization is [Nummelin and Isachsen (2024)]( https://doi.org/10.1029/2023MS003806) and further support is provided by [Sterl et al. (2024)](https://doi.org/10.1175/JPO-D-23-0142.1).

jmaerz:
Dear @AleksiNummelin , many thanks for the detailed description and documentation of those files!

TomasTorsvik:
Thanks @AleksiNummelin for the detailed description. I suppose the topography aware eddy parameterization will be the default for NorESM2.3, so we should make sure this is mentioned in the NorESM2.3 documentation. We haven't started on a NorESM2.3 documentation yet, but it's probably time to do so.

PR #337, opened 2024-04-29
Changed default eddy parameterization for isopycnic vertical coordinate

matsbn:
Changed the default eddy parameterization for vcoord_type = isopyc_bulkml to the same that is currently default for vcoord_type = cntiso_hybrid.

The associated previous default namelist values for vcoord_type = isopyc_bulkml were:

```
EDANIS = .false.
EDDF2D = .false.
EDSPRS = .true.
RHSCTP = .false.
EGC = 0.85
EGMNDF = 100.
EGMXDF = 1500.
RHISCF = 0.
```

Note that EGC and EGMXDF had grid specific values and that EGMNDF and EGMXDF are dimensional parameters, here provided with unit m<sup>2</sup>s<sup>-1</sup>. New default namelist values are:

```
EDANIS = .true.
EDDF2D = .true.
EDSPRS = .false.
RHSCTP = .true.
EGC = 3.0
EGMNDF = 1.
EGMXDF = 2500.
RHISCF = 0.5
```

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -996,8 +996,7 @@
     <category>diffusion</category>
     <group>diffusion</group>
     <values>
-      <value>.false.</value>
-      <value blom_vcoord="cntiso_hybrid" ocn_grid="tnx1v4">.true.</value>
+      <value>.true.</value>
     </values>
     <desc>If true, eddy diffusivity has a 2d structure (l)</desc>
   </entry>
@@ -1007,8 +1006,7 @@
     <category>diffusion</category>
     <group>diffusion</group>
     <values>
-      <value>.true.</value>
-      <value blom_vcoord="cntiso_hybrid" ocn_grid="tnx1v4">.false.</value>
+      <value>.false.</value>
     </values>
     <desc>Apply eddy mixing suppression away from steering level (l)</desc>
   </entry>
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):
Looks fine to me.

I see that you now have "mks" as the default and "cgs" as the optional, but "cgs" is still the default in `config_component.xml`. Would it make sense to transition to "mks" as default for the NorESM2.3 release?

matsbn:
> Looks fine to me.
> 
> I see that you now have "mks" as the default and "cgs" as the optional, but "cgs" is still the default in `config_component.xml`. Would it make sense to transition to "mks" as default for the NorESM2.3 release?

In my opinion, we should run some coupled tests (with interactive DMS at least) to verify climatological similarity, both with hybrid and isopycnic coordinate. Another thing missing is the ability to restart from a simulation with CGS units and continue with MKS units. Although a bit-reproducible restart is unlikely, I think it would be a useful functionality. I'm a bit pressed on time now, but will see if some progress on this can be made.

TomasTorsvik:
We plan to make a release candidate tag for NorESM2.3 after merging this PR.

AleksiNummelin (approved):
This looks good. Just to reiterate on the namelist options (similar to the recent pull request on the topographic beta files), one should be aware that the new eddy parameterization is designed to have 

EDANIS = .true.
EDDF2D = .true.
EDSPRS=.false.

and in particular having EDSPRS and EDANIS true at the same time would not make sense (they are trying to do more or less the same thing). This information, as well as other details of the  new parameterization, should go into documentation latest by the time of the NorESM3 release.

JorgSchwinger:
Wouldn't it make sense then to have a check of such namelist options to ensure that the settings make sense, and abort (or issue a warning) if both EDSPRS and EDANIS = true?

JorgSchwinger (approved):
Just to understand this: What is the reason that some parameters (e.g. EGC) that were dependent on the grid resolution have the same value for all grids now?

Otherwise it looks fine to me.

matsbn:
> Just to understand this: What is the reason that some parameters (e.g. EGC) that were dependent on the grid resolution have the same value for all grids now?
> 
> Otherwise it looks fine to me.

I do not think the different tunings that have been used, particularly for EGC, for different grids are necessarily transferable with the new eddy parameterization. Fine-tuning must be done individually for each grid, but the current values are likely a good starting point.

PR #341, opened 2024-05-08
add missing configurations for the ATRC option

YanchunHe:
add missing configurations for the ATRC option

ATRC is used as compiler option in BLOM, but it is not defined.

add ATRC option in the namelist building script and in the XML files.

### DIFF
--- a/cime_config/buildlib
+++ b/cime_config/buildlib
@@ -16,6 +16,7 @@ set OCN_NCPL = `./xmlquery OCN_NCPL --value`
 set BLOM_COUPLING = `./xmlquery BLOM_COUPLING --value`
 set RUNDIR = `./xmlquery  RUNDIR --value`
 set BLOM_TRACER_MODULES = `./xmlquery BLOM_TRACER_MODULES --value`
+set BLOM_ATRC = `./xmlquery BLOM_ATRC --value`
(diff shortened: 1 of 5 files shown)


TomasTorsvik:
@YanchunHe 
Hi, I see you start from the `v1.1.0` tag when you add your changes, and want to push to `master`. However, there are some conflicts due to the source code having evolved. The current `master` should contribute to `NorESM2.3`. If you want these changes to be available for CMIP6-version of NorESM2 (currently `NorESM2.0.7` is the latest), changes should be added on top of `v1.4.0`, and for NorESM2.1 the relevant tag is `v1.5.1.1`. Do you want to include the ATRC option for all these NorESM versions, or only the upcoming `NorESM2.3`? In any case, I think it would make the process easier if you start from a more up to date source code.

TomasTorsvik:
@YanchunHe - I have created a new PR #346 that adds this functionality on top of the current `master` branch. If you need this for CMIP6-type simulations, I suggest to start with the `v1.4.0` tag. In either case, I suggest that this should not be merged to `master` as it will create conflicts. I therefore label it "invalid" for now.

YanchunHe:
This has been merged in #346

I close here.

PR #342, opened 2024-05-10
new CESM stream relaxation capability for sss and sst

mvertens:
This introduces new CDEPS stream capability for use of BLOM in CESM to permit new relaxation to salinity and temperature. 
The capability is still triggered by trxday and srxday - but the new time and spatial interpolation are done in the new files:
`drivers/nuopc/ocn_stream_sss.F90 and drivers/nuopc/ocn_stream_sst.F90`

This capability permits spatial interpolation of forcing data as well as the ability to use multi-year time samples (for example daily data for a number of years). The data is not all read in at once - but only two time periods are read in at a time to do the time interpolation. **The advantage of this new capability is that only one relaxation file can be specified and can be used for multiple resolutions.**

The code uses the CDEPS inline stream capability in $SRCROOT/components/cdeps/streams.

New namelist groups stream_sss and stream_sst are introduced that turn this new capability on:

```F90
&stream_sss
  STREAM_SSS_DATA_FILENAME = '/cluster/work/users/mvertens/noresm/blom_sss_stream/run/sss_climatology_N1850_f19_tn14_20190621_1600-1699_cdf5.nc'
  STREAM_SSS_MESH_FILENAME = '/cluster/shared/noresm/inputdata/share/meshes/tnx1v4_20170601_cdf5_ESMFmesh.nc'
  STREAM_SSS_VARNAME = 'sss'
  STREAM_SSS_YEAR_ALIGN = 1
  STREAM_SSS_YEAR_FIRST = 1
  STREAM_SSS_YEAR_LAST = 1
/

&stream_sst
  STREAM_SST_DATA_FILENAME = '/cluster/shared/noresm/inputdata/atm/cam/sst/sst_HadOIBl_bc_1x1_1850_2017_c180507_cdf5.nc'
  STREAM_SST_MESH_FILENAME = '/cluster/shared/noresm/inputdata/atm/cam/sst/sst_HadOIBl_bc_1x1_clim_c101029_ESMFmesh_120520.nc'
  STREAM_SST_VARNAME = 'SST_cpl', 'ice_cov'
  STREAM_SST_YEAR_ALIGN = 1
  STREAM_SST_YEAR_FIRST = 1850
  STREAM_SST_YEAR_LAST = 1850
/
```

Still to do:

- [x]  A preliminary validation was done to show that the salinity and sst using these new streams was reasonable. I've also show the effect of the horizontal interpolation below. 
- [x]  The full BLOM test suite has been run successfully. No baseline comparison was made since the changes are expected to change answers.

### DIFF
--- a/cesm/mod_cesm.F90
+++ b/cesm/mod_cesm.F90
@@ -27,11 +27,10 @@ module mod_cesm
    use mod_constants, only: pi
    use mod_time, only: nstep
    use mod_xc
-   use mod_forcing, only: trxday, srxday, swa, nsf, lip, sop, eva, rnf, rfi, &
+   use mod_forcing, only: swa, nsf, lip, sop, eva, rnf, rfi, &
                           fmltfz, sfl, ztx, mty, ustarw, slp, abswnd, &
                           lamult, lasl, ustokes, vstokes, atmco2, atmbrf, &
                           flxdms, flxbrf
-   use mod_ben02, only: initai, rdcsic, rdctsf, fnlzai
    use mod_seaice, only: ficem
    use mod_checksum, only: csdiag, chksummsk
 #ifdef HAMOCC
@@ -39,7 +38,6 @@ module mod_cesm
 #endif
 
    implicit none
-
    private
 
    character(len = 256) :: &
@@ -117,29 +115,9 @@ contains
    end subroutine inicon_cesm
 
    subroutine inifrc_cesm
-   ! ---------------------------------------------------------------------------
-   ! Initialize climatological fields for surface restoring and interpolation of
-   ! CESM forcing fields.
-   ! ---------------------------------------------------------------------------
-
-      ! If SST restoring is requested, prepare interpolation of surface fields
-      ! and read climatological sea-ice concentration and surface temperature.
-      if (trxday > 0._r8) then
-        call initai
-        call rdcsic
-        call rdctsf
-      endif
-
-      ! If SSS restoring is requested, read climatological sea surface salinity.
-      if (srxday > 0._r8) call rdcsss
 
       ! Initialize diagnosing/application of relaxation fluxes.
       call idarlx
-
-      ! Deallocate memory used for interpolation of surface fields.
-      if (trxday > 0._r8) then
-        call fnlzai
-      endif
 
       ! Initialize time level indexes
       l1ci = 1
--- a/cesm/mod_cesm.F90
+++ b/cesm/mod_cesm.F90
@@ -27,10 +27,11 @@ module mod_cesm
    use mod_constants, only: pi
    use mod_time, only: nstep
    use mod_xc
-   use mod_forcing, only: swa, nsf, lip, sop, eva, rnf, rfi, &
+   use mod_forcing, only: trxday, srxday, swa, nsf, lip, sop, eva, rnf, rfi, &
                           fmltfz, sfl, ztx, mty, ustarw, slp, abswnd, &
                           lamult, lasl, ustokes, vstokes, atmco2, atmbrf, &
-                          flxdms, flxbrf
+                          flxdms, flxbrf, use_stream_relaxation
+   use mod_ben02, only: initai, rdcsic, rdctsf, fnlzai
    use mod_seaice, only: ficem
    use mod_checksum, only: csdiag, chksummsk
 #ifdef HAMOCC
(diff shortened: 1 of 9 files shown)


TomasTorsvik:
@mvertens @matsbn 
From what I see in the code changes, I understand the CESM stream relaxation method replaces the relaxation method provided by `mod_ben02`. I don't know if this capability has been used recenlty, I have not looked into this myself. If this PR is merged, are there parts of `mod_ben02` that becomes redundant?

mvertens:
@TomasTorsvik - this capability is used in many places in CESM to obtain forcing (CAM, CTSM, POP) and of course all of the CDEPS data components. The way I implemented this was to only activate this in CESM - and let ben02 still have backwards compatibility. The key issue is that I wanted to make sure that you could still do stand-alone testing without needing to compile in CDEPS (which then requires PIO2 and ESMF).
The new capability from CDEPS enables you to interpolate with different types of interpolation algorithms in the horizontal - but also provides the ability to do horizontal interpolation on a set of vertical levels and then do vertical interpolation if the levels are different from the model. This functionality is used by POP for 3d relaxation. The key issue is that from the point of view of BLOM - the interfaces to do this are very simple and can be leveraged anywhere in the code base if that is desired.

cesm/mod_cesm.F90
```diff
@@ -27,19 +27,17 @@ module mod_cesm
    use mod_constants, only: pi
    use mod_time, only: nstep
    use mod_xc
-   use mod_forcing, only: trxday, srxday, swa, nsf, lip, sop, eva, rnf, rfi, &
+   use mod_forcing, only: swa, nsf, lip, sop, eva, rnf, rfi, &
                           fmltfz, sfl, ztx, mty, ustarw, slp, abswnd, &
                           lamult, lasl, ustokes, vstokes, atmco2, atmbrf, &
                           flxdms, flxbrf
-   use mod_ben02, only: initai, rdcsic, rdctsf, fnlzai
```
TomasTorsvik, line 34: It seems these functions are no longer used anywhere in BLOM after this PR. Is that intentional?

TomasTorsvik:
@mvertens - thanks for the explanation, generally speaking the suggested changes look reasonable to me. I just wanted to point out that the code change removes calls to some functions provided by `mod_ben02` (`initai`, `rdcsic`, `rdctsf` and `fnlzai`) that are not called anywhere else in the code, so it would seem these routines become redundant.

mvertens:
@TomasTorsvik - sorry for not being clear. Yes - those changes were only needed for cesm - so they were removed. My goal was to still keep the namelist variables srxday and trxday the same for both stand-alone and cesm - but when using cesm those variables would trigger the new stream functionality.

TomasTorsvik (approved):

cesm/mod_cesm.F90
mvertens: Actually there are calls to initai, rdcsic and rdctsf in mod_ben02.F in the routine` inifrc_ben02clim`
```F90
c --- Prepare interpolation of surface fields.
      call initai
c
c --- Read climatological sea-ice concentration.
      call rdcsic
c
c --- Read climatological surface temperature.
      call rdctsf
```

And inifrc_ben02clim is called from inifrc.F90 if expcnf is 'ben02clim' - so I don't think we can remove these routines unless ben02clim is a mode that will no longer be used. Am I missing something?

mvertens:
Screenshot 2024-05-13 at 9 53 38 AM
Screenshot 2024-05-13 at 9 53 56 AM

mvertens:
The following are images showing both the spatial interpolation as well as time interpolation functionality of the new streams capability:
The  figure is just the So_s sent to the mediator by blom. These two figure look identical between what is generated by BLOM master and this PR.
Screenshot 2024-05-13 at 9 53 38 AM
The next figure is the difference at 4 days. Except for a handful of points the differences are on the order of 1.e-4 or less.
Screenshot 2024-05-13 at 9 53 56 AM

matsbn:
Great with stream capability for BLOM relaxation fields in NorESM! If I'm not mistaken, it seems the stream capability requires NUOPC and that the old relaxation procedure is unavailable with MCT. OMIP type experiments, that require SSS relaxation, with MCT is still used a lot for testing, so it would be good to find a way to still support those use cases.

mvertens:
@matsbn - thanks for pointing this out. The stream capability is actually available with the data functionality in mct. Its just that its must less flexible. I will add this in. However, as we move to 2.5 - mct will no longer exist so at some point we need to decide when blom master will no longer support it.

mvertens:
Just to add - testing the mct stream capability will need to be done in the noresm2.3 code base and not the noresm2.5 code base.

mvertens:
@matsbn @TomasTorsvik - after some more prototyping I've concluding that your original suggestion today of using the original interpolation for mct was actually easier. The problem is that additional domain files need to be specified for mct - and that makes things a lot more complicated. I've introduce a new namelist use_stream_relaxation that is set to .true. for nuopc and .false. for mct by default - and the code triggers off of this namelist and is therefore backwards compatible with mct but allows new stream usage for nuopc. I've pushed that back to this PR - and verified that using this branch in noresm2_3_develop for SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel works.

matsbn (approved):
I think that is a good and pragmatic choice for MCT, thanks!

PR #345, opened 2024-05-15
point to the correct location for stream sss relaxation

mvertens:
In the latest PR to master - the new stream relaxation file was pointing to my own directory rather than inputdata. This PR fixes that.


TomasTorsvik (approved):
@mvertens - thanks for catching this

PR #346, opened 2024-05-22
Add missing configurations for the ATRC option

TomasTorsvik:
add missing configurations for the ATRC option

ATRC is used as compiler option in BLOM, but it is not defined.

add ATRC option in the namelist building script and in the XML files.

This PR duplicates the essence of #341, but use the updated `master` branch as a starting point. This PR differs from #341 only with respect to changes that have been introduced for build config files, specifically `buildnml` and `buildcpp` replacing functionality that was previously in `buildlib`.

This PR technically allows activation of the ATRC CPP flag. I don't know if this option is up to date.

If this functionality is needed for CMIP6-type simulations, I suggest to start from the `v1.4.0` tag and make an update on the `NorESM2.0.7` release branch.

### DIFF
--- a/cime_config/buildcpp
+++ b/cime_config/buildcpp
@@ -80,6 +80,7 @@ def buildcpp(case):
     ocn_grid = case.get_value("OCN_GRID")
     turbclo = case.get_value("BLOM_TURBULENT_CLOSURE")
     tracers = case.get_value("BLOM_TRACER_MODULES")
+    blom_atrc = case.get_value("BLOM_ATRC")
     blom_unit = case.get_value("BLOM_UNIT")
     pio_typename = case.get_value("PIO_TYPENAME", subgroup="OCN")
 
@@ -124,6 +125,9 @@ def buildcpp(case):
                 blom_cppdefs = blom_cppdefs + " -DHAMOCC"
             else:
                 expect(False, "tracer module {} is not recognized".format(module))
+
+    if blom_atrc:
+        blom_cppdefs = blom_cppdefs + " -DATRC"
 
     if blom_unit == "mks":
         blom_cppdefs = blom_cppdefs + " -DMKS"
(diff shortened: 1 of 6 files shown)


JorgSchwinger (approved):
Looks fine to me. I don't know if everything related to this option is functional but that doesn't prevent us from implementing a switch to activate it.

TomasTorsvik:
Setting the ATRC option to TRUE will not change anything since in `mod_tracers.F90` we have the setting:
```
   ! Number of age tracers.
   integer, parameter :: natr = 0
```

I did a test where I set `natr = 1` in the source code, which results in a segmentation fault at run time in `blom_init`, so apparently something is either missing or not up to date for the ATRC functionality. I probably won't do any more testing on this at the moment, unless there is an urgent need to have this working for NorESM2.3.

YanchunHe:
> Setting the ATRC option to TRUE will not change anything since in `mod_tracers.F90` we have the setting:
> 
> ```
>    ! Number of age tracers.
>    integer, parameter :: natr = 0
> ```
> 
> I did a test where I set `natr = 1` in the source code, which results in a segmentation fault at run time in `blom_init`, so apparently something is either missing or not up to date for the ATRC functionality. I probably won't do any more testing on this at the moment, unless there is an urgent need to have this working for NorESM2.3.

Thank you, Tomas!

`natr` in `mod_param.F90` should be set together with `ntrocn`.  `ntrocn` should be a minimum value of double of the `natr`.

This should work.  But I can do a test later.

TomasTorsvik:
@YanchunHe - I realized that `ntrocn` needed to at least double `natr` after making the last comment. When setting `natr = 1` and `ntrocn = 2` the model runs without errors.

TomasTorsvik:
Regression tests against NorESM:BLOM tags `noresm2.1.1:dev1.5.1.3` show bit-identical output.

PR #347, opened 2024-05-24
compilation bugfix for nuopc cap

mvertens:
The latest commits to master break the compilation of the nuopc cap. It would be great if regression testing could be carried out on these updates to BLOM since this is the development code moving forwards.

### DIFF
--- a/drivers/nuopc/ocn_comp_nuopc.F90
+++ b/drivers/nuopc/ocn_comp_nuopc.F90
@@ -38,9 +38,8 @@ module ocn_comp_nuopc
    use nuopc_shr_methods, only : ChkErr, set_component_logging, &
                                  get_component_instance, state_setscalar, &
                                  alarmInit
-   use shr_file_mod, only: shr_file_getUnit, shr_file_getLogUnit, &
-                           shr_file_setLogUnit
    use shr_cal_mod, only : shr_cal_ymd2date
+   use shr_file_mod, only: shr_file_getUnit, shr_file_getLogUnit, shr_file_setLogUnit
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):

TomasTorsvik:
@mvertens - is there a branch/tag of NorESM or an Externals_???.cfg file that defines latest development version of NorESM2.5? I agree we should include a test with nuopc cap as this is the direction we will be heading. How will we be able to keep track on what is the latest development version of NorESM?

JorgSchwinger:
Shouldn't we, just after the release for NorESM2.3 is done, switch over to nuopc? 

If 2.5 is going to be based on nuopc, I don't see any particular reason to keep running with mct?

mvertens:
@TomasTorsvik - very good point. I am in the process of putting this external together.  I'll have something in place by early next week and send out a note.

jmaerz:
@JorgSchwinger , be aware that some -optional- parts of NorESM will still require work to enable the usage of nuopc... - but technically, I agree, would be then the time to do the move towards nuopc.

TomasTorsvik:
> Shouldn't we, just after the release for NorESM2.3 is done, switch over to nuopc?
> 
> If 2.5 is going to be based on nuopc, I don't see any particular reason to keep running with mct?

In general, yes, I agree. But there might be some interest in running NorESM2.3 for scientific purposes as this will likely be the last version of NorESM with CMIP6-type model components.

mvertens:
Just as an addition note - the regression tests for BLOM will only include DATM, DROF, BLOM and CICE6. They will not include the other NorESM components. Those will be tested in both their 'stand-alone' configurations as well as fully coupled configurations.

PR #348, opened 2024-05-27
remove unnecesarry unit conversion g<->kg for tracer field in thermf_cesm.F

JorgSchwinger:
In the calculation of virtual fluxes in `thermf_cesm` there is a conversion of units for the tracer field (back and forth), which has no function, is physically meaningless, and therefore confusing. This PR removes this unit conversion.

The results are not bit-for-bit, but in the range that is expected from numerical round-off (figure below show difference for the average of the 2nd year for a 2deg NOINYOC compset)

Surface DIC
Screenshot from 2024-05-27 07-51-13

surface pCO2 (note the output has been stored with scale-factor and offset, so even "numerical noise" can result in smooth differences if one of those changes - max in this figure is 5e-12 mu atm)
Screenshot from 2024-05-27 08-06-14

### DIFF
--- a/cesm/thermf_cesm.F
+++ b/cesm/thermf_cesm.F
@@ -196,7 +196,7 @@ c
             endif
 #    endif
 #  endif
-            trflx(nt,i,j)=-trc(i,j,k1n,nt)*fwflx*g2kg
+            trflx(nt,i,j)=-trc(i,j,k1n,nt)*fwflx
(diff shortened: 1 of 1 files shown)


matsbn (approved):
Looks good and great to clean this up.

jmaerz (approved):
Hi @JorgSchwinger , thanks for taking care of this! Approved.

PR #360, opened 2024-06-27
new mechanism for configuring noresm to send dms and bromo from ocn -> atm

mvertens:
This is a simple change to the blom buildnml that will enable CAM to determine if DMS, BROMO (and upcoming NH3 and N2O) and fluxes are being sent from the ocean and need to be received by CAM.
The logic is the following:
-  CMEPS will introduce new logical xml variables DMS_EMIS_OCN and BRF_EMIS_OCN - by default these will be FALSE. 
-  BLOM buildnml will determine if these variables should be set to TRUE - and reset them in the $CASEROOT.
-  CMEPS will generate new namelist variables flds_dms and flds_brf based on the setting of the these xml variables. These are treated by the system as ESMF config variables that all components have access to.
- CAM will check on the status of the presence and status of flds_dms and flds_brf and advertise that these are import fields, etc.
This change should be backwards compatible with MCT - since if the xml variables are not in CASEROOT,  buildnml will simply not do anythign.

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -322,6 +322,17 @@ def buildnml(case, caseroot, compname):
                     fwrite.write(line)
     shutil.move(namelist_file_temp, namelist_file)
 
+    # change xml variable settings if appropriate
+    if "ecosys" in case.get_value("BLOM_TRACER_MODULES"):
+        dms_emis = case.get_value("DMS_EMIS")
+        if dms_emis is not None or dms_emis is False:
+            case.set_value('DMS_EMIS','TRUE')
+            print ("   ***Resetting DMS_EMIS to TRUE***")
+        brf_emis = pg_blom.get_value('use_bromo')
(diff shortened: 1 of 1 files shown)


jmaerz:
Thanks for detailing this out!

TomasTorsvik:
@mvertens - Thanks! This looks fine to me.
@jmaerz ,  @JorgSchwinger - perhaps we also want this option in noresm2.3?

TomasTorsvik (approved):

JorgSchwinger:
NorESM 2.3 is MCT based, right? In which case this does not do anything according to Mariana. If that is correct we don't need it for 2.3

mvertens:
@JorgSchwinger - you are correct. This does nothing for 2.3 - but it should be backwards compatible.

TomasTorsvik:
I will merge this and create a new dev tag.

PR #366, opened 2024-07-09
Fix single column compilation issue after BLOM restructuring

jmaerz:
@mvertens and @JorgSchwinger: The restructuring in BLOM introduced an issue for compilation in single column mode without OpenMPI. Some un-used parameters removed. This is fixed in this PR. Please give it a quick review.


JorgSchwinger (approved):

mvertens:
@jmaerz - thanks for the fix.

PR #367, opened 2024-07-09
Hamocc kwrbioz off - enable remin and primary prod throughout full water column

### LINKED ISSUE #365: Enable primary production beyond current `kwrbioz` and enable POC remin above
As stated in #340 , with this issue, we aim at removing the separation of processes of primary production and POC remineralization by `kwrbioz`. In a first approach, we provide a switch which makes it possible to switch off this separation.

jmaerz:
This is a first draft PR to introduce the possibility to perform the processes of remineralization and primary production throughout the whole water column (#365). I essentially switch off any remineralization-related processes in the primary production-oriented loop and leave those to the second loop, if `lkwrbioz_off=true`, while keeping phytoplankton and zooplankton mortality in the primary production loop.

There are remaining issues with the current version:
- How to best define export production and related values? I currently would stick to the definitions as they are - likely making only a small error when it comes to the vertical resolution, since most production likely takes place as before in the euphotic zone - only the remineralization part will differ (and will influence primary production) 

Once the version is tested, one could consider removing the flexibility completely and use the current setting (`lkwrbioz_off=true`) as default by deleting unnecessary code parts.

Closes #365

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3653,6 +3653,16 @@
     <desc>Switch for cyano-bluefix in euphotic zone only</desc>
   </entry>
 
+  <entry id="lkwrbioz_off">
+    <type>logical</type>
+    <category>bgcnml</category>
+    <group>bgcnml</group>
+    <values>
+      <value>.true.</value>
+    </values>
+    <desc>Switch for primary production and remineralization throughout the whole water column</desc>
+  </entry>
+
   <entry id="do_oalk">
     <type>logical</type>
     <category>bgcnml</category>
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3658,7 +3658,7 @@
     <category>bgcnml</category>
     <group>bgcnml</group>
     <values>
-      <value>.true.</value>
+      <value>.false.</value>
     </values>
     <desc>Switch for primary production and remineralization throughout the whole water column</desc>
   </entry>
--- a/hamocc/mo_control_bgc.F90
+++ b/hamocc/mo_control_bgc.F90
@@ -61,7 +61,8 @@ module mo_control_bgc
   logical           :: do_oalk                = .false. ! apply ocean alkalinization
   logical           :: with_dmsph             = .false. ! apply DMS with pH dependence
   logical           :: use_M4AGO              = .false. ! run with M4AGO settling scheme
-  logical           :: leuphotic_cya          = .true. ! allow cyanobacteria to grow only in euphotic zone
+  logical           :: leuphotic_cya          = .true.  ! allow cyanobacteria to grow only in euphotic zone
+  logical           :: lkwrbioz_off           = .true.  ! if true, allow remin and primary prod throughout full water column
   integer           :: sedspin_yr_s           = -1      ! start year for sediment spin-up
   integer           :: sedspin_yr_e           = -1      ! end   year for sediment spin-up
   integer           :: sedspin_ncyc           = -1      ! sediment spin-up sub-cycles
--- a/hamocc/mo_control_bgc.F90
+++ b/hamocc/mo_control_bgc.F90
@@ -62,7 +62,7 @@ module mo_control_bgc
   logical           :: with_dmsph             = .false. ! apply DMS with pH dependence
   logical           :: use_M4AGO              = .false. ! run with M4AGO settling scheme
   logical           :: leuphotic_cya          = .true.  ! allow cyanobacteria to grow only in euphotic zone
-  logical           :: lkwrbioz_off           = .true.  ! if true, allow remin and primary prod throughout full water column
+  logical           :: lkwrbioz_off           = .false. ! if true, allow remin and primary prod throughout full water column
   integer           :: sedspin_yr_s           = -1      ! start year for sediment spin-up
   integer           :: sedspin_yr_e           = -1      ! end   year for sediment spin-up
   integer           :: sedspin_ncyc           = -1      ! sediment spin-up sub-cycles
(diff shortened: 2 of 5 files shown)


JorgSchwinger (approved):
Looks good to me. 

Should be tested thoroughly that (a) lkwrbioz_off=false is bit-for-bit, and (b) lkwrbioz=true shows an expected behavior.

I'd suggest to remove the switch as soon as this has be tested with reasonable results and make "one-loop" configuration the standard.

jmaerz:
Hi @JorgSchwinger , testing is in progress. Wrt removing the switch: you suggest to first perform the current PR and then deleting stuff in a second step after the current PR has merged?

JorgSchwinger:
Yes, that's what I meant.

TomasTorsvik (approved):

TomasTorsvik:
Looks fine to me.
I will be unavailable for further code review until the end of July.

jmaerz:
Sorry for being a bit slow on these things these days... - I did a 10 year run (that I guess, I should continue to one full century) and after 10 years, integrated primary production is boostet up to about 58GtC/yr with an export production of about 8.6GtC/yr - let's see where that goes with a longer run time. I still need to check wrt master which I will perform hereafter.

jmaerz:
Dear @mvertens , I am running into issues, when trying setup a new baseline run. What I performed:
```
# Basefolder for the two different model versions:
CHKFLD="chk-kwrbioz_off"

BASENAME="NorESM-master"
CHKNAME="NorESM-kwrbioz"

#TESTSUITE="--xml-category  aux_blom_noresm"
TESTSUITE="SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel"

PROJECT="nn2980k"
#BASELINEROOT="/cluster/shared/noresm/noresm_baselines"
BASELINEROOT="/cluster/work/users/jma105/noresm/regression_testing/"
BASELINENAME="baseline_${TESTSUITE}" 
ROOTFLD="/cluster/projects/${PROJECT}/${USER}/${CHKFLD}"
BASEROOT="${ROOTFLD}/${BASENAME}"
CHKROOT="${ROOTFLD}/${CHKNAME}"

# Creating the new baseline:
$BASEROOT/cime/scripts/create_test ${TESTSUITE}                     \
                                   --xml-machine betzy              \
                                   --xml-compiler intel             \
                                   --generate ${BASELINENAME}       \
                                   --baseline-root ${BASELINEROOT}  \
                                   --project ${PROJECT}

```
this runs into the error:
```
Test 'SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel' failed in phase 'SETUP' with exception 'ERROR: Fatal error in case.cmpgen_namelists: Setting resource.RLIMIT_STACK to -1 from (8388608, -1)
....

/cluster/projects/nn2980k/jma105/chk-kwrbioz_off/NorESM-master/components/blom/cime_config/namelist_definition_blom.xml validates
        ERROR: No variable BRF_EMIS_OCN found in case'
```
The variable `BRF_EMIS_OCN` isn't present in the xml-namelist file.

JorgSchwinger:
> Sorry for being a bit slow on these things these days... - I did a 10 year run (that I guess, I should continue to one full century) and after 10 years, integrated primary production is boostet up to about 58GtC/yr with an export production of about 8.6GtC/yr - let's see where that goes with a longer run time. I still need to check wrt master which I will perform hereafter.

This sounds a bit suspicious to me ("normal" PP would be around 40 PgC, right?) so a quite substantial increase! - But yes, let's see where it goes.

jmaerz:
Well, PP is a rather uncertain value - ranging from about 40-60 PgC/yr - so the new value would be at the upper limit. Through remineralization you have a substantial gain of recycled nutrients, which fosters PP. I hope to get results on the longer term (100a) evolution by tmw.

jmaerz:
@mvertens , ok, for now, I add `and config['blom_coupling']=='full'` in l. 329 of `blom/cime_config/buildnml`:
```
   # TODO: add n2o_emis and nh3_emis logic
    if "ecosys" in case.get_value("BLOM_TRACER_MODULES")  and config['blom_coupling']=='full':
```
to circumnavigate the issue - switching off air-sea flux settings. However, I am not 100% sure, if it is the right thing to do on the long term. Any comment on this?

jmaerz:
I performed now the regression testing (after adding the little addition above to `buildnml` to switch off the coupler fields to the un-used atmosphere). It fails, as expected, on the namelist (NLCOMP) due to the additionally introduced switch, all other tests are passed wrt current `master` in combination with ` release-noresm2.1.1`. I now put the default value of `lkwrbioz_off` to .false. (so it remains the same with former `master`) and see how the 100a run develops (hoping that subsequent automatic re-submission also catches the excluded nodes).

jmaerz:
The image shows the pattern of depth-integrated primary production after 60a run time. 
kwrbioz_off-ppint

Unfortunately, the run crashed due to time limit issues (again - see #369) and only went until year 60. However, from the monitoring output, I can see that after -somewhat expected- 20 years, the primary production stabilizes at around 60PgC/yr. Export production (as of currently defined output `epc100`) increased up to about 10 PgC/yr, while the fluxes (`carflx0100`) out of the euphotic zone at 100m is about 3.9PgC/yr (which is comparably low to observations) of which about 0.5PgC/yr reach the seafloor - hence, the question about adequate output adjustments for `lkwrbioz_off=.true.` remains and we can discuss this after the summer break.  

As briefly discussed with @JorgSchwinger , I push the PR, while the default behavior remains as the `master` before (`lkwrbioz_off=.false.`). The optional `lkwrbioz_off=.true.` will likely require some re-tuning and further investigation & re-considering adequate output.

jmaerz:
Potential tuning parameters are the mortality of phytoplankton and zooplankton as well as its grazing pressure on PP + remineralization rates and sinking velocity in the euphotic zone.

PR #370, opened 2024-07-13
More transparent co2 gas-exchange formulation

JorgSchwinger:
This PR abandons the use of a fit for CO2 solubility that already includes corrections for non-ideality and water vapor (Weiss and Price 1980). Instead these corrections are now calculated separately (also following Weiss and Price 1980). This is slightly more accurate (the correction for non-ideality was fixed at 1 atm pressure before). However, the main reason for introducing this is that diagnostic quantities can be calculated in a more easy and transparent way (e.g. the actual CO2 partial pressure at the air-sea interface, fugacity of CO2).

Results are not bit-for-bit, but very close (visually indistinguishable in time-series of a 100 year test-simulation). Note, however, that the output variable `pCO2` previously actually was the fugacity of CO2, not the partial pressure. This is now fixed and an extra output variable `fCO2` is introduced (the difference is a small offset of about 1 muatm)

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -5413,7 +5413,7 @@
       <value is_test="yes">4,2,2</value>
       <value is_test="yes" empty_hist="yes">0,0,0</value>
     </values>
-    <desc>Surface fugacity [uatm]</desc>
+    <desc>Surface CO2 fugacity [uatm]</desc>
   </entry>
 
   <entry id="srf_pco2">
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -5428,6 +5428,18 @@
     <desc>Surface pCO2 [uatm]</desc>
   </entry>
 
+  <entry id="srf_xco2">
+    <type>integer(3)</type>
+    <category>diabgc</category>
+    <group>diabgc</group>
+    <values>
+      <value>0,2,2</value>
+      <value is_test="yes">4,2,2</value>
+      <value is_test="yes" empty_hist="yes">0,0,0</value>
+    </values>
+    <desc>CO2 dry air mixing ratio that would be in equilibrium with the surface ocean [ppm]</desc>
+  </entry>
+
   <entry id="srf_pco2_gex">
     <type>integer(3)</type>
     <category>diabgc</category>
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -5413,7 +5413,7 @@
       <value is_test="yes">4,2,2</value>
       <value is_test="yes" empty_hist="yes">0,0,0</value>
     </values>
-    <desc>Surface fugacity at 1 atm total pressure [uatm]</desc>
+    <desc>Surface fugacity [uatm]</desc>
   </entry>
 
   <entry id="srf_pco2">
@@ -5425,7 +5425,7 @@
       <value is_test="yes">4,2,2</value>
       <value is_test="yes" empty_hist="yes">0,0,0</value>
     </values>
-    <desc>Surface pCO2 at 1 atm total pressure [uatm]</desc>
+    <desc>Surface pCO2 [uatm]</desc>
   </entry>
 
   <entry id="srf_pco2_gex">
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -5380,7 +5380,7 @@
     <desc>Piston velocity (kwco2)  [m s-1]</desc>
   </entry>
 
-  <entry id="srf_kwco2khm">
+  <entry id="srf_kwco2sol">
     <type>integer(3)</type>
     <category>diabgc</category>
     <group>diabgc</group>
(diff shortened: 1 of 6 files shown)


hamocc/mo_carchm.F90
```diff
@@ -384,6 +387,11 @@ subroutine carchm(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,prho,pglat,omask,psicomo
                 Kh_nh3  = (1.-psicomo(i,j)) * Kh_nh3
               endif
 
+
+              ! -----------------------------------------------------------------
+              ! Calculatte and apply surface fluxes
```
jmaerz: Calculate - no coffee involved ;-)

hamocc/mo_carchm.F90
```diff
@@ -256,13 +259,11 @@ subroutine carchm(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,prho,pglat,omask,psicomo
...
             if (k.eq.1) then
-              ! Determine CO2 pressure and fugacity (in micoatm)
-              ! NOTE: equation below for pCO2 needs requires CO2 in mol/kg
-              pco2 = cu * 1.e6 / Kh
-              if (use_natDIC) then
-                natpco2 = natcu * 1.e6 / Kh
-              endif
 
+              ! -----------------------------------------------------------------
+              ! Calculatte Schmidt numbers, solubilities, and piston velocities
```
jmaerz: Calculate

cime_config/namelist_definition_blom.xml
```diff
@@ -5301,7 +5301,7 @@
       <value is_test="yes">4,2,2</value>
       <value is_test="yes" empty_hist="yes">0,0,0</value>
     </values>
-    <desc>CO2 solubility under moist air assumption (kh) [mol kg-1 atm-1] </desc>
+    <desc>Surface fugacity [uatm]</desc>
```
jmaerz: CO2 fugacity(?)

hamocc/mo_carchm.F90
```diff
@@ -310,6 +311,7 @@ subroutine carchm(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,prho,pglat,omask,psicomo
                 ! Schmidt number water phase
                 sch_nh3_w  = mu_w   /(diff_nh3_w * rrho * 1000.)
               endif
+
```
jmaerz, line 315: Here and below (l 354): not sure, if you added or deleted whitespaces.

hamocc/mo_carchm.F90
```diff
@@ -399,43 +407,56 @@ subroutine carchm(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,prho,pglat,omask,psicomo
                 atnh3  = atm(i,j,iatmnh3)
               endif
 
-              ! Ratio P/P_0, where P is the local SLP and P_0 is standard pressure (1 atm). This is
-              ! used in all surface flux calculations where atmospheric concentration is given as a
-              ! mixing ratio (i.e. partial presure = mixing ratio*SLP/P_0 [atm])
-              rpp0 = ppao(i,j)/101325.0
+              ! Sea level pressure in atm. This is used in all surface flux calculations where 
+              ! atmospheric concentration is given as a mixing ratio (i.e. partial presure = mixing ratio*SLP/P_0 [atm])
```
jmaerz: partial pre**ss**ure

hamocc/mo_carchm.F90
```diff
@@ -399,43 +407,56 @@ subroutine carchm(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,prho,pglat,omask,psicomo
...
+              pH2O  = exp(24.4543 - 67.4509*(100.0/tk) - 4.8489*log(tk/100.0) - 0.000544*s)
+
+              ! Calculate the CO2 concentration in equilibrium with atmospheric x' (atco2=mole fraction of CO2 in dry air [ppm])
+              cu_sat = Kh0*atco2*1e-6*(rpp0-pH2O)*fc
+
+              fluxd = cu_sat*kwco2*dtbgc*rrho ! cu_sat and cu are in mol/kg. Multiply by rrho (g/cm^3) 
+              fluxu = cu    *kwco2*dtbgc*rrho ! to get fluxes in kmol/m^2
+
+              ! Set limit for CO2 outgassing to avoid negative DIC concentration, set minimum DIC concentration to 1e-5 kmol/m3
+              fluxu = min(fluxu,fluxd-(1e-5 - ocetra(i,j,k,isco212))*pddpo(i,j,1))
```
jmaerz, line 430: Here and below: Not sure, but it could be worth considering  parameter that defines the minimum value (could be added in `mo_param_bgc.F90`)

jmaerz (approved):
Hi @JorgSchwinger , looks good to me, while I didn't check each individual line wrt units, etc. - I trust that you checked it carefully. I hope it is CMIP7 future proof. Consider to inform Yanchun about output changes (for cmorization scripts) - if applicable.

hamocc/mo_carchm.F90
JorgSchwinger: I think I added a blank line for better readability

hamocc/mo_carchm.F90
JorgSchwinger: I agree, but I'll leave this for later...

PR #372, opened 2024-07-30
Corrected some bugs when using MCT

matsbn:
This PR fixes an issue that correct log file was not used when running NorESM simulations using MCT. It also omits the new logic for sending DMS and BROMO to CMEPS if NUOPC is not used, since this caused an configuration error with MCT.


mvertens (approved):
Thanks for fixing this.

PR #374, opened 2024-08-12
Fixes having BLOM run with newer version of CMEPS and fixes threading compilation

### LINKED ISSUE #373: BLOM does not compile when threading is turned on
In the ERP system test - threading is turned on in the compilation even if the number of threads is 1. In this case, BLOM was not able to compile successfully. The culprit  was mod_nctools.F90. This was with BLOM%ECO. Is anyone running BLOM with the number of threads > 1. Will this work?

mvertens:
Fixes the ability to run BLOM when PIO_NETCDF_FORMAT=64bit_data (the current default).
Fixes #373 for the compilation issue.

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -757,6 +757,7 @@
     <values>
       <value>0</value>
       <value pio_netcdf_format_ocn="64bit_offset">1</value>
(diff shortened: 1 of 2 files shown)


matsbn (approved):
Looks fine!

TomasTorsvik (approved):
Looks good to me

PR #375, opened 2024-08-16
Feature hybrid enhancements 5

matsbn:

- Optimization of the neutral diffusion algorithm (fewer computations involving the vertical polynomial reconstruction, avoid multiple extraction of polynomial coefficients of the vertical reconstruction, avoid insignificant neutral sublayers, try to keep temporary arrays in stack instead of heap memory). Seems to give a 10% speedup using BLOM%ECO, tnx1v4 and 354 processors.
- Further modified the transition between isopycnic and constant pressure interfaces for hybrid vertical coordinate and introduced lateral interface smoothing in the transition zone. This modification has been much used the last few months in testing and further development of the hybrid vertical coordinate, but a PR has been postponed due to the recent code refactorisation. Default is now to use this modification.
- Improved floating-point accuracy and robustness in segment integration for vertical remapping.


meson.build
```diff
@@ -14,6 +14,9 @@ if fcc.get_id() == 'gcc'
                            '-ffpe-trap=invalid,zero,overflow'],
                           language: 'fortran')
   endif
+  if get_option('buildtype') == 'release'
+    add_project_arguments('-ffp-contract=off'.split(),language: 'fortran')
+  endif
```
TomasTorsvik, line 19: I see from [gcc documentation](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html) that this option disables floating-point expression contraction.
- why do we want to use this option? Is it for performance or accuracy?
- why is it only enabled for buildtype == release?

TomasTorsvik (commented):
I'm sure this is fine, but I wonder about an option that you included in the meson.build file.

meson.build
matsbn: By using the "-ffp-contract=off" option, optimizations that alter the result of floating-point expressions, such as fused multiply-add (FMA), are omitted. This gives better reproducibility between different optimization levels. It is somewhat the same as the use of "-fp-model precise" for intel compilers. Some performance is lost, but errors discovered in optimized code can easier by debugged with no optimization and debug flags.

PR #377, opened 2024-08-19
add 05degree blom

### LINKED ISSUE #363: Compilation fails with current BLOM master
@jmaerz @TomasTorsvik @mvertens  When I try to compile NorESM release 2.1.1 with current BLOM master, I get an error:

The number of actual arguments cannot be greater than the number of dummy arg
uments.   [INI_PARAMBGC]
    call ini_parambgc(idm,jdm)

which is correct, since the routine in `hamocc_ini` is indeed called with these arguments, while `ini_parambgc` has no argumnents.

I am confused: Isn't the compilation tested in regression and continuous integration testing? Did I do something stupid when checking the model out?

### LINKED ISSUE #369: Runs are slow on betzy
Dear @mvertens , @JorgSchwinger , @TomasTorsvik ,
I am not entirely sure, whether this is the best place to ask (but I feel such things could be shared somewhere over a channel amongst sigma2 & NorESM users), but when I am running a simple ocean-only setup for a few years as I have done before, betzy seems to be extremely slow compared to before (as in before the module changes carried out by sigma2). It seems to take a time about 3x longer than usual to run a single year. Did you encounter this as well or am I am doing something wrong after the recent merges to `master`?

### LINKED ISSUE #373: BLOM does not compile when threading is turned on

mvertens:
- add a 1/2 degree blom grid - also will require CICE and ccs_config PRs  and is implemented in NorESM2.5 development branch (https://github.com/mvertens/NorESM/tree/feature/noresm2_5_alpha04_v3. 
- fix in nuopc cap for runid that has too short
- addition of new tests for short term archiving, 2 degree, intel-oneapi, etc
- simplification of  buildlib since master will now only support nuopc coupling infrastructure and cdeps
- ran new regression test successfully 

- NOTE: for HAMOCC need (**this should be done in another PR**) - fdepfile, swaclimfile, rivinfile, ndepfile


mvertens:
@matsbn @TomasTorsvik - I'd really like to have this as part of noresm2_5_alpha04 that I'm working on creating this week. Consequently, I thought we could move the HAMOCC capability for the half-degree blom grid to an upcoming PR and get this PR accepted first. I've implemented a new set of regression tests that more fully cover other new grids and archiving/resubmit capability as part of this PR.

TomasTorsvik:
> @matsbn @TomasTorsvik - I'd really like to have this as part of noresm2_5_alpha04 that I'm working on creating this week. Consequently, I thought we could move the HAMOCC capability for the half-degree blom grid to an upcoming PR and get this PR accepted first. I've implemented a new set of regression tests that more fully cover other new grids and archiving/resubmit capability as part of this PR.

I agree it makes sense to deal with HAMOCC capability separately, and not halt the progress on the physics side.

drivers/nuopc/ocn_comp_nuopc.F90
```diff
@@ -401,7 +401,7 @@ subroutine InitializeAdvertise(gcomp, importState, exportState, clock, rc)
          if (ChkErr(rc, __LINE__, u_FILE_u)) return
          read(cvalue,*) nthrds
       endif
-!$    call omp_set_num_threads(nthrds)
+      !$    call omp_set_num_threads(nthrds)
```
TomasTorsvik, line 404: For all other OMP calls, the prefix `!$omp <something openmp>` is used. I find it strange that this is not required here. Is this documented somewhere?

TomasTorsvik (approved):
I don't get all the details, but overall it looks fine to me.

I have one question regarding OMP, but it may not be an issue, and it's not specific to 0.5 degree grid.

drivers/nuopc/ocn_comp_nuopc.F90
mvertens: Good question. This is not being used at this point. I'll raise it as an issue since as you mention it does not effect this PR.

matsbn:
Thanks for this @mvertens !

With the changes to buildlib*, will BLOM still work with an MCT based NorESM?

The changes in https://github.com/NorESMhub/BLOM/pull/375 are rather significant for the application of the hybrid coordinate, so we should consider if that should be part of a noresm2_5_alpha04 release.

mvertens:
@matsbn - yes - BLOM will still work with an MCT based NorESM - but my understanding (based on talking with @TomasTorsvik ) is that master will no longer be used in noresm2.1.x and noresm2.3.x developments - there are now separate branches for this development. Is this not the case? If this is the case - we should consider removing the mct code from master moving forwards.

TomasTorsvik:
@matsbn , @mvertens 
We have branches `release-1.4` -> `noresm2.0`, `release-1.5` -> `noresm2.1`. We also have a tag `v1.6.1` -> `noresm2_3_develop`, but we haven't made a `release-1.6` branch yet. Technically, we could decide to abandon MCT on `master` and just branch off `v1.6.1` if further MCT updates are needed.

I think we should discuss this in a BLOM-core meeting before making a decision.

matsbn:
It is still very useful in ongoing projects to be able to swap in current master in a MCT based NorESM case. We don't really have separate branches fulfilling this purpose to my understanding. The "release-*" branches potentially could, but was intended for final testing before release and later maintenance, bug fixes or documentation enhancements but not answer-changing. A great discussion point for a BLOM-core meeting!

matsbn:
I tried the PR with a NorESM2.0.6 setup and get a buildlib failure. Since I have ongoing projects and numerous development tests on this old NorESM base that could still benefit greatly from master compatibility, this PR should ideally not break it.

mvertens:
@matsbn - sorry about that. I had assumed that we no longer needed the 2.0.6 functionality in master. Its easy for me to add it back in.

mvertens:
@matsbn - I've back out the changes and this should work for you now.

matsbn:
It builds now, but some of the modified compsets does not work with NorESM2. I think we should find a way to keep the old compsets in parallel for some time. Maybe the old compsets could be renamed to start with "N2", e.g NOIIAOC20TR -> N2OIIAOC20TR. For this specific compset, also directly creating a case with the complete old compset string "20TR_DATM%JRA_SLND_CICE%NORESM-CMIP6_BLOM%ECO_DROF%JRA_SGLC_SWAV" fails with a shr_stream_findBounds error. I wonder if the removed RUN_STARTDATE modification is the reason.

mvertens:
@matsbn - I'll fix that as well.

mvertens:
@matsbn - I've pushed fixes to the compsets back that you suggested. Hopefully this all works now.

matsbn (approved):
Works well now (also with ancient NorESM version :) Many thanks!

PR #381, opened 2024-08-21
Climatology stream checksums only when requested

matsbn:
Ensured that checksums of climatology streams are only computed and output when checksum diagnostics are requested. Obtaining the checksums can have a substantial performance penalty.


mvertens (approved):
Thanks for this.

PR #385, opened 2024-08-22
Improved backward compatibility of restart

matsbn:
The additional restart fields introduced in https://github.com/NorESMhub/BLOM/pull/371 are now only written and read if needed.


mvertens (approved):
Thank you so much for this!

PR #391, opened 2024-09-06
More efficient PE layout for NOINYOC compset

JorgSchwinger:
The current PE-layout for the NOINYOC compset on the tnx1v4 grid is very inefficient (16.56 simulated_years/day with hybrid coordinates on betzy). The new layout proposed here runs 27 years/day using the SAME number of nodes. This was achieved by increasing the PE-count for BLOM to 431 (new patch file created) and reducing the CICE PE-count to 80. Ice still runs much faster with the reduced PE-count. Results are bit-for-bit despite the changed number of CICE PEs.


TomasTorsvik (approved):
@JorgSchwinger - thanks, looks fine to me.

Would it make sense to introduce the same PE layout for the `a%TL319.+oi%tnx1v4` configuration?

```
  <grid name="a%TL319.+oi%tnx1v4">
    <mach name="any">
      <pes pesize="M" compset="_DATM.*_BLOM%ECO">
        <comment>none</comment>
        <ntasks>
          <ntasks_atm>128</ntasks_atm>
          <ntasks_rof>128</ntasks_rof>
          <ntasks_ice>128</ntasks_ice>
          <ntasks_ocn>354</ntasks_ocn>
          <ntasks_cpl>128</ntasks_cpl>
          <ntasks_lnd>128</ntasks_lnd>
          <ntasks_glc>128</ntasks_glc>
          <ntasks_wav>128</ntasks_wav>
        </ntasks>
```

JorgSchwinger:
This would definitely make sense, yes. I have not tested it yet though. So I was hesitant to modify it without testing. I suggest I will do this in a separate PR later.

mvertens (approved):

PR #392, opened 2024-09-09
Change M4AGO tag to dev-1.0.1

TomasTorsvik:
Update `release-1.6` branch with latest M4AGo tag.

### DIFF
--- a/Externals_BLOM.cfg
+++ b/Externals_BLOM.cfg
@@ -6,7 +6,7 @@ local_path = pkgs/CVMix-src
 required = True
 
 [M4AGO]
-tag = dev-1.0.0
+tag = dev-1.0.1
 protocol = git
 repo_url = https://github.com/jmaerz/M4AGO-sinking-scheme
 local_path = pkgs/M4AGO-sinking-scheme


jmaerz (approved):
Perfectly fine for me. I hope it resolves https://github.com/NorESMhub/NorESM/issues/556#issuecomment-2337746717 , while I am not sure about the wished cime default behavior.

TomasTorsvik:
Thanks, it solves the problem for now.

It looks like there is a general issue with referencing tags in submodules that requires an additional `git checkout <tag>` if it's not the last commit on a branch (see e.g. https://www.geeksforgeeks.org/how-to-specify-a-branchtag-when-adding-a-git-submodule/). Don't know how this is handled internally in `manage_externals`.

I need to make a similar PR to the `master` branch as well.

jmaerz:
i @TomasTorsvik , the current BLOM version in NorESM2.3 devel points to the dev-1.0.1 version, which is why the checkout resulted in a  conflict (which is why I only updated to that version) - technically, there is a more recent commit that added some more README info.

Issue #340, opened 2024-05-07
Plans for BLOM/iHAMOCC for NorESM2.5


TomasTorsvik:
Tentative plans for default (and optional) settings for BLOM/iHAMOCC towards NorESM2.5
See also #438 

- We will keep the MCT support until the technical release of NorESM2.5 (https://github.com/NorESMhub/BLOM/issues/439) 

- cgs units support will phase out - likely keep unit support for the restart files reading to enable restart from former restart files

- Plans for further development have been moved to a new overview issue: #462 
 

### Planning for release-1.7

Tasks under this heading can go into a new `release-1.7`, and will be carried forward into NorESM2.5.

**BLOM-specific**
- [x] BLOM code refactorization
- [x] hybrid coordinates (default) (isopycnal optional)
- [x] Add 2 degree grid resolution
- [x] Add 1/2 degree grid resolution
- [x] Fix some blom tnx2v1 namelist parameters #463 
- [x] support git fleximod #465 and #466 

**iHAMOCC-specific**
- [x] (potentially) making output files more cdo-compatible #355
- [x] Potentially remove separation of processes into euphotic and non-euphotic zone processes - refer to #365 and #367 
- [x] Make preformed tracers optional - to enable switching them off for HR runs - done via #394
- [x] Remove the option to run cyanobacteria (bluefix) in the MLD (default remains to run them in the euphotic zone) - see #395 and #396
- [x] bugfix sediment extended N-cycle #413 
- [x] dust flux to sediment output #420
- [x] Unify N-deposition reading - done with #410 (for nuopc) - see #397 and #398 for some further discussion. For post- BLOM tag v1.6.x simulations, nuopc is expected to handle the input, while for tag v1.6.x simulations with the extended nitrogen cycle switched on, a limited set of input files are generated by @jmaerz (request in case of interest)
- [x] Make i) T-dep. and ii) O2-dep. of remineralization available to default model code via thus far optional switch- see #400 and #424 , where the first step with a switch is taken  for the water column - sediment thus far missing (only realized for the extended N cycle) - #454 
- [x] #449
- [x] New tag for M4AGO  #460
- [x] Update POC remin rate in sediment #461 and #464

jmaerz:
**iHAMOCC-specific**

- [ ] Cold start of preformed (silicate) tracer  - handle all preformed tracers equally

TomasTorsvik:
This should maybe go into the milestone description at some point. Just edit the first entry if you have more things to add.

[...]

Issue #356, opened 2024-06-18
Initialization of preformed tracers


jmaerz:
Hi @JorgSchwinger and @TomasTorsvik ,
I was just checking, which data products would be available for initializing the preformed tracers. One reference that I found is:
Carter et al. 2020:  *Preformed Properties for Marine Organic Matter and Carbonate Mineral Cycling Quantification.* https://agupubs.onlinelibrary.wiley.com/doi/10.1029/2020GB006623 - however, it is lacking preformed DIC. It seems, however, that one can back-calculate preformed DIC from the other available tracers via Eq. (5a) in Li et al. 2000:  *Remineralization Ratios in the Subtropical North Pacific Gyre*. One thing in general is the potentially deviating stoichiometry used for deriving the preformed tracer fields. In case you agree, I would fill the mixed layer data with GLODAPv2 data (on which the product is based) which then could serve as initialization (maybe we also should consider moving to WOA18 or GLODAPv2 where possible for the CMIP7 cycle). Thoughts, opinions?

Once done, the restart part for preformed silicate can be cleaned up.

TomasTorsvik:
Hi @jmaerz ,
I don't have any insights to offer, but it seems reasonable to use GLODAPv2 or WOA18. I see there is also WOA23 available now https://www.ncei.noaa.gov/news/World-Ocean-Atlas-2023-v2 , mainly extending the datasets to 2022. GLODAPv3 is expected end of 2025, so that is probably too late for CMIP7 fast track.

JorgSchwinger:
Regarding WOA version, I would wait until input4MIPs is being updated. Currently we use the OMIP-recommended input-files based on WOA13. I suggest we update this when new input4MIPs data is published.

Regarding the preformed, I didn't have the chance to read through the articles (on a conference right now), but I share the concern about differing stociometric ratios.

jmaerz:
I am not sure, if one can correct in some way for the stoichiometry (ongoing discussion). Another option would then be spinning up the preformed tracers in i.e. a 2 degree ocean and interpolating to a 1 degree ocean...

PR #394, opened 2024-09-09
Making preformed tracers optional to enable HR runs without them

jmaerz:
@JorgSchwinger and @TomasTorsvik , as suggested in #340 , preformed tracers could be made optional to enable HR runs without them. With this PR, this is achieved, while the default is still to run iHAMOCC with preformed tracers.

@matsbn , the new introduced switch `use_pref_tracers` via `xmlchange HAMOCC_PREF_TRACERS=FALSE` might be of interest for you, in the case that you want to run HR simulations including iHAMOCC, while wanting to reduce the BGC costs.

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -31,7 +31,7 @@
     <default_value>FALSE</default_value>
     <group>build_component_blom</group>
     <file>env_build.xml</file>
-    <desc>Set preprocessor option to activate age tracer code. Requires module iage or ecosys</desc>
+    <desc>Set namelist option to activate age tracer code. Requires module iage or ecosys</desc>
   </entry>
 
   <entry id="BLOM_TURBULENT_CLOSURE">
@@ -151,7 +151,7 @@
     </values>
     <group>run_component_blom</group>
     <file>env_run.xml</file>
-    <desc>Set preprocessor option to activate the extended nitrogen cycle code. Requires module ecosys</desc>
+    <desc>Set namelist option to activate the extended nitrogen cycle code. Requires module ecosys</desc>
   </entry>
 
   <entry id="HAMOCC_ATMNDEPC">
@@ -184,7 +184,7 @@
     <default_value>FALSE</default_value>
     <group>run_component_blom</group>
     <file>env_run.xml</file>
-    <desc>Set preprocessor option to activate the M4AGO sinking scheme. Requires module ecosys</desc>
+    <desc>Set namelist option to activate the M4AGO sinking scheme. Requires module ecosys</desc>
   </entry>
 
   <entry id="HAMOCC_PREF_TRACERS">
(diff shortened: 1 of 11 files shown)


cime_config/config_component.xml
```diff
@@ -187,6 +187,15 @@
     <desc>Set preprocessor option to activate the M4AGO sinking scheme. Requires module ecosys</desc>
   </entry>
 
+  <entry id="HAMOCC_PREF_TRACERS">
+    <type>logical</type>
+    <valid_values>TRUE,FALSE</valid_values>
+    <default_value>TRUE</default_value>
+    <group>run_component_blom</group>
+    <file>env_run.xml</file>
+    <desc>Set preprocessor option to activate the preformed tracer code. Requires module ecosys</desc>
```
JorgSchwinger: I suggest replacing "preprocessor" by "namelist"

hamocc/mo_aufr_bgc.F90
```diff
@@ -400,13 +400,15 @@ subroutine aufr_bgc(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,kplyear,kplmon,kplday,o
...
-    if(lread_prefsi) then
-      call read_netcdf_var(ncid,'prefsilica',locetra(1,1,1,iprefsilica),2*kpke,0,iotype)
+    if (use_pref_tracers) then
+      call read_netcdf_var(ncid,'prefo2',locetra(1,1,1,iprefo2),2*kpke,0,iotype)
+      call read_netcdf_var(ncid,'prefpo4',locetra(1,1,1,iprefpo4),2*kpke,0,iotype)
+      call read_netcdf_var(ncid,'prefalk',locetra(1,1,1,iprefalk),2*kpke,0,iotype)
+      call read_netcdf_var(ncid,'prefdic',locetra(1,1,1,iprefdic),2*kpke,0,iotype)
+      if(lread_prefsi) then
+        call read_netcdf_var(ncid,'prefsilica',locetra(1,1,1,iprefsilica),2*kpke,0,iotype)
+      endif
```
JorgSchwinger, line 411: Is there a reason why the preformed silicate is treated differently from the other preformed tracers? Does it make sense to have the capability to restart with a preformed tracer when it is not in the restart file only for silicate but not for the others? I do not see so many use cases for this so maybe we can just omit the `lread_prefsi` related code?

JorgSchwinger (commented):
Looks good to me, although I'm not sure the special treatment for preformed silicate during restart is necessary. See my comment in the code

hamocc/mo_aufr_bgc.F90
jmaerz: Hi Jörg, since preformed silicate wasn't in the restart files before, start-up from restart files was an issue, hence this little quirky treatment that was put in to enable running the model from a restart file without it. Once we settle with #356 or have enough spun-up restart files with preformed silicate, this won't be an issue any longer and it can be removed. Hence, I would keep this treatment for now. We can put this clean-up task into #340 (and #356).

cime_config/config_component.xml
jmaerz: Done.

TomasTorsvik (approved):
@jmaerz - Thanks, looks fine from my side.

jmaerz:
@JorgSchwinger , any further comments or can I merge?

JorgSchwinger:
Please go ahead!

PR #401, opened 2024-09-13
fix initialization of sedfluxb

JorgSchwinger:


jmaerz (approved):
Hi @JorgSchwinger , great catch! - many thanks! - it is/was an open issue on my mindmap that something is still fishy with the written out burial fluxes - this catch hopefully resolves it. Surprised that the model didn't complain about it.

JorgSchwinger:
This was @tjiputra's  catch. FYI, Jerry discovered that C-isotopes don't work (produce inf after some years) when sediment spinup is switched on. It's difficult to trace since there is no sediment output for C-isotopes.

jmaerz:
Sorry for having caused that trouble.

Issue #354, opened 2024-06-10
How to run BLOM in tagged version v1.6.0 close/identical to the CMIP6 circulation field?


jmaerz:
@matsbn ,@AleksiNummelin (and @TomasTorsvik , since it might be relevant for your scientific work as well), I have a short question: if one wants to run the current master (as of tag v1.6.0) and would want to get the circulation as close as possible to the circulation represented in CMIP6, I would need to use:
- `./xmlchange BLOM_UNIT=cgs `

and then resetting the eddy parametrization values (looking into both, description and changes made in PR #337) from:
```
EDANIS = .true.
EDDF2D = .true.
EDSPRS = .false.
RHSCTP = .true.
EGC = 3.0
EGMNDF = 1.
EGMXDF = 2500.
RHISCF = 0.5
```
to
```
EDANIS = .false.
EDDF2D = .false.
EDSPRS = .true.
RHSCTP = .false.
EGC = 0.85          ! for tnx1v4 and 0.5 for tnx2v1
EGMNDF = 100.e4     
EGMXDF = 1500.e4    ! for tnx1v4 and 1000.e4 for tnx2v1 
RHISCF = 0.
```
right? - or do I miss something and further things would need to be changed? - would this setting result in identical or only close to CMIP6 circulation? Once answered, it could be added to the v1.6.0 release note https://github.com/NorESMhub/BLOM/releases/tag/v1.6.0

A brief `user_nl_blom` snippet would be greatly appreciated. 

**For tested `user_nl_blom` and correct settings, see:** https://github.com/NorESMhub/BLOM/wiki/new-BLOM-with-CMIP6-settings 

and below: https://github.com/NorESMhub/BLOM/issues/354#issuecomment-2410651675

jmaerz:
@matsbn , I would be grateful, if you could have a look and give advice on how to run the 1.6.x version of BLOM as close as possible to the CMIP6 circulation. If I am not mistaken, @TomasTorsvik and his works would also benefit from this information.

TomasTorsvik:
@jmaerz , @matsbn - yes, I'm also interested in running with BLOM/iHAMOCC `v1.6.x` in a setup that resembles the CMIP6 circulation.

matsbn:
Sorry for my very late response to this. I do believe using ``xmlchange BLOM_UNIT=cgs`` and setting
```
EDANIS = .false.
EDDF2D = .false.
EDSPRS = .true.
RHSCTP = .false.
[...]

PR #269, opened 2023-08-09
Merge extended nitrogen cycle, M4AGO scheme and preformed Si tracer into master

jmaerz:
This merge request aims at bringing in i) the **extended nitrogen cycle** developments carried out in the **ESM2025 project**, ii) the preformed silicate tracer, iii) optionally bluefix (cyanobacteria) only in the euphotic zone, and  iv) the **M<sup>4</sup>AGO sinking scheme** (Maerz et al. 2020) as submodule into iHAMOCC/BLOM/NorESM into `master`. 

The merge is carried out to enable a better overview of potentially needed adaptations wrt e.g. the nuopc cap. Interactive coupling of the extended nitrogen cycle to the atmosphere is currently only available through the MCT coupler and only with a certain CAM branch (so non-default for the extended nitrogen cycle, while it is anyways only run optionally). 

**Note 1:** This is a branch still in (minor) progress (mainly concerns tuning and thus model parameter values). At current stage and in preparation for NorESM2.3, the extended nitrogen cycle and the M<sup>4</sup>AGO scheme are **optional** and can be switched on via `xmlchange`: 

```
./xmlchange HAMOCC_EXTNCYCLE=TRUE
./xmlchange HAMOCC_M4AGO=TRUE
./xmlchange HAMOCC_ATMNDEPC=FALSE
./xmlchange HAMOCC_N2OC=FALSE
```
where the last two are in place to enable the interactive coupling to the atmosphere (non-default - CAM branch under development). 

**Note 2:** The M<sup>4</sup>AGO submodule is providing the M<sup>4</sup>AGO sinking scheme as expected, while the source code is up to further changes. Note that the ocean biogeochemistry currently needs about 1-2 year(s) until the sinking scheme gives reasonable sinking velocities and fluxes, which is due to changes in POM inventories. 

**About the extended nitrogen cycle:**
The extended nitrogen cycle incorporates **two new tracers**: ammonium and nitrite. 
The extended nitrogen cycle features an **explicit representation of the following processes**:
- ammonification (detritus -> NH<sub>4</sub>)
- updated bluefix approach (N<sub>2</sub>-fixation) with NH<sub>4</sub> release, also adjusted for the new BLOM coordinates (euphotic zone only as option or as before)
- 2-step nitrification (NH<sub>4</sub> -> NO<sub>2</sub> and NO<sub>2</sub> -> NO<sub>3</sub>), with N<sub>2</sub>O release during the first step, both steps resulting in dark-ocean carbon fixation
[...]

PR #402, opened 2024-09-13
fix sedfluxb for release-1.6 branch

TomasTorsvik:
Same as PR #401 , but for release-1.6

### DIFF
--- a/hamocc/mo_carbch.F90
+++ b/hamocc/mo_carbch.F90
@@ -233,7 +233,7 @@ contains
     endif
     allocate (sedfluxb(kpie,kpje,nsedtra),stat=errstat)
     if(errstat.ne.0) stop 'not enough memory sedfluxb'
-    sedfluxo(:,:,:) = 0.0
+    sedfluxb(:,:,:) = 0.0
 
     if (mnproc.eq.1) then
       write(io_stdo_bgc,*)'Memory allocation for variable satn2o ...'


jmaerz (approved):
Hi @TomasTorsvik , thanks for providing this fix in the release branch. Any chance to cherry pick this bug fix on top of former release branches (i.e. BLOM version for NorESM2.1(?) that enables MCT-based, to large extend CMIP6-backward-compatible BLOM runs)?

TomasTorsvik:
> Hi @TomasTorsvik , thanks for providing this fix in the release branch. Any chance to cherry pick this bug fix on top of former release branches (i.e. BLOM version for NorESM2.1(?) that enables MCT-based, to large extend CMIP6-backward-compatible BLOM runs)?

@jmaerz - I'm happy to have this fix also for NorESM2.1, I didn't check if it was relevant for `release-1.5`. I guess we don't do bug fixes for NorESM2.0 (unless it was introduced there at some stage after v2.0.1). 

In this case I found it easier to just re-apply the code change. With cherry-pick I'm always a bit nervous about pulling in changes that were not intended, but in principle I don't have anything against it.

jmaerz:
I fear, it was introduced with the N-cycle #269 ( 916352e) - hence any release afterward (+ NorESM pointing to those releases)... I am really sorry for this hassle.

TomasTorsvik:
@jmaerz - No worries. I gather this doesn't affect `release-1.5` in that case?

jmaerz:
Hi @TomasTorsvik , trying to browse through the release history, yes, it would affect 1.6.0 onward. As far as I understand, changing the BLOM `user_nl` should enable running the model in a mode that is to large extend backwards compatible (apart from small bug fixes like the sediment sinking, sediment alkalinity, etc.) - trying to figure this out through #354 which means that the current fix should suffice then. Thanks again!

PR #403, opened 2024-09-16
update dms exchange and simplify file opens

mvertens:
**Summary:**

- Replace  `call shr_file_setLogUnit(shrlogunit)` with  ` call shr_log_setLogUnit(shrlogunit)`. This also required the follow change.
- Remove nfu as a module variables in mod_xc.F90 and replaces all open files as follows (where nfu is now a local variable). This is done extensively in the CESM code base now.
```F90
   open (unit=nfu, file =..... ) (old)
   open (newunit=nfu, file = ....) (new)
```
- Send dms to the mediator based on a new NUOPC configuration variable `flds_dms`. CAM will also query this configuration variable and therefore if dms is sent by BLOM it will now be received by CAM and used.
- Send bromo to the mediator based on a new NUOPC configuration variable `flds_bromo`.  Currently this is simply a place holder for when CAM will be ready to use it.

**Testing:** 

 `./create_test --xml-category aux_blom_noresm --xml-machine betzy --compare noresm2_5_alpha04_dev1.6.1.5 --baseline-root ~/baselines/blom_develop --generate noresm2_5_alpha06_dev1.x.x.x `
- All non-hamocc cases were bfb but hamocc cases differed. It would be good to generate new baselines when hamocc changes answers.
- When this PR is accepted it should be tagged and x.x.x above should be replace by the tag values.


phy/mod_xc.F90
```diff
@@ -92,7 +92,7 @@ module mod_xc
 
   ! line printer unit (stdout) and file unit with default values 6 and
   ! 12, respectively
```
TomasTorsvik:
```suggestion
  ! line printer unit (stdout) with default value 6
```

TomasTorsvik (approved):
@mvertens - looks good from hamocc point of view. Had a few hamocc updates lately that are answer changing.

mvertens:
@matsbn - I'd like to get this update into a new noresm alpha tag that we can use for the next set of simulations. This will require updates to CAM and CMEPS as well. Currently CAM does not use the DMS coming from BLOM and this PR will fix that.

matsbn (approved):
Looks good! Standalone seems to work with the CI test, but I have not tested myself with NorESM2.* and MCT. There we obtained the file unit with shr_file_getUnit(), but I assume there will not be any conflict.

TomasTorsvik:
@mvertens - I see @matsbn approved this PR. Do you plan to merge it yourself? If not, I can merge it now and make the new BLOM tag.

mvertens:
@TomasTorsvik - if you could merge it now and make a new blom tag - that would be great! Thanks!

TomasTorsvik:
@mvertens - I have created a new tag on the `master` branch - `dev1.6.1.6`

PR #406, opened 2024-09-26
Fix namelist creating for multiple instances

blcc:
Get correct NTASK_OCN in buildcpp, which will be NINST_OCN times when run in multiple instance.
Move pg_blom into instance loop to reset it in every instance.

### DIFF
--- a/cime_config/buildcpp
+++ b/cime_config/buildcpp
@@ -38,6 +38,8 @@ def create_dimmod(case):
     blom_vcoord = case.get_value("BLOM_VCOORD")
     ntasks_ocn = case.get_value("NTASKS_OCN")
     objroot = case.get_value("OBJROOT")
+    if case.get_value("NINST_OCN") > 1:
+        ntasks_ocn = case.get_value("NTASKS_PER_INST_OCN")
 
     gridconf_dir = os.path.join(comp_root_dir_ocn, "bld", ocn_grid)
     kdm_file = os.path.join(gridconf_dir, "kdm." + blom_vcoord)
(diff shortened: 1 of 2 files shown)


cime_config/buildcpp
```diff
@@ -38,6 +38,8 @@ def create_dimmod(case):
     blom_vcoord = case.get_value("BLOM_VCOORD")
     ntasks_ocn = case.get_value("NTASKS_OCN")
     objroot = case.get_value("OBJROOT")
+    if case.get_value("NINST_OCN") > 1:
+        ntasks_ocn = case.get_value("NTASKS_PER_INST_OCN")
```
TomasTorsvik: I would prefer to define `ntasks_ocn` once, not overwrite
```
if (case.get_value("NINST_OCN") > 1) then
   ntasks_ocn = case.get_value("NTASKS_PER_INST_OCN")
else
   ntasks_ocn = case.get_value("NTASKS_OCN")
end if
```

TomasTorsvik (approved):
@blcc - I think this looks fine, but suggest not to overwrite `ntasks_ocn`.
Can merge this one you have made the change (or explain why you want to keep the code as it is).

cime_config/buildcpp
blcc: You're right, it is more clear with if-else.
I'll send updated PR tomorrow.

blcc:
Hi Tomas,
I just found NTASKS_PER_INST_OCN output same result with NTASKS_OCN when only one instance.
So that I replace it.

TomasTorsvik:
@blcc - I'm not convinced this is a good solution, since users may expect that setting the "NTASKS_OCN" option will set the `ntasks_ocn` variable. We should not change functionality, especially in the CMIP6 version of the model. I think its OK to make `ntasks_ocn` dependant on "NINST_OCN", since I don't think anyone will change "NINST_OCN" without knowing what they are doing.

blcc:
It also make point.
I change it back.

TomasTorsvik:
@blcc - I merged this PR, and have created a new tag `v1.4.3` on the `release-1.4` branch, for reference when you update the  `Externals.cfg` file.

blcc:
Thank you very much.

PR #407, opened 2024-09-27
use only submodules for noresm checkout

mvertens:
This PR does several things.
1) In preparation for moving to git-fleximod (which uses simply an extended version of the .gitmodules file) Externals_BLOM.cfg was removed. manage_externals will still be used - but it will only look at the .gitmodules file.
2) It is preferable to use https in .gitmodules rather than ssh because ssh requires you to have a github account, ssh keys and knowledge of using them. https is used across the rest of NorESM and CESM - so this change will facilitate that.


TomasTorsvik (approved):
Looks fine to me.

matsbn (approved):
A couple of years ago it seemed GitHub was recommending to go away from https and using tokens or ssh, if I remember correctly. If https still works for all practical purposes, I have no objections.

TomasTorsvik:
You can not push changes to github over https (or maybe it requires 2FA identification?). As long as you only check out the code it should be fine. For code developers it is probably easier to use ssh  or tokens (although, I haven't really looked into the option of using tokens).

TomasTorsvik:
@mvertens , @matsbn - I'm taking the liberty to merge this, and create a tag on the `master` branch, `dev1.6.1.7` will be available shortly.

mvertens:
@TomasTorsvik, @matsbn  - thank you for the quick response to this.

gold2718:
> You can not push changes to github over https (or maybe it requires 2FA identification?). As long as you only check out the code it should be fine. For code developers it is probably easier to use ssh or tokens (although, I haven't really looked into the option of using tokens).

You can push changes to GitHub over HTTPS. Quoting from the [NorESM2.1 documentation](https://github.com/NorESMhub/NorESM-docs/blob/noresm2.1/contribute/vc_github.rst#clone-a-repository)

> To upload data over HTTPS it is necessary to provide a personal access token (PAT) that can be created for a user account, which replaces the password for user authentication when pushing code to a GitHub repository.

Do we need to enhance this section?

TomasTorsvik:
My mistake. I just never took the time to look into the tokens option properly. The same for the github-cli option. Locally, I'm only using SSH. I have the impression that github changes the access options from time to time, so I think it's better to point to github documentation in this case.

jmaerz:
@gold2718 that means to copy github-credentials (PAT) to the HPC system, right (or is there a different way that I am unaware of)? Thus far, I also use ssh agent to access the repos. Maybe worth an edit into the docu (after testing), it seems to be easy to exchange default `https@` with `git@` without changing the repo using the global `.gitconfig`:
```
[url "git@github.com:"]
  insteadOf = https://github.com/
```
(untested thus far, though - just found it, when the question came up on my side).

PR #408, opened 2024-09-30
Add glc runoff to blom from cmeps

mvertens:
Currently cmeps sends to blom the sum of runoff from both lnd and glc. However, cmeps has the flexibility to send separate values if blom advertises that it wants to receive the separate values and then blom is responsible for summing thes fluxes.
There are outstanding issues with negative runoff from the glc component and as a result it would be very helpful for diagnostic purposes to have separate fluxes sent from the mediator to the ocean. In particular, @mpetrini-norce would find this a valuable addition to diagnosing this problem.

This is targeted for going into noresm2_5_alpha06. New baselines have to be generated since noresm2_5_alpha06 will have a new cice6 update which changes answers. Testing is currently started and will be updated here.

### DIFF
--- a/drivers/nuopc/ocn_comp_nuopc.F90
+++ b/drivers/nuopc/ocn_comp_nuopc.F90
@@ -66,7 +66,6 @@ module ocn_comp_nuopc
 
    private
 
-   integer, parameter :: cslen = 80  ! Short character string length.
    integer, parameter :: cllen = 265 ! Long character string length.
    character(len=*), parameter :: modname = '(ocn_comp_nuopc)'
    character(len=*), parameter :: u_FILE_u = &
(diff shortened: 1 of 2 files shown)


TomasTorsvik (approved):
Looks fine to me

matsbn (approved):
This looks fine. I then assume that the the mediator takes care of not including GLC runoff in fields associated with Foxx_rof[li] when Forr_rof[li]_glc are requested (I guess they must have been included before).

mvertens:
@mats - 

>  I then assume that the the mediator takes care of not including GLC runoff in fields associated with Foxx_rof[li] when Forr_rof[li]_glc are requested (I guess they must have been included before).

That is correct. The mediator determines what the ocean model advertises as expecting and then decides if its going to send one summed up field or two separate fields.

TomasTorsvik:
@mvertens , @matsbn - I will merge this PR and make a new tag. The new tag is `dev1.6.1.8`.

mvertens:
@TomasTorsvik  - thanks! I'm doing the regression testing now. I have created a new baseline - noresm2_5_alpha06_dev1.6.1.6 that includes all of the new alpha06 externals (particularly cice6). I'm creating a new baseline noresm2_5_alpha06_dev1.6.1.7.

PR #410, opened 2024-10-02
always obtain ndep from datm or cam using cmeps/cdeps

mvertens:
In noresm2_5, ndep is always sent to blom by either cam or datm - so there is no need to read a file.
This PR implements the functionality in blom to use this import data rather than reading a file.
- a new namelist variable (`use_nuopc_ndep`) has been added to the namelist group `config_bgc`
- by default use_nuopc_ndep will always be true if the coupling framework is using nuopc
- if use_nuopc_ndep is true, then hamocc will bypass reading in ndep from a file and instead use import values.

----

**Test 1:  ERS_Ld3.ne30pg3_tn14.N1850.betzy_intel.allactive-defaultio**
The left hand screen is the dev1.6.1.8 output for ndep whereas the right hand screen is the new output for ndep
The two have different behavior. The difference is due to the differerent ndep forcing files:

- use_nuopc_ndep = .false.  blom reads
/cluster/shared/noresm/inputdata/ocn/blom/bndcon/ndep_1850_CMIP6_tnx1v4_20171106.nc
- use_nuopc_ndep = .true.  blom receives from cam
/cluster/shared/noresm/inputdata/lnd/clm2/ndepdata/fndep_clm_WACCM6_CMIP6piControl001_y21-50avg_1850monthly_0.95x1.25_c180802.nc
- note that /cluster/shared/noresm/inputdata/ocn/blom/bndcon/ndep_1850_CMIP6_tnx1v4_20171106.nc
was obtained from an original fv1.9x2.5 dataset whereas cam is read in an fv09.x1.25 dataset.

Screenshot 2024-10-07 at 4 05 57 PM
 
----

**Test 2: ERS_Ld3.TL319_tn14.NOIIAJRAOC.betzy_intel.blom-hybrid**
- in the new case - datm is reading the following file **_and then interpolating it to the model grid_** (TL319) and then the **_mediator in turn interpolations this to the blom grid (tnx1v4)_**
`/cluster/shared/noresm/inputdata/lnd/clm2/ndepdata/fndep_clm_hist_b.e21.BWHIST.f09_g17.CMIP6-historical-WACCM.ensmean_1849-2015_monthly_0.9x1.25_c180926.nc`
- in the original case - blom is reading in the file 
`/cluster/shared/noresm/inputdata/ocn/blom/bndcon/ndep_1850_CMIP6_tnx1v4_20171106.nc`
But this is incorrect since the compset is for 2000 (see #411) 
- in the new case, using `xmlchange DATM_PRESNDEP=clim_1850` now results in the following two comparisons
Screenshot 2024-10-02 at 10 42 54 PM

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3839,6 +3839,18 @@
    <desc>File name (incl. full path) for atmopheric N-deposition data</desc>
   </entry>
 
+  <entry id="use_nuopc_ndep">
+    <type>logical</type>
+    <category>bgcnml</category>
+    <group>bgcnml</group>
+    <values>
+      <value>.false.</value>
+      <value comp_interface="nuopc">.true.</value>
+    </values>
+    <desc>if .true., use NUOPC to obtain NDEP (always imported to model) and ignore the
+    settings of do_ndep_coupled and ndepfile, whereas the setting of do_ndep is still used</desc>
+  </entry>
+
   <entry id="do_ndep_coupled" modify_via_xml="HAMOCC_ATMNDEPC">
     <type>logical</type>
     <category>bgcnml</category>
@@ -8760,17 +8772,6 @@
       <value comp_interface="nuopc">.true.</value>
     </values>
     <desc>if .true., use NUOPC stream relaxation capability</desc>
-  </entry>
-
-  <entry id="use_nuopc_ndep">
-    <type>logical</type>
-    <category>limits</category>
-    <group>limits</group>
-    <values>
-      <value>.false.</value>
-      <value comp_interface="nuopc">.true.</value>
-    </values>
-    <desc>if .true., use NUOPC to obtain NDEP (always imported to model) </desc>
   </entry>
 
   <!-- =========================  -->
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -892,6 +892,26 @@
     <desc>0 = netcdf, 1 = pnetcdf</desc>
   </entry>
 
+  <entry id="use_diag">
+    <type>logical</type>
+    <category>limits</category>
+    <group>limits</group>
+    <values>
+      <value>.false.</value>
+    </values>
+    <desc>optionally turn on additional diagnostics</desc>
+  </entry>
+
+  <entry id="use_arctic">
+    <type>logical</type>
+    <category>limits</category>
+    <group>limits</group>
+    <values>
+      <value>.true.</value>
+    </values>
+    <desc>if region includes arctic ocean</desc>
+  </entry>
+
   <!-- ========================= -->
   <!-- namelist group: vcoord -->
   <!-- ========================= -->
@@ -6040,7 +6060,7 @@
       <value is_test="yes">0,0,0</value>
       <value is_test="yes" empty_hist="yes">0,0,0</value>
     </values>
-    <desc>AMELIST FOR DIAGNOSTIC iHAMOCC OUTPUT</desc>
+    <desc>namelist for diagnostic iHAMOCC output</desc>
   </entry>
 
   <entry id="flx_car0100">
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3841,8 +3841,8 @@
 
   <entry id="use_nuopc_ndep">
     <type>logical</type>
-    <category>bgcnml</category>
-    <group>bgcnml</group>
+    <category>config_bgc</category>
+    <group>config_bgc</group>
     <values>
       <value>.false.</value>
       <value comp_interface="nuopc">.true.</value>
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -636,17 +636,6 @@
     </desc>
   </entry>
 
-  <entry id="use_stream_relaxation">
-    <type>logical</type>
-    <category>limits</category>
-    <group>limits</group>
-    <values>
-      <value>.false.</value>
-      <value comp_interface="nuopc">.true.</value>
-    </values>
-    <desc>if .true., use NUOPC stream relaxation capability</desc>
-  </entry>
-
   <entry id="trxday">
     <type>real</type>
     <category>limits</category>
@@ -8736,6 +8725,32 @@
       <value is_test="yes" empty_hist="yes">0,0,0</value>
     </values>
     <desc>(ssster) [mol m-3]</desc>
+  </entry>
+
+  <!-- =========================  -->
+  <!-- namelist group: forcing fields using nuopc -->
+  <!-- =========================  -->
+
+  <entry id="use_stream_relaxation">
+    <type>logical</type>
+    <category>limits</category>
+    <group>limits</group>
+    <values>
+      <value>.false.</value>
+      <value comp_interface="nuopc">.true.</value>
+    </values>
+    <desc>if .true., use NUOPC stream relaxation capability</desc>
+  </entry>
+
+  <entry id="use_nuopc_ndep">
+    <type>logical</type>
+    <category>limits</category>
+    <group>limits</group>
+    <values>
+      <value>.false.</value>
+      <value comp_interface="nuopc">.true.</value>
+    </values>
+    <desc>if .true., use NUOPC to obtain NDEP (always imported to model) </desc>
   </entry>
 
   <!-- =========================  -->
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3847,19 +3847,7 @@
       <value>.false.</value>
       <value comp_interface="nuopc">.true.</value>
     </values>
-    <desc>if .true., use NUOPC to obtain NDEP (always imported to model) and ignore the
-    settings of do_ndep_coupled and ndepfile, whereas the setting of do_ndep is still used</desc>
-  </entry>
-
-  <entry id="do_ndep_coupled" modify_via_xml="HAMOCC_ATMNDEPC">
-    <type>logical</type>
-    <category>bgcnml</category>
-    <group>bgcnml</group>
-    <values>
-      <value>.false.</value>
-      <value blom_n_deposition="yes" hamocc_atmndepc="yes" hamocc_extncycle="yes">.true.</value>
-    </values>
-    <desc>Switch to couple nitrogen deposition. Requires do_ndep.</desc>
+    <desc>if .true., use NUOPC to obtain NDEP (always imported to model).</desc>
   </entry>
 
   <entry id="do_n2onh3_coupled" modify_via_xml="HAMOCC_N2OC">
--- a/hamocc/mo_param1_bgc.F90
+++ b/hamocc/mo_param1_bgc.F90
@@ -242,7 +242,8 @@ contains
     use mo_control_bgc, only: bgc_namelist,get_bgc_namelist, io_stdo_bgc
     use mo_control_bgc, only: use_BROMO,use_AGG,use_WLIN,use_natDIC,use_CFC,use_cisonew,           &
                               use_sedbypass,use_PBGC_OCNP_TIMESTEP,use_PBGC_CK_TIMESTEP,           &
-                              use_FB_BGC_OCE, use_BOXATM,use_extNcycle,use_pref_tracers,use_nuopc_ndep
+                              use_FB_BGC_OCE, use_BOXATM,use_extNcycle,use_pref_tracers,           &
+                              use_nuopc_ndep
     integer :: iounit
 
     namelist / config_bgc / use_BROMO,use_AGG,use_WLIN,use_natDIC,use_CFC,use_cisonew,             &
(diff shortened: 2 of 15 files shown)


hamocc/mo_read_ndep.F90
```diff
@@ -226,17 +228,28 @@ subroutine get_ndep(kpie,kpje,kbnd,kplyear,kplmon,omask,ndep,patmnhxdep,patmnoyd
...
+        if (mnproc == 1) then
           write (io_stdo_bgc,*) 'iHAMOCC: getting NOy and NHx deposition from atm'
         endif
 
+        if (use_nuopc_ndep .and. .not. use_extNcycle) then
+          ! Reduced and oxidized forms will all enter the NO3 pool
+          !$omp parallel do private(i)
+          do  j=1,kpje
+            do  i=1,kpie
+              ndep(i,j,idepnoy) = ndep(i,j,idepnoy) + ndep(i,j,idepnhx)
```
jmaerz: I am a bit puzzled in this line: if `use_extNcycle=.false.`, `ndep` is of size `(..,..,1)` and not `(..,..,2)` - see `mo_param1_bgc`. `patmnoydep` and `patmnhxdep` hold the fluxes coming from the nuopc cap, right? - hence I would suggest:
```
ndep(i,j,idepnoy) = (patmnoydep(i,j)+patmnhxdep(i,j))*fatmndep
```

jmaerz: I I am not mistaken, this would also affect the former loop above.

cime_config/namelist_definition_blom.xml
```diff
@@ -903,6 +892,26 @@
...
+  <entry id="use_arctic">
+    <type>logical</type>
+    <category>limits</category>
+    <group>limits</group>
+    <values>
+      <value>.true.</value>
+    </values>
+    <desc>if region includes arctic ocean</desc>
+  </entry>
+
```
jmaerz, line 914: Question out of curiosity: what is the use_arctic switch for?

jmaerz:
In the figures above (2nd test case): Something seem to be odd - the pattern look quite different and it doesn't seem having to do with the displayed ranges.

hamocc/mo_read_ndep.F90
mvertens: @jmaerz - thanks for catching this. You are totally right. I agree with your suggested change.

cime_config/namelist_definition_blom.xml
mvertens: use_arctic has a different communication pattern in mod_xc.F90. 
In particular - it checks that all arctic patch tiles must be the same size or empty,
and empty tiles must be "twinned" across the top boundary. It ensures that the latitudinal tile dimension is closed/arctic and handles the halo in the arctic differently. @matsbn - do you want to add more?

hamocc/mo_read_ndep.F90
```diff
@@ -202,21 +202,24 @@ subroutine get_ndep(kpie,kpje,kbnd,kplyear,kplmon,omask,ndep,patmnhxdep,patmnoyd
...
 
     ! if N-deposition is switched off set ndep to zero and return
     if (.not. do_ndep) then
       ndep(:,:,:) = 0.0
       return
     endif
 
+    ! If  use_nuopc_ndep, nitrogen deposition is ALWAYS obtained from the
+    ! nuopc mediator
     if (use_extNcycle .and. do_ndep_coupled) then
```
jmaerz, line 217: Dear @mvertens , I am not fully getting your logics here. If I am not mistaken, the source of N-deposition (whether it comes from CAM or from input files) will be decided at the level of the nuopc, right - so the `do_ndep_coupled` should appear somewhere else? - currently, the extended nitrogen cycle wouldn't be supported by nupc without the coupled fields to CAM, if I am not mistaken. If so, I would rewrite the code as follows (sorry for the formatting - it's a bit hard on github to keep the overview in a small text box):
```
if (use_nupc_ndep) then
         ! get N-deposition from atmosphere (either CAM or data)
        ndep(:,:,:) = 0.
        fatmndep = 365.*86400./mw_nitrogen
       
   if (use_extNcycle) then
      !$omp parallel do private(i)
        do j=1,kpje
          do i=1,kpie
            ! convert from kgN/m2/s to climatological input file units: kmolN/m2/yr
            if (patmnoydep(i,j) > 0.) then
              ndep(i,j,idepnoy) = patmnoydep(i,j)*fatmndep
            endif
            if (patmnhxdep(i,j) > 0.) then
              ndep(i,j,idepnhx) = patmnhxdep(i,j)*fatmndep
            endif
          enddo
        enddo
        !$omp end parallel do
        if (mnproc == 1 .and. first_call) then
          write (io_stdo_bgc,*) 'iHAMOCC: getting NOy and NHx deposition from atm'
        endif
   else ! standard N-cycle version
      !$omp parallel do private(i)
        do j=1,kpje
          do i=1,kpie
            ! convert from kgN/m2/s to climatological input file units: kmolN/m2/yr
            ! reduced and oxidized forms will all enter the NO3 pool
            ndep(i,j,idepnoy) = (patmnoydep(i,j)+patmnhxdep(i,j))*fatmndep
          enddo
        enddo
        !$omp end parallel do
   end if
else ! MCT coupler part - don't using nuopc
 ! read ndep data from file
      if (kplmon /= oldmonth) then
        month_in_file=(max(startyear,min(endyear,kplyear))-startyear)*12+kplmon
        if (mnproc == 1) then
          write(io_stdo_bgc,*) 'Read N deposition month ',month_in_file,' from file ',trim(ndepfile)
        endif
        ncstat=nf90_open(trim(ndepfile),nf90_nowrite,ncid)
        if (use_extNcycle) then
          call read_netcdf_var(ncid,'nhxdep',nhxdepread,1,month_in_file,0)
          call read_netcdf_var(ncid,'noydep',noydepread,1,month_in_file,0)
        else
          call read_netcdf_var(ncid,'ndep',ndepread,1,month_in_file,0)
        endif
        ncstat=nf90_close(ncid)
        oldmonth=kplmon
      endif
      !$omp parallel do private(i)
      do j=1,kpje
        do i=1,kpie
             if (use_extNcycle) then
            ndep(i,j,idepnoy) = noydepread(i,j)
            ndep(i,j,idepnhx) = nhxdepread(i,j)
           else
            ndep(i,j,idepnoy) = ndepread(i,j)
          endif
        enddo
      enddo
      !$omp end parallel do

    endif
endif
```

mvertens:
@jmaerz - I think a chat at this point would be helpful. My logic was the following:
- You are correct that for nuopc - when regardless of if you are using the extended nitrogen cycle or not - you always get the ndep fields from cam. So you logic is correct. 
- In terms of do_ndep_coupled - that is not needed for nuopc but is needed for MCT if you are using the extended nitrogen cycle. The previous logic was ` if (use_extNcycle .and. do_ndep_coupled) then`. I don't see that case in your block. Do we no longer need that?
- I supposed that we could introduce another flag - get_ndep_from_atm which would contain the following logic:
`get_ndep_from_atm = ( use_nuopc_ndep .or.  (use_extNcycle .and. do_ndep_coupled))` 
and use this instead of `if (use_nuopc_ndep)`. Does that make sense?
- I mentioned to @matsbn this morning that supporting both nuopc and mct in blom at this point is getting quite complicated and that it would be good to separate out the branches. This would provide a lot of simplification to the blom code.

jmaerz:
Hi @mvertens , as a follow up on our call:
Here is the path to the nitrogen input data downloaded from esgf:
`/cluster/projects/nn2980k/jma105/input_restart/bnd/Ndeposition` and then there should be a folder with `preparation` under which the original files from the esgf node are located (sorry, just having issues logging in to betzy...). Ideally, these should be the input files for any component, as far as I can see/understand it.

jmaerz:
@mats and @JorgSchwinger , just to inform you, I just discussed with @mvertens  that for the optional extended nitrogen cycle, I don't see a need in `master` to support the online-coupling to CAM through the MCT coupler for nitrogen deposition (while deposition via input files is supported through both, nuopc and MCT - for a number of cases, the input files are currently not yet there for the extended nitrogen cycle, though). **For the current default iHAMOCC, nothing changes (while there might be small changes in the input files, which we currently try to sort out)**. In ESM2025, I will likely rely on NorESM2.3, where the MCT online-coupling was/is supported. Any run for CMIP7 will anyway rely on nuopc, where both, the current default and the extended nitrogen cycle are supported with both options.

mvertens:
@jmaerz - I think the code is behaving correctly at this point and the differences we are seeing are due to differences in the forcing datasets. I think this PR will ensure that CAM and CDEPS will determine what is sent for ndeps to BLOM and will be used consistently by both CTSM and BLOM. Are you okay with this PR - or do you still have concerns? I'm happy to run other tests.

mvertens:
@TomasTorsvik @matsbn @JorgSchwinger - could you please provide feedback on this PR? I'd like to ensure that we have consistency for the NDEP received by BLOM and CLM and have CAM be control this forcing.

hamocc/mo_param1_bgc.F90
```diff
@@ -242,12 +242,12 @@ subroutine init_indices()
     use mo_control_bgc, only: bgc_namelist,get_bgc_namelist, io_stdo_bgc
     use mo_control_bgc, only: use_BROMO,use_AGG,use_WLIN,use_natDIC,use_CFC,use_cisonew,           &
                               use_sedbypass,use_PBGC_OCNP_TIMESTEP,use_PBGC_CK_TIMESTEP,           &
-                              use_FB_BGC_OCE, use_BOXATM,use_extNcycle,use_pref_tracers
+                              use_FB_BGC_OCE, use_BOXATM,use_extNcycle,use_pref_tracers,use_nuopc_ndep
```
jmaerz: Could you add a line break here?

mvertens: Done.

jmaerz (approved):
Hi @mvertens , thanks for the work and thanks for the discussions on it. From my side, it looks ok now and can be merged.

TomasTorsvik (approved):
@mvertens - thanks! Look fine as far as I can see.
Forgot to add my review. Doing it now.

TomasTorsvik (approved):

TomasTorsvik:
@mvertens - I see you haven't assigned anyone for merging this PR. I can do it if you want.

jmaerz:
We discussed to wait for @JorgSchwinger 's approval/last check.

hamocc/mo_read_ndep.F90
```diff
@@ -226,17 +227,24 @@ subroutine get_ndep(kpie,kpje,kbnd,kplyear,kplmon,omask,ndep,patmnhxdep,patmnoyd
...
-        if (mnproc .eq. 1) then
-          write (io_stdo_bgc,*) 'iHAMOCC: getting NOy and NHx deposition from atm'
-        endif
+        !$omp end parallel do
+      else
+        !$omp parallel do private(i)
+        do j=1,kpje
+          do i=1,kpie
+            ! reduced and oxidized forms will all enter the NO3 pool
+            ndep(i,j,idepnoy) = (patmnoydep(i,j)+patmnhxdep(i,j))*fatmndep
```
JorgSchwinger, line 236: Why is here no if-block `patmnoydep(i,j) > 0. .and.  patmnhxdep(i,j) > 0.`? Either this is needed (then it should be here) or it isn't needed, but then it also could be removed in the block above for better readability and consistency?

phy/mod_ifdefs.F90
```diff
@@ -38,41 +38,13 @@ module mod_ifdefs
...
   logical :: use_MKS = .false.
 #endif
-#ifdef DIAG
-  logical :: use_DIAG = .true.
-#else
-  logical :: use_DIAG = .false.
-#endif
 
+  ! Namelist input
+  logical :: use_diag = .false.
```
JorgSchwinger, line 48: Is this the right place to define this variable? Where in BLOM are the other namelist variables defined?

JorgSchwinger (approved):
Looks good to me. I agree that it becomes complicate to support several options. So I would support to remove  the possibility to read ndep from file (inside HAMOCC) in NorESM2.5

hamocc/mo_read_ndep.F90
jmaerz: Hi @mvertens and @JorgSchwinger , the `if` statements above were introduced at some point due to how it was written formerly in `mo_hamocc4bcm` - not sure, if it was really needed there for the online coupling to the atmosphere (it was handled this way also for bromoforme and I simply oriented along it). If CAM has always values >= zero,. then it can be removed, otherwise, it would be good to introduce it here as well.

jmaerz:
Hi @mvertens , after the discussion with Jörg, I feel free to perform this PR (I hope that's ok with you). I will afterward introduce the small `if` in the reading of the N-deposition. If something else is coming up, we can work on it from the updated `master` branch again. Many thanks for all the work!

mvertens:
@jmaerz - thanks for merging this! I think your plan sounds good.
@jmaerz @JorgSchwinger @matsbn @TomasTorsvik  - 

>  I agree that it becomes complicate to support several options. So I would support to remove the possibility to read ndep from file (inside HAMOCC) in NorESM2.5

The simplest way to remove the possibility to read in ndep  for noresm2.5 is to only support nuopc capability in master. I think it will get more and more confusing to have all the older namelists that are needed only by mct and the new namelists needed by nuopc. This will be even more evident as I bring in changes for using streams to read in other forcing datasets. I'm not sure when we will be okay with splitting master off - but it would be good to discuss this in an upcoming meeting.

PR #412, opened 2024-10-15
replaced imp.load_source with importlib.machinery.SourceFileLoader

mvertens:
the module imp no will longer be supported in python3.12 and later. 
Currently we are using python3.11 in noresm to deal with this limitation. But the desire is to move to python 3.12 in the near future.
The suggested recommendation is to replace 
`imp.load_source` with  `importlib.machinery.SourceFileLoader`
Testing: verified that ERR_Ld3.TL319_tn14.NOIIAJRAOC.betzy_intel.blom-hybrid built successfully (when the same change was made in cice)

### DIFF
--- a/cime_config/buildlib
+++ b/cime_config/buildlib
@@ -3,7 +3,7 @@
 """
 build blom library
 """
-import os, shutil, sys, glob, imp
+import os, sys
 
 _CIMEROOT = os.environ.get("CIMEROOT")
 if _CIMEROOT is None:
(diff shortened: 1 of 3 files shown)


TomasTorsvik:
We get errors with the mpi builds in the `mod_xc.F90` file. Don't understand why this should break due to changes in cime_config, so maybe it's something in the test setup that goes wrong.

TomasTorsvik (approved):
Thanks, @mvertens , looks fine to me.

PR #414, opened 2024-10-18
Adjust pkgs meson file for transitioning M4AGO

jmaerz:
**This PR informs about potential issues when setting up a run with meson - cime-managed runs are NOT affected.**

The PR provides support for any of the current M4AGO versions provided meson >= 0.57 is used. 

**Background**
I recently worked on the M4AGO-sinking-scheme package and introduced a new file, which requires to be explicitly set via `meson.build`. Typically, a particular tag/sha of M4AGO is checked out (where this change isn't needed), but if one for any reason moves on to M4AGO `master`, this will create issues when compiling with meson. The code snippet is applicable for meson versions >=0.57, where the meson file system module has been introduced. Otherwise, the file needs to be manually added in the `meson.build` file. Hope it helps in the current transition period for M4AGO which is for making it more general and future-ready. 

@JorgSchwinger and @TomasTorsvik : if wished, I can push it to `master` - I was just hesitant since the required meson release was Feb. 2021, which isn't too long ago. 

@matsbn : I put you in here explicitly, since I know that you are using meson every now and then (not sure, which meson version you're running, though).

### DIFF
--- a/pkgs/meson.build
+++ b/pkgs/meson.build
@@ -1,3 +1,4 @@
+fs = import('fs') # import fs system module: https://mesonbuild.com/Fs-module.html
 sources += files('CVMix-src/src/shared/cvmix_background.F90',
 'CVMix-src/src/shared/cvmix_convection.F90', 
 'CVMix-src/src/shared/cvmix_ddiff.F90',
@@ -13,4 +14,7 @@ if get_option('ecosys')
    sources += files('M4AGO-sinking-scheme/src/mo_m4ago_core.f90',
                     'M4AGO-sinking-scheme/src/mo_m4ago_physics.f90',
        		    'M4AGO-sinking-scheme/src/mo_ihamocc4m4ago.f90')
+   if fs.exists('M4AGO-sinking-scheme/src/mo_m4ago_kind.F90')
+     sources += files('M4AGO-sinking-scheme/src/mo_m4ago_kind.F90')
+   endif
 endif


TomasTorsvik (approved):
@jmaerz - thanks, this looks fine to me
One question, is the testing on file existing intended as a transitional fix, or will this be permanent? I'm wondering, if we need to introduce such a test every time a file is added/removed/renamed in a submodule, or only until a new tag version is available?

jmaerz:
Hi @TomasTorsvik , to my understanding and according to the meson documentation, each `meson.build` file in the subfolders needs to include individual files explicitly to be able to compile and track changes (similar to former `make`) to only re-compile the changed dependencies. Hence, whenever a name has changed or a file deleted/added, one needs to adopt it. The checking for the existence of the file here is more of a transition phase (I introduced a `mo_m4ago_kind.F90` file to explicitly set the working precision). That said, the PR is more for information (but I can push it, if wished) and should just make the transitioning as smooth as possible for potential meson users. A normal checkout currently just gets the right version and the snippet wouldn't be needed - but the moment you checkout M4AGO `master` for some reason, meson build would fail without including the file. Again, this does not affect normal cime configurations and thus NorESM users (it is only about BLOM, e.g. in the 1D setup).

TomasTorsvik:
@jmaerz - thanks for the explanation. Based on what you say, I'm leaning towards keeping this as a draft PR for now. Once there is a new tag on M4AGO that we want to use as default, we can look at this again.

TomasTorsvik:
Hi @jmaerz , from our latest conversation I understand that including this in `master` will make your testing procedure easier. This PR shouldn't break anything (except parhaps for old `meson` versions) so I'm fine with bringing it in.

JorgSchwinger (approved):
I can't say anything about this - I have never used a non-CESM-coupled version of BLOM and I have no knowledge about how meson works.

Issue #416, opened 2024-10-23
1D test setup broken after PR #410


jmaerz:
Hi @mvertens , cc @TomasTorsvik and @JorgSchwinger ,
after your PR #410 that I pulled in, the 1D setup seem to be broken and runs fail with the error message below. I haven't looked deeper into it, but maybe you could have a look? - it seem to be coming from changes in BLOM, not in iHAMOCC. 

I did a run test with a MCT-setup on betzy and it didn't crash - so the issue seems to only affect runs in the 1D setup. 

```
init_timevars: number of baroclinic time steps per day:           24

 rdlim: resolved parameters:
 nday1:                                   0
 nday2:                                  90
 nyear ,nmonth ,nday ,nsec :     1990 ,01 ,01 ,00000
 nyear0,nmonth0,nday0,nsec0:     1990 ,01 ,01 ,00000
 barotr. steps per barocl. step:         60
 runid:                          SC_20210824
 rstfrq:                                 30
 rstmon:                                  T
 rstann:                                  F

 itest =    1, jtest =    1 found on processor     1

bigrid: nreg = 2
bigrid: global domain with arctic patch

stderr:
At line 4095 of file ../phy/mod_xc.F90
Fortran runtime error: Index '-3' of dimension 2 of array 'a' below lower bound of -2

Error termination. Backtrace:
#0  0x7f54ba8b3d11 in ???
#1  0x7f54ba8b4859 in ???
#2  0x7f54ba8b4ed6 in ???
#3  0x56103539a12f in __mod_xc_MOD_xctilr
	at ../phy/mod_xc.F90:4095
#4  0x56103521132a in __mod_bigrid_MOD_bigrid
	at ../phy/mod_bigrid.F90:126
#5  0x5610352d9c22 in __mod_inigeo_MOD_inigeo
	at ../phy/mod_inigeo.F90:123
#6  0x561035212b55 in __mod_blom_init_MOD_blom_init
	at ../phy/mod_blom_init.F90:116
#7  0x56103539c5f8 in blom
	at ../drivers/nocoupler/blom.F:38
#8  0x5610351f37c2 in main
	at ../drivers/nocoupler/blom.F:26

```

JorgSchwinger:
Could it be that there is a `#ifdef ARCTIC` left somewhere? Or that the new logical arctic-switch is not set correctly in this setup?

jmaerz:
[...]

PR #417, opened 2024-10-23
Introduce if statement in N-deposition also for default iHAMOCC

jmaerz:
This is only a minor PR that introduces the if statement as requested by @JorgSchwinger  in PR #410. I checked the current `master` on betzy, if it is still running despite issue #416 and it is still running (didn't check for bfb, though, since this was done in PR410).


TomasTorsvik (approved):
@jmaerz - great, thanks!

PR #418, opened 2024-10-23
Fix diagnostic writing for coupled N-deposition

jmaerz:
This PR fixes the writing of diagnostics for N-deposition.  It has been already fixed in `master` by PR #410.

### DIFF
--- a/cesm/mod_cesm.F90
+++ b/cesm/mod_cesm.F90
@@ -242,7 +242,7 @@ contains
       call ncdefvar('atmbrf_da', 'x y', ndouble, 8)
       call ncdefvar('atmn2o_da', 'x y', ndouble, 8)
       call ncdefvar('atmnh3_da', 'x y', ndouble, 8)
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):
@jmaerz - approved!
Is this all that you what you want to "cherry-pick" from #410 into the `release-1.6` branch?

jmaerz:
Hi @TomasTorsvik , yes, I think so - in NorESM2.3, MCT is still the default, if I am not mistaken?!

TomasTorsvik:
> Hi @TomasTorsvik , yes, I think so - in NorESM2.3, MCT is still the default, if I am not mistaken?!

Yes, that is correct.

jmaerz:
Then I am not seeing any reason to pick more than this change from #410. Or do you?

TomasTorsvik:
> Then I am not seeing any reason to pick more than this change from #410. Or do you?

No, I think it's fine.

PR #420, opened 2024-10-23
Dust fluxes output and add descriptions in the xml-namelist file

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , as discussed, this PR brings in dust output and some descriptions in the xml-namelist file. I tried to fill all iHAMOCC gaps to my best knowledge. I cleaned the way, M4AGO sinking velocity is handled in `mo_ocprod` (wasn't ideal before).

Together with this PR, I introduced the xmlchange switch `HAMOCC_SINKING_SCHEME` (and removed the compset HAMOCC_M4AGO switch) that allows to select among the four supported sinking schemes: [WLIN, AGG, M4AGO, CONST], the default remains WLIN.

Ideally, this PR also enters the v1.6.x release version. Sorry for this delayed PR for the release tag - it just adds some potentially valuable output (and provides a more descriptive iHAMOCC xml-namelist definition file).

### DIFF
--- a/hamocc/mo_bgcmean.F90
+++ b/hamocc/mo_bgcmean.F90
@@ -877,7 +877,7 @@ CONTAINS
       jdustflx2000(n)=i_bsc_m2d*min(1,FLX_DUST2000(n))
       if (FLX_DUST4000(n) > 0) i_bsc_m2d=i_bsc_m2d+1
       jdustflx4000(n)=i_bsc_m2d*min(1,FLX_DUST4000(n))
-      if (FLX_CAL_BOT(n) > 0) i_bsc_m2d=i_bsc_m2d+1
+      if (FLX_DUST_BOT(n) > 0) i_bsc_m2d=i_bsc_m2d+1
       jdustflx_bot(n)=i_bsc_m2d*min(1,FLX_DUST_BOT(n))
 
       if (.not. use_sedbypass) then
--- a/hamocc/mo_bgcmean.F90
+++ b/hamocc/mo_bgcmean.F90
@@ -121,6 +121,8 @@ module mo_bgcmean
        & FLX_BSI2000   =0    ,FLX_BSI4000   =0    ,FLX_BSI_BOT   =0    ,  &
        & FLX_CAL0100   =0    ,FLX_CAL0500   =0    ,FLX_CAL1000   =0    ,  &
        & FLX_CAL2000   =0    ,FLX_CAL4000   =0    ,FLX_CAL_BOT   =0    ,  &
+       & FLX_DUST0100  =0    ,FLX_DUST0500  =0    ,FLX_DUST1000  =0    ,  &
+       & FLX_DUST2000  =0    ,FLX_DUST4000  =0    ,FLX_DUST_BOT  =0    ,  &
        & FLX_SEDIFFIC  =0    ,FLX_SEDIFFAL  =0    ,FLX_SEDIFFPH  =0    ,  &
        & FLX_SEDIFFOX  =0    ,FLX_SEDIFFN2  =0    ,FLX_SEDIFFNO3 =0    ,  &
        & FLX_SEDIFFSI  =0    ,FLX_SEDIFFNH4 =0    ,FLX_SEDIFFN2O =0    ,  &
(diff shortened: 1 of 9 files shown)


hamocc/mo_bgcmean.F90
```diff
@@ -859,6 +867,19 @@ subroutine alloc_mem_bgcmean(kpie,kpje,kpke)
...
+      jdustflx0100(n)=i_bsc_m2d*min(1,FLX_DUST0100(n))
+      if (FLX_DUST0500(n) > 0) i_bsc_m2d=i_bsc_m2d+1
+      jdustflx0500(n)=i_bsc_m2d*min(1,FLX_DUST0500(n))
+      if (FLX_DUST1000(n) > 0) i_bsc_m2d=i_bsc_m2d+1
+      jdustflx1000(n)=i_bsc_m2d*min(1,FLX_DUST1000(n))
+      if (FLX_DUST2000(n) > 0) i_bsc_m2d=i_bsc_m2d+1
+      jdustflx2000(n)=i_bsc_m2d*min(1,FLX_DUST2000(n))
+      if (FLX_DUST4000(n) > 0) i_bsc_m2d=i_bsc_m2d+1
+      jdustflx4000(n)=i_bsc_m2d*min(1,FLX_DUST4000(n))
+      if (FLX_CAL_BOT(n) > 0) i_bsc_m2d=i_bsc_m2d+1
```
JorgSchwinger: This should be FLX_DUST_BOT(n)!

hamocc/mo_ncout_hamocc.F90
```diff
@@ -585,6 +594,12 @@ subroutine ncwrt_bgc(iogrp)
     call wrtsrf(jcalflx2000(iogrp),  FLX_CAL2000(iogrp),  rnacc*1e3/dtbgc,0.,cmpflg,'calflx2000')
     call wrtsrf(jcalflx4000(iogrp),  FLX_CAL4000(iogrp),  rnacc*1e3/dtbgc,0.,cmpflg,'calflx4000')
     call wrtsrf(jcalflx_bot(iogrp),  FLX_CAL_BOT(iogrp),  rnacc*1e3/dtbgc,0.,cmpflg,'calflx_bot')
+    call wrtsrf(jdustflx0100(iogrp),  FLX_DUST0100(iogrp),  rnacc*1e3/dtbgc,0.,cmpflg,'dustflx0100')
+    call wrtsrf(jdustflx0500(iogrp),  FLX_DUST0500(iogrp),  rnacc*1e3/dtbgc,0.,cmpflg,'dustflx0500')
+    call wrtsrf(jdustflx1000(iogrp),  FLX_DUST1000(iogrp),  rnacc*1e3/dtbgc,0.,cmpflg,'dustflx1000')
+    call wrtsrf(jdustflx2000(iogrp),  FLX_DUST2000(iogrp),  rnacc*1e3/dtbgc,0.,cmpflg,'dustflx2000')
+    call wrtsrf(jdustflx4000(iogrp),  FLX_DUST4000(iogrp),  rnacc*1e3/dtbgc,0.,cmpflg,'dustflx4000')
+    call wrtsrf(jdustflx_bot(iogrp),  FLX_DUST_BOT(iogrp),  rnacc*1e3/dtbgc,0.,cmpflg,'dustflx_bot')
```
JorgSchwinger, line 602: Did you check that the 1e3 is correct here? Usually this converts from kmol -> mol, is it here from kg -> g?

JorgSchwinger (approved):
Looks good to me

hamocc/mo_bgcmean.F90
jmaerz: Thanks for catching this!

hamocc/mo_ncout_hamocc.F90
jmaerz: Yes, it's from kg -> g.

TomasTorsvik:
Looks fine for me, but I will hold off on the review until the PR is updated with namelist option.

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , I tinkered around with the component settings and the xml-namelist and found a way to get to the wanted behavior by making changes to `builnml`, `config_component.xml` and `namelist_definition_blom.xml` - so no further fortran code changes were needed.

I introduced a new xmlchange switch named `HAMOCC_SINKING_SCHEME` (and removed the compset HAMOCC_M4AGO switch) that allows to select among the four supported schemes: `[WLIN,AGG,M4AGO,CONST]`, default is `WLIN`. Buildnml sets then the `use_XXX` to the respective values (very explicit, maybe better ways are available).

I hope that satisfies the needs that we have. Tomas an I, we discussed that we could introduce another small subroutine that allows to replace all the different places of `if...else if ...` statements in `mo_ocprod` for choosing the sinking velocities by a simple function/subroutine call. This could be done for the `master` afterward - just to make the whole sinking routine a bit cleaner. I feel that this PR is already big for playing it on the v1.6.xxx release version, which is why one could just do it for `master`, if wished for - it would ease the maintenance at least.

TomasTorsvik (approved):
@jmaerz - thanks! Looks good to me.

JorgSchwinger:
Looks fine, thank you.

PR #422, opened 2024-10-24
Remove redundant OMP call

### LINKED ISSUE #379: In ocn_comp_nuocp.F90, the prefix `!$omp <something openmp>` is used in a non-standard way.
For all other OMP calls, the prefix `!$omp <something openmp>` is used. I find it strange that this is not required here. Is this documented somewhere?

_Originally posted by @TomasTorsvik in https://github.com/NorESMhub/BLOM/pull/377#discussion_r1723052983_

TomasTorsvik:
This call to `omp_set_num_threads(nthrds)` has never been used and would appear to be redundant in the the NUOPC driver. I suggest to remove it.

Closes #379


TomasTorsvik:
@matsbn , @mvertens - any views on this?

mvertens (approved):

PR #423, opened 2024-10-24
Apply patch of latest commit in master for the dustflux output PR #420 to release1.6 branch

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , this is the PR to replay the dustflux output commit #420 to the the `release1.6` branch. I ended up using 
```
git show HEAD > dust.patch
git apply --reject --ignore-whitespace dust.patch
```
which resulted in a few rejected hunks of the patch that then needed to be inserted manually. A `cherrypick` didn't work out for some reason.

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -83,7 +83,7 @@ def buildnml(case, caseroot, compname):
     hamocc_extncycle          = case.get_value("HAMOCC_EXTNCYCLE")
     hamocc_n2oc               = case.get_value("HAMOCC_N2OC")
     hamocc_atmndepc           = case.get_value("HAMOCC_ATMNDEPC")
-    hamocc_m4ago              = case.get_value("HAMOCC_M4AGO")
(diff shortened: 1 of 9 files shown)


TomasTorsvik:
@jmaerz - Thanks! This looks fine to me.
I still don't understand why the cherry-pick option didn't work, but I suspect it has something to do with how git compares changes introduced in the `release-1.6` and `master` branches. The [git-cherry-pick](https://git-scm.com/docs/git-cherry-pick) documentation describes what happens when git is unable to apply changes, which is the case we have run into here.
As the `master` branch develops further, it will likely become increasingly difficult to rely on cherry-picking changes back to the `release-1.6` branch. Since fewer changes are introduced in `release-1.6`, I wonder if there may be less issues in applying patches from `release-1.6` to `master`. Maybe something to test, if we have changes in the future that we know will go into both branches.

TomasTorsvik (approved):

PR #424, opened 2024-10-25
Introduce switch dor T- and O2-dependent remineralization

### LINKED ISSUE #400: Step-wise introduction of T-dep. and O2-dep of remineralization?
Hi @JorgSchwinger and @TomasTorsvik , as we discussed to move on and enable T-dep. and O2-dependent remineralization (#340): shall we perform this for the default iHAMOCC version in a two-step process (i.e. providing switches, which may then later be removed)?

jmaerz:
With this PR, the option will be provided to enable T- and O2-dependent remineralization for all model versions (thus far, this was only for the extended nitrogen cycle and when M4AGO was switched on). It is the first step in the two step process towards making the T- and O2-dependency default (#400). 

The remineralization rate for POM `drempoc` might need adjustment, when switching on the T- and O2-dependent remineralization. I suspect that the rate as set, when `HAMOCC_SINKING_SCHEME=M4AGO` is switched on won't apply for the default sinking scheme `WLIN`. I suspect the rate for `WLIN` being of similar order as currently set (so one could give it a try without changing `drempoc`). 

**Thus far, this only concerns the water column, but I can also introduce it to the sediment (as done for the N-cycle)** - comments on this? I suspect it could be of interest to enable it for the sediment as well in the default model at least for the aerobic remineralization. 

I put this as PR, while I anticipate that there will be changes wrt the sediment, once I receive your comments.

closes #400

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3882,6 +3882,16 @@
     <desc>Switch for primary production and remineralization throughout the whole water column</desc>
   </entry>
 
+  <entry id="lTO2depremin">
+    <type>logical</type>
+    <category>bgcnml</category>
+    <group>bgcnml</group>
+    <values>
+      <value>.false.</value>
+    </values>
+    <desc>Switch for T- and O2-dependent remineralization of POM</desc>
+  </entry>
+
   <entry id="do_oalk">
     <type>logical</type>
     <category>bgcnml</category>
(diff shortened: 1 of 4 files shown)


JorgSchwinger (approved):
Looks good to me, thanks for implementing this.

Regarding the sediment - if it is not too much work, I would vote for introducing it to the sediments, too.

TomasTorsvik (approved):
Looks fine to me.
I prefer small, incremental PRs (easier for debugging), but if you are already looking at implementing this for the sediments, I'm fine with including that as well.

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , I push this PR now, while I envisage to bring in the T- & O2-dep also for the aerobic sediment remineralization (other processes upon further discussion). I will open another PR for the sediment, once it is ready (may take some time, while the implementation should be straight forward).

PR #425, opened 2024-10-25
Revert setting of use_ARCTIC variable

### LINKED ISSUE #416: 1D test setup broken after PR #410

TomasTorsvik:
This PR reverts the setting of the `use_arctic` variable back to being determined by a CPP flag.

The `use_arctic` variable is applied in the subroutine `xcspmd` from `mod_xc`, which is called in `blom_init`before `rdlim`. Hence, `use_arctic` is used before the namelist variables are read. It would be nice if we could remove the ARCTIC CPP flag, but it will have to be done in a different way.

Testing:
- This PR works for the stand-alone `single_column` test run.
- PASS for test SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel based on noresm2_5_alpha07

### DIFF
--- a/cime_config/buildcpp
+++ b/cime_config/buildcpp
@@ -89,6 +89,9 @@ def buildcpp(case):
     if pio_typename == "pnetcdf":
         blom_cppdefs = blom_cppdefs + " -DPNETCDF"
 
+    if ocn_grid in ["tnx2v1", "tnx1.5v1", "tnx1v1", "tnx1v3", "tnx1v4", "tnx0.5v1", "tnx0.25v1", "tnx0.25v3", "tnx0.25v4", "tnx0.125v4"]:
+        blom_cppdefs = blom_cppdefs + " -DARCTIC"
+
     if ocn_grid in ["gx1v5", "gx1v6", "tnx1v1", "tnx1v3", "tnx1v4", "tnx0.5v1", "tnx0.25v1", "tnx0.25v3", "tnx0.25v4", "tnx0.125v4"]:
         blom_cppdefs = blom_cppdefs + " -DLEVITUS2X"
 
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -902,16 +902,6 @@
     <desc>optionally turn on additional diagnostics</desc>
   </entry>
 
-  <entry id="use_arctic">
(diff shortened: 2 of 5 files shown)


mvertens (approved):
@TomasTorsvik - thanks for reverting this.

PR #429, opened 2024-11-01
workaround for upper/lower case issue for namelist entry lTO2depremin

### LINKED ISSUE #428: setting lTO2depremin namelist entry via user_nl_blom does not work
Setting the lTO2depremin namelist entry via user_nl_blom does not work, the build script stops with the following error:

`ocn_in_paramgen.OcnInParamGenError: Variable 'lto2depremin' not found in any namelist definition files. Please double-check 'user_nl_blom'.`

Apparently, the mixed upper and lower case writing of the namelist entry seems to be a problem, since the upper case letters are removed. Is this a known problem @mvertens ?

I changed the definition in `namelist_definition_blom.xml` to be all lowercase, which worked. So I propose to use this solution as a workaround (I will issue a PR in a moment), although I think that mixed upper/lowercase writing of namelist entries should be supported.

JorgSchwinger:
This PR provides a workaround to use the newly introduced namelist variable `lTO2depremin` by renaming it to `lto2depremin` (closes #428). In general, I believe this problem should be fixed in cime (@mvertens)?

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3872,7 +3872,7 @@
     <desc>Switch for primary production and remineralization throughout the whole water column</desc>
   </entry>
 
-  <entry id="lTO2depremin">
+  <entry id="lto2depremin">
     <type>logical</type>
     <category>bgcnml</category>
     <group>bgcnml</group>


jmaerz (approved):
Hi Joerg, yes, unfortunately a know issue - I should have known it...

TomasTorsvik (approved):

PR #431, opened 2024-11-19
New compsets for JRA forcing and fix for PE layout

JorgSchwinger:
This PR introduces a few new compsets for JRA forcing. The JRA-repeat year forcing also requires updates to datm and drof configuration files, which I will bring later in in a seperate PR.

I also include a fix for the PE layout for the NOINYOC compset because the model was hanging (during startup) in release 2.1.3 (the PE layout was working under release 2.1.1). It might be a coincidence, but changing the order of CICE and DATM/DROF did the trick.

### DIFF
--- a/cime_config/config_compsets.xml
+++ b/cime_config/config_compsets.xml
@@ -92,8 +92,18 @@
   </compset>
   <compset>
     <!-- not valid for noresm2.5 -->
-    <alias>NOIIAJRAOC</alias>
+    <alias>N2NOIIAJRAOC</alias>
     <lname>2000_DATM%JRA_SLND_CICE%NORESM-CMIP6_BLOM%ECO_DROF%JRA_SGLC_SWAV</lname>
+  </compset>
+
+  <compset>
+    <alias>NOIIAJRAOC1850</alias>
+    <lname>1850_DATM%JRA_SLND_CICE_BLOM%ECO_DROF%JRA_SGLC_SWAV</lname>
+  </compset>
+  <compset>
+    <!-- not valid for noresm2.5 -->
+    <alias>N2NOIIAJRAOC1850</alias>
+    <lname>1850_DATM%JRA_SLND_CICE%NORESM-CMIP6_BLOM%ECO_DROF%JRA_SGLC_SWAV</lname>
   </compset>
 
   <compset>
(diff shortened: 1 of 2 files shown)


mvertens:
@JorgSchwinger - are the changes for datm and drof for the mct or nuopc data models. I am assuming mct? If so - Will these changes be needed for DATM and DROF for nuopc?  It would be helpful to put links here for the PRs that you are issuing for mct. 
@matsbn - I still believe that it would be helpful to set a time when we are no longer trying to support both mct and nuopc in the same code base.

matsbn:
I also assume this comes with one or more PRs for the data components since "JRARYF1961" will currently not be recognised by DATM/DROF for either MCT or NUOPC. In the current NorESM2.5 (NUOPC) datm and drof, the following repeat year forcings are supported for JRA:
```
    <desc option="JRA-RYF8485"> JRA55 Repeat Year Forcing v1.3 1984-1985</desc>
    <desc option="JRA-RYF9091"> JRA55 Repeat Year Forcing v1.3 1990-1991</desc>
    <desc option="JRA-RYF0304"> JRA55 Repeat Year Forcing v1.3 2003-2004</desc>
```
Maybe good to follow that naming convention for the data component options.

On a somewhat related note, specifying DATM%JRA with NUOPC gives interannual JRA forcing for years 1958-2016, which differs from the year range 1958-2018 when using MCT and NorESM2. I am right now testing to replace DATM%JRA with DATM%JRA-1p4-2018 (and similarly for DROF) when using NUOPC which should provide the same year range as used before. I see in the latest MOM-interface that they use DATM%JRA-1p5-2023 for interannual JRA forcing, so maybe we should consider to update the data components to have the more extended JRA forcing available.

@mvertens, yes we should try to set a deadline for when to only support NUOPC. Personally, although I'm trying to do most of my development and testing with NUOPC, I still find it very useful to have the latest developments available for some work that requires MCT based NorESM.

JorgSchwinger:
@mvertens @matsbn 

Since we need this for ongoing projects, I would (for now) really like to include this in the NorESM2.3 release. For cime, I have created a draft PR here:

https://github.com/NorESMhub/cime/pull/89

This should later go also into the NUOPC based version, where the configuration files look slightly different. I have noticed that there is already (several) JRA repeat-year forcings, but they are all later than the 1980s. I have based the RYF-1961 on experience of Judith Hauck who have used JRA-RYF-1961 for their spinup (1961 is the earliest year in JRA with neutral ENSO and neutral other climate indices).

I'm currently testing this and will run ~400 years with this forcing and compare to the JRA-IAF. The diagnostics are here:

https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/jschwing/NOIIAJRAOC1850_TL319_tn14_42/
https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/jschwing/NOINYJRARYF1961OC_TL319_tn14_42/

TomasTorsvik:
@JorgSchwinger - If you want this for NorESM2.3, you should push this to the branch `release-1.6`, which has already been created for that release. Then we should tag the `release-1.6` branch again.

The `master` branch is now the development branch for NorESM2.5. I suppose we want JRA forcing also for NorESM2.5, but I don't know if we need it with MCT support. The plan going forward is to eventually remove MCT.

JorgSchwinger:
The compsets (which this PR defines) should be carried forward in any case. For NorESM2.5 implementation is still missing in cime/CDEPS this will be done later. For BLOM for release 2.3 this commit could be cherry-picked?

TomasTorsvik:
> The compsets (which this PR defines) should be carried forward in any case. For NorESM2.5 implementation is still missing in cime/CDEPS this will be done later. For BLOM for release 2.3 this commit could be cherry-picked?

Yes, once it has been approved and merged, I can do a cherry-pick to `release-1.6`.

matsbn (approved):
I think this should be fine to merge.

TomasTorsvik (approved):

mvertens (approved):

Issue #390, opened 2024-09-06
Testing with hybrid coordinates show spikes in surface pCO2 - problem in Bothnian Sea?


TomasTorsvik:
There is a problem with the hybrid model run as seen in pCO2 time series. This can be traced down to the Bothnian Sea, where a couple of grid points have (annual mean) pCO2 that reaches several 10 000 mu atm. One observation that could be related is that the layer thickness in this region is very large: The top layer is >30 m as opposed to almost everywhere else where it is around 2.5 m. The 2-degree hybrid run is generally relatively noisy (given the normal year forcing) and there seem to be occasional convection(?) events in the Southern Ocean (best seen in CO2- and O2-fluxes).

Diagnostics: https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/jschwing/

Short explanation of runs:
-----------------------------

NOINYOC_T62_tn21_27: CMIP6 version of the model
NOINYOC_T62_tn21_37: Current master (sediment acceleration is switched on by default for 800 years) but increased sediment remineralization of POC (disso_poc = 5.0e-6)
NOINYOC_T62_tn21_38: Same as 37, but increased sediment remineralization of POC (disso_poc = 5.0e-6) AND increased background diffusivity (BDMC2 = 1.2e-5)
NOINYOC_T62_tn21_39: Same as 37, but with extendend N-cycle turned ON
NOINYOC_T62_tn21_40: Same as 39, but with hybrid coordinates turned ON
NOINYOC_T62_tn21_41: Same as 40, but with neutral diffusion turned off

All runs (except the CMIP6 one of course) are based on current master

JorgSchwinger:
@matsbn @TomasTorsvik @jmaerz, I have now finished a 200 year hybrid-coordinate NOINYOC run (CORE normal year forcing) on the tnx1v4 grid for comparison with the same configuration on the tnx2v1 grid.

The 1-degree resolution shows in principle the same thing in the northern Baltic, just a bit weaker: There is elevated, noisy pCO2 in the northern Baltic sea, which can be even seen as large spikes in the global time-series (blue is the tnx1v4 grid):

![NorDiag_gts_T62_tn14_40-tn21_40_0001-0200_p09](https://github.com/user-attachments/assets/87e1ed73-5122-42c8-b2ee-4f759e6b588b)
 
[...]

PR #432, opened 2024-11-22
Enforce a range of layer interfaces to constant minimum pressure levels for BLOM hybrid vertical coordinate

matsbn:
Currently for BLOM hybrid vertical coordinate, a water column lighter that the reference potential density of the k = 2 interface would occupy only the first model layer. With this PR, a range of layer interfaces with indices [1, k_range_plevel] will be enforced to the prescribed constant minimum pressure levels, always ensuring some vertical resolution near the surface. Setting the namelist variable k_range_plevel = 1, reproduces the old behaviour. The default value of k_range_plevel has been set to 4 in NorESM configurations, which changes results for vcoord = 'cntiso_hybrid'.

@TomasTorsvik and @JorgSchwinger, it would be interesting to see if this PR would mitigate some of the issues reported in https://github.com/NorESMhub/BLOM/issues/390.

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -1120,6 +1120,16 @@
     <desc>Valid mehtods: 'nudge', 'direct'</desc>
   </entry>
 
+  <entry id="k_range_plevel">
+    <type>integer</type>
+    <category>ale_regrid_remap</category>
+    <group>ale_regrid_remap</group>
+    <values>
+      <value>4</value>
+    </values>
+    <desc>Interface index range from surface enforced to constant pressure levels</desc>
+  </entry>
+
   <entry id="regrid_nudge_ts">
     <type>real</type>
     <category>ale_regrid_remap</category>
--- a/phy/mod_ale_regrid_remap.F90
+++ b/phy/mod_ale_regrid_remap.F90
@@ -74,6 +74,7 @@ module mod_ale_regrid_remap
    integer :: &
         upper_bndr_ord = 6, &
         lower_bndr_ord = 4, &
+        k_range_plevel = 1, &
         dktzu = 4, &
         dktzl = 2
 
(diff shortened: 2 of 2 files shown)


JorgSchwinger (approved):
I'm not knowledgeable enough to understand the details, but otherwise it looks good to me

JorgSchwinger:
I'm happy to test this for the NOINYOC compsets. I would suggest to do this after this has been merged (since the old behavior can be easily reproduced by namelist changes)

TomasTorsvik (approved):
@matsbn - Thanks! Looks fine to me.

One question: Is the code block starting from line 625 in mod_ale_regrid_remap.F90 :  `do while (kt <= kdmx(i))` sometimes skipped (i.e. can we have `kdmx(i) < kt` in some locations)?

matsbn:
> @matsbn - Thanks! Looks fine to me.
> 
> One question: Is the code block starting from line 625 in mod_ale_regrid_remap.F90 : `do while (kt <= kdmx(i))` sometimes skipped (i.e. can we have `kdmx(i) < kt` in some locations)?

It may happen that kdmx(i) < kt. Then all destination interfaces for the regridding, p_dst(k,i), with k >= kt will be set to the pressure of the deepest interface (at the ocean floor). With k_range_plevel = 1, this would mean that the whole water column is placed in the top layer (k = 1), while k_range_plevel > 1 will ensure some resolution near the surface.

JorgSchwinger:
Related to this, is there a minimum layer thickness that is maintained?

matsbn:
> Related to this, is there a minimum layer thickness that is maintained?

Yes, for interior layers (not layers that can have zero thickness at the ocean floor), the minimum thickness is controlled by the namelist variable dpmin_interior. Currently this is set to 0.1 m (converted to a pressure thickness in the code).

PR #437, opened 2024-11-27
Modified compsets with JRA interannual forcing to use year range 1958-2018

matsbn:
With this both NUOPC and MCT based NorESM will use the same year range for JRA interannual forcing (1958-2018). This is then consistent with what was used for NorESM2 in CMIP6/OMIP2.

This PR also modifies the processor configuration for TL319_tnx1v4 grid combination to be consistent with or without iHAMOCC enabled. This avoids the PIO_STRIDE related crash on Betzy reported in https://github.com/NorESMhub/NorESM/issues/593.

### DIFF
--- a/cime_config/config_compsets.xml
+++ b/cime_config/config_compsets.xml
@@ -78,7 +78,7 @@
 
   <compset>
     <alias>NOIIAJRA</alias>
-    <lname>2000_DATM%JRA_SLND_CICE_BLOM_DROF%JRA_SGLC_SWAV</lname>
+    <lname>2000_DATM%JRA-1p4-2018_SLND_CICE_BLOM_DROF%JRA-1p4-2018_SGLC_SWAV</lname>
   </compset>
   <compset>
     <!-- not valid for noresm2.5 -->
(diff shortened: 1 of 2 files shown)


JorgSchwinger (approved):
With MCT, JRA v1.3 is used /cluster/shared/noresm/inputdata/ocn/jra55/v1.3_noleap

The new compset names suggest that JRA v1.4 is used? Is this version different from 1.3?

Otherwise, looks good to me

matsbn:
> With MCT, JRA v1.3 is used /cluster/shared/noresm/inputdata/ocn/jra55/v1.3_noleap
> 
> The new compset names suggest that JRA v1.4 is used? Is this version different from 1.3?
> 
> Otherwise, looks good to me

In an email from Hiroyuki Tsujino to an OMIP-CMIP email-list, he mentions:

> Version 1.3.2 and 1.4.0 differ only in runoff around Greenland: version 1.3.2 uses climatological (1961-1990) runoff from  Greenland (Bamber et al.2012) and version 1.4.0 uses interannually varying (1958-2016) runoff from Greenland and Canadian Arctic Archipelagos (Bamber et al.2018)

For NorESM2, we actually use v1.4 for the OMIP2 runoff forcing, combined with v1.3 for the atmospheric forcing. So I think the combined forcing in NorESM2 should be quite consistent with JRA v1.4, although I have not checked carefully. Anyway, with NorESM2.5 and beyond, identical results is not expected with NorESM2 and also we now use CICE6 compared to CICE5 in OMIP type experiments. Therefore, I think it is fine to use "pure" JRA v1.4 with NorESM2.5. Furthermore, JRA v1.5 is now supported by CDEPS (https://github.com/NorESMhub/CDEPS/issues/9), extending the JRA forcing to 2023, and we should soon consider to try out this protocol which might be relevant for CMIP7/OMIP2.

JorgSchwinger:
Thanks for the info @matsbn, I agree

TomasTorsvik (approved):

PR #446, opened 2024-12-09
Update constants K1 and K2 for carbon chemistry

JorgSchwinger:
This PR introduces updated equilibrium constants K1 and K2 for the carbonate system (Waters et al. 2014), which are valid down to low salinities. The limits for temperature and salinity for the carbon chemistry and gas-echange are adjusted such that salinities down to 5 psu are now taken into account. The limiting is done consistently for the sediment and water column using parameters, which values are set in `mo_carchm`. This PR (together with #432) solves the problems described in #390.

Changes are mainly observed in the Baltic Sea where low salinities occur (see screenshots below, new results to the left). Note the unstable behavior in the previous results, with very high and very low pCO2 in adjacent grid-cells in the northern Baltic Sea. Global time series otherwise indicate only very minor changes.

Screenshot from 2024-12-09 11-11-31
Screenshot from 2024-12-09 11-12-24

@jmaerz , please critically review to make sure that there are no problems related to the extended nitrogen cycle when salinities get as low as 5 psu. The model runs can be found here

/projects/NS2980K/schwinger/NOINYOC_T62_tn21_44  (old)
/projects/NS2980K/schwinger/NOINYOC_T62_tn21_45  (new)

### DIFF
--- a/hamocc/mo_carchm.F90
+++ b/hamocc/mo_carchm.F90
@@ -25,7 +25,7 @@ module mo_carchm
   public  :: carchm
   public  :: carchm_solve
 
-  ! Maximum numnber of iterations for carbon chemistry solver
+  ! Maximum number of iterations for carbon chemistry solver
   integer, parameter :: niter=20
 
   ! Accuracy for test of convergence in carbon chemistry solver
@@ -223,7 +223,7 @@ contains
             tk100= tk/100.0
             
             rrho = prho(i,j,k)                   ! seawater density [g/cm3]
-            prb  = ptiestu(i,j,k)*98060*1.027e-6 ! pressure in unit bars, 98060 = onem
+            prb  = ptiestu(i,j,k)*98060.0*1.027e-6 ! pressure in unit bars, 98060 = onem
             
             tc   = ocetra(i,j,k,isco212) / rrho  ! convert to mol/kg
             ta   = ocetra(i,j,k,ialkali) / rrho
@@ -268,7 +268,7 @@ contains
             oxy=ox0+ox1/tk100+ox2*alog(tk100)+ox3*tk100+s*(ox4+ox5*tk100+ox6*tk100**2)
             satoxy(i,j,k)=exp(oxy)*oxyco
 
-            if (k.eq.1) then
+            if (k==1) then
 
               ! -----------------------------------------------------------------
               ! Calculate Schmidt numbers, solubilities, and piston velocities
@@ -302,7 +302,7 @@ contains
 
                 ! molecular viscosity of sea water
                 ! (Matthaeus 1972, Richards 1998,assuming salinity s in per mille = ~PSU)
-                p_dbar =  ppao(i,j)*1e-4  ! sea level pressure (Pa *1e-5 -> bar *10-> dbar
+                p_dbar =  ppao(i,j)*1.0e-4  ! sea level pressure (Pa *1e-5 -> bar *10-> dbar
                 mu_w   = 1.79e-2 - 6.1299e-4 * t + 1.4467e-5 * t2 - 1.6826e-7 * t3          &
                        & - 1.8266e-7 * p_dbar + 9.8972e-12 * p_dbar*p_dbar + 2.4727e-5 * s  &
                        & + s * (4.8429e-7 * t - 4.7172e-8 * t2 + 7.5986e-10 * t3)           &
@@ -312,7 +312,7 @@ contains
  
                 ! diffusion coeff in air (m2/s) Fuller 1966 / Johnson 2010
                 ! division by pressure: ppao [Pa]; in Fuller, p is a factor for denominator [atm]
-                diff_nh3_a =  1e-7 * tk**1.75 * M_nh3 / (ppao(i,j)/101325.0)
+                diff_nh3_a =  1.0e-7 * tk**1.75 * M_nh3 / (ppao(i,j)/101325.0)
 
                 ! Johnson 2010 - (34) cm2/s -> m2/s (1e-8*1e-4=1e-12)
                 ! closer to fit for Li & Gregory of: 9.874e-6*exp(2.644e-2*t)
@@ -348,14 +348,14 @@ contains
                 a_sf = exp(-80.0343  + 117.232 *(100/tk) +  29.5817*log(tk100)       &
                      &     +s*(0.033518-0.0373942*(tk100)+0.00774862*(tk100)**2))
                 ! conversion from mol/(l * atm) to kmol/(m3 * pptv)
-                a_11 = 1e-12 * a_11
-                a_12 = 1e-12 * a_12
-                a_sf = 1e-12 * a_sf
+                a_11 = 1.0e-12 * a_11
+                a_12 = 1.0e-12 * a_12
+                a_sf = 1.0e-12 * a_sf
               endif
               if (use_BROMO) then
                 !Henry's law constant [dimensionless] for Bromoform from Quack and Wallace (2003; GBC)
                 ! temp=[0,20]
-                a_bromo = exp(13.16 - 4973*(1/tk))
+                a_bromo = exp(13.16 - 4973.0*(1.0/tk))
               endif
               if (use_extNcycle) then
                 !Henry number for NH3 (Paulot et al. 2015, )
@@ -391,7 +391,7 @@ contains
                 ! wind drag coeff (-)
                 cD_wind = (u_star / (pfu10(i,j) + eps_safe))**2.
                 ! gas transfer velocity on gas phase side (m/s)
-                ka_nh3  = 1e-3 + u_star/ (13.3*sch_nh3_a + (eps_safe + cD_wind)**(-0.5) - 5. + log(sch_nh3_a)/(2.*kappa))
+                ka_nh3  = 1.0e-3 + u_star/ (13.3*sch_nh3_a + (eps_safe + cD_wind)**(-0.5) - 5. + log(sch_nh3_a)/(2.*kappa))
                 ! gas transfer velocity on liquid phase side (m/s) Nightingale 2000b - 3600*100: cm/h -> m/s
                 kw_nh3  =  (0.24*pfu10(i,j)**2 + 0.061*pfu10(i,j))*sqrt(600./sch_nh3_w)/360000.
 
@@ -435,7 +435,7 @@ contains
               pH2O  = exp(24.4543 - 67.4509*(100.0/tk) - 4.8489*log(tk/100.0) - 0.000544*s)
 
               ! Calculate the CO2 concentration in equilibrium with atmospheric x' (atco2=mole fraction of CO2 in dry air [ppm])
-              cu_sat = Kh0*atco2*1e-6*(rpp0-pH2O)*fc
+              cu_sat = Kh0*atco2*1.0e-6*(rpp0-pH2O)*fc
 
               fluxd = cu_sat*kwco2*dtbgc*rrho ! cu_sat and cu are in mol/kg. Multiply by rrho (g/cm^3) 
               fluxu = cu    *kwco2*dtbgc*rrho ! to get fluxes in kmol/m^2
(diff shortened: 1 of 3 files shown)


hamocc/mo_carchm.F90
```diff
@@ -901,17 +906,19 @@ subroutine carchm_solve(saln,tc,ta,sit,pt,K1,K2,Kb,Kw,Ks1,Kf,Ksi,K1p,K2p,K3p,ah1
       ah2o = sqrt( ( tc - ac )**2 + 4. * ( ac * K2 / K1 ) * ( 2. * tc - ac ) )
       ah2  = 0.5 * K1 / ac *( ( tc - ac ) + ah2o )
       erel = ( ah2 - ah1 ) / ah2
-      if (abs( erel ).ge.eps) then
-        ah1 = ah2
+      if (abs( erel ) >= eps) then
+        ! make sure [H+] stays between pH 5 and 11
+        ah1 = max(1.0e-11,min(ah2,1.0e-5))
```
TomasTorsvik:
```suggestion
        ah1 = max(min_ah,min(ah2,max_ah))
```
Maybe introduce this as (private) static parameters in mo_carchm? Reduce risk of divergence between carchm_solve and carchm_solve_dicsat.

hamocc/mo_carchm.F90
```diff
@@ -971,9 +975,11 @@ subroutine carchm_solve_dicsat(saln,co2_sat,ta,sit,pt,Kh0,K1,K2,Kb,Kw,Ks1,Kf,Ksi
       ah2o = sqrt((K1*co2_sat)**2 + 4.*ac*2.*K1*k2*co2_sat)
       ah2  = (K1*co2_sat + ah2o)/(2.*ac)
       erel = ( ah2 - ah1 ) / ah2
-      if (abs( erel ).ge.eps) then
-        ah1 = ah2
+      if (abs( erel ) >= eps) then
+        ! make sure [H+] stays between pH 5 and 11
+        ah1 = max(1.0e-11,min(ah2,1.0e-5))
```
TomasTorsvik: ... and here

TomasTorsvik (approved):
@JorgSchwinger - Thanks, looks fine to me!

hamocc/mo_carchm.F90
```diff
@@ -240,11 +245,9 @@ subroutine carchm(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,prho,pglat,omask,psicomo
               ta   = ocetra(i,j,k,inatalkali) / rrho
               ah1  = nathi(i,j,k)
 
-              call CARCHM_SOLVE(s,tc,ta,sit,pt,K1,K2,Kb,Kw,Ks1,Kf,Ksi,K1p,K2p,K3p,ah1,ac,niter)
+              call CARCHM_SOLVE(psao(i,j,k),tc,ta,sit,pt,K1,K2,Kb,Kw,Ks1,Kf,Ksi,K1p,K2p,K3p,ah1,ac)
```
jmaerz: Why `psao` here (and not just `s` to be consistent with the call above)?

hamocc/mo_carchm.F90
```diff
@@ -254,8 +257,8 @@ subroutine carchm(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,prho,pglat,omask,psicomo
               natco3(i,j,k) = natcc * rrho
             endif
 
-            ! solubility of O2 (Weiss, R.F. 1970, Deep-Sea Res., 17, 721-735) for moist air
-            ! at 1 atm; multiplication with oxyco converts to kmol/m^3/atm
+            ! solubility of O2 (Weiss, R.F. 1970, Deep-Sea Res., 17, 721-735) for moist air at
+            ! 1 atm; multiplication with oxyco converts to kmol/m^3/atm, temp=[-1,40], saln=[0,40]
```
jmaerz, line 267: here and below: I would suggest to write `temp_min`,`temp_max` , similar for saln (or at least correct the values below: -2 -> -1, 30 -> 40, ...)

hamocc/mo_carchm.F90
```diff
@@ -413,11 +420,12 @@ subroutine carchm(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,prho,pglat,omask,psicomo
               rpp0  = ppao(i,j)/101325.0
 
               ! calculate correction for non-ideality of CO2 (fugacity coefficient, Weiss and Price 1980)
-              Bvir  = -1636.75 + 12.0408*tk -0.0327957*tk**2 + 0.0000316528*tk**3
-              delta = 57.7-0.118*tk
+              Bvir  = -1636.75 + 12.0408*tk -0.0327957*tk**2 + 0.0000316528*tk**3   ! temp=[-12,47]
```
jmaerz, line 429: temp ranges in comments?

hamocc/mo_carchm.F90
```diff
@@ -765,20 +770,24 @@ subroutine carchm_kequi(temp,saln,prb,Kh0,K1,K2,Kb,Kw,Ks1,Kf,Ksi,K1p,K2p,K3p,Ksp
 
     ! Kh0 = [CO2]/ fCO2 (fCO2 = fugacity of CO2 in air)
     ! Weiss (1974), note this does not include a correction for the effect of moist air at 
-    ! air-sea interface [mol/kg/atm]
+    ! air-sea interface [mol/kg/atm], temp=[-1,45], saln=[0,45]
```
jmaerz, line 783: ranges?

hamocc/mo_carchm.F90
```diff
@@ -765,20 +770,24 @@ subroutine carchm_kequi(temp,saln,prb,Kh0,K1,K2,Kb,Kw,Ks1,Kf,Ksi,K1p,K2p,K3p,Ksp
...
     nKhwe74 = ad1+ad2/tk100+ad3*log(tk100)+s*(bd1+bd2*tk100+bd3*tk100**2)
     Kh0     = exp( nKhwe74 )
     ! K1 = [H][HCO3]/[H2CO3]   ; K2 = [H][CO3]/[HCO3]
-    ! Millero p.664 (1995) using Mehrbach et al. data on seawater scale
-    K1 = 10**( -1.0 * ( 3670.7 * invtk - 62.008 + 9.7944 * dlogtk - 0.0118 * s + 0.000116 * s2 ) )
-    K2 = 10**( -1.0 * ( 1394.7 * invtk + 4.777 - 0.0184 * s + 0.000118 * s2 ) )
+    ! Waters et al. (2014) on total pH scale, temp=[0,45], saln=[0,45]
```
jmaerz, line 787: here and below - do the comments provide the valid ranges for the formulations? - or should it be what comes from BLOM?

hamocc/mo_powach.F90
```diff
@@ -432,7 +430,7 @@ subroutine powach(kpie,kpje,kpke,kbnd,prho,omask,psao,ptho,lspin)
             K2p = keqb( 9,i,j)
             K3p = keqb(10,i,j)
 
-            call carchm_solve(saln,c,alk,sit,pt,K1,K2,Kb,Kw,Ks1,Kf,Ksi,K1p,K2p,K3p,ah1,ac,niter)
+            call carchm_solve(psao(i,j,kbo(i,j)),c,alk,sit,pt,K1,K2,Kb,Kw,Ks1,Kf,Ksi,K1p,K2p,K3p,ah1,ac)
```
jmaerz, line 433: why not remaining consistent with what is calculated in the water column? - do we expect the sediment salinity always being between 5 and 40?

jmaerz:
@JorgSchwinger when you refer to the N-cycle: did you mean it as general comment or particularly with respect to the changes that you were putting in?
Further: could you provide the full reference to the updated constants?
And: any incentives to upgrade the whole carbon chemistry system to mocsy (www.geosci-model-dev.net/8/485/2015/
) in line with preferred CMIP protocols? That would also avoid the salinity issues (https://doi.org/10.5194/gmd-10-2169-2017).

hamocc/mo_carchm.F90
JorgSchwinger: Yes, this would make sense, will change accordingly

hamocc/mo_carchm.F90
JorgSchwinger: Yes, this didn't make sense. Fixed.

hamocc/mo_carchm.F90
JorgSchwinger: The temperature and salinity ranges given in comments indicate the valid range of the fit. This is usually the range for which measurements have been available, taken from the original publication. I have added a comment in carchm_solve to clarify this

hamocc/mo_powach.F90
JorgSchwinger: I don't understand this comment. It was inconsistent before (water column saln=[25,40], sediment saln=[0,40]), now it is consistent. I think a range of [5,40] should be good enough.

JorgSchwinger:
> @JorgSchwinger when you refer to the N-cycle: did you mean it as general comment or particularly with respect to the changes that you were putting in?

I meant in particular with respect to the changes in the salinity range (temperature range is actually decreased from [-3,40] to [-1,40]). There are a few formulas (Schmidt number, etc.) that contain salinity as an parameter. Usually these are quite robust against extrapolation, but I just wanted to make sure you are aware of this.

Waters et al. (2014): http://dx.doi.org/10.1016/j.marchem.2014.07.004

This is essentially what mocsy uses, too (the same set of measurements, a slightly different fit, won't make a difference). So regarding a switch to mocsy, I'm not against it, but right now I don't see a large incentive (a lot of work for getting almost identical results).

hamocc/mo_powach.F90
jmaerz: Hi @JorgSchwinger, ok, I got confused through the doubling above of 
`s = min(saln_max,max(saln_min,psao(i,j,k)))`
which is sometimes also calculated before the call to `carchm_solve` - so my comment above was also due to this confusion.

JorgSchwinger: Yes, I agree the doubling is not so nice. I finally put it in, because we also use carchm_solve in the sediment and I didn't see a much better solution.

jmaerz: Hi @JorgSchwinger , maybe I am missing something, but we could leave out the 
`s = min(saln_max,max(saln_min,saln))`
in `carchm_kequi` and `carchm_solve` since one can calculate it before (as now also done for the water column - would be needed to put in again in `mo_powach`, which was my initial cause of confusion).

jmaerz:
I am not sure, if it is needed, but we previously had the minimum temperature of `-3` in `mo_carchm` congruent with the one used in `mo_ocprod`. Now this deviates from each other. Not sure, if the congruency is needed, though. If you prefer to do it, I would recommend to follow the approach with `temp_max`,`temp_min` as for salinity as suggested by Tomas.

JorgSchwinger:
> I am not sure, if it is needed, but we previously had the minimum temperature of `-3` in `mo_carchm` congruent with the one used in `mo_ocprod`. Now this deviates from each other. Not sure, if the congruency is needed, though. If you prefer to do it, I would recommend to follow the approach with `temp_max`,`temp_min` as for salinity as suggested by Tomas.

I don't think we need consistency between `mo_carchm` and `mo_ocprod` here. The ranges in temperature and salinity solely ensure that the parameterizations are not used out of their range. Not too far out out of their range one should better say, because we have been (and still are) exceeding the range for which observational data has been available for the functional fits of some of the quantities. I think this is not too worrying, since most of functional fits are quite "friendly", they show a smooth behavior outside the data range, they don't go crazy as some polynomial fits do (maybe an exception would be the Wannikhoff Schmidt numbers, which are polynomials in t).

I reduced `temp_min` from -3 to -1 because none of the functional fits in `mo_carchm` is actually based on observations for such low temperatures (some are "valid" down to -1, including the Waninkhoff 2014 stuff). I have not checked any ranges for temperature and salinity in `mo_ocprod`. I'm pretty sure that there is no data down to -3 for e.g. the Eppley temperate curve(?). So we could also adjust this value, but in general this is independent from the changes in`mo_carchm`, I'd say.

jmaerz:
Hi @JorgSchwinger , I am fine with the decision - just wanted to raise it to be aware of it. I'll have an additional look into the N-cycle stuff tmw (while I assume that there isn't any issue by your changes).

hamocc/mo_carchm.F90
```diff
@@ -25,9 +25,27 @@ module mo_carchm
   public  :: carchm
   public  :: carchm_solve
 
-  private :: carchm_kequi
-  private :: carchm_solve_dicsat
+  ! Maximum numnber of iterations for carbon chemistry solver
```
jmaerz: number

jmaerz (approved):
Hi @JorgSchwinger , I am currently don't see a reason, why the nitrogen cycle formulations should now fail, when the range of applicable temperatures has been rather narrowed compared to before. So thanks for making the model more proof for smaller salinities!
Maybe one thing and if you have time: the compiler should handle it, but in the in the routine a number of float values are written as int (without the dot), something one could change...

JorgSchwinger:
OK, I'll fix the integers before merging.

PR #449, opened 2024-12-12
Add time attributes to be cdo/xr-conform

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , when I was running some python-xarray routines to plot the netcdf-inventory written out by iHAMOCC, I realized that xarray had issues to interpret the time dimension correctly. Some attributes for the time dimension were missing, which I introduce with this PR. I hope it doesn't break any of your own post-processing pipelines - please check carefully and let me know.

@TomasTorsvik , since you have written this inventory part: please check carefully (also, whether the time is correctly set - I assumed that the previous global attribute `date`  was linked to the time dimension). 

Only tested in single column mode. Limited idea, why the build fails here.

### DIFF
--- a/hamocc/mo_inventory_bgc.F90
+++ b/hamocc/mo_inventory_bgc.F90
@@ -827,7 +827,6 @@ contains
         !--- Define global attributes
         call nccheck(NF90_PUT_ATT(ncid,NF90_GLOBAL,'title','Global inventory for marine bgc') )
         call nccheck(NF90_PUT_ATT(ncid,NF90_GLOBAL,'history','Global inventory for marine bgc') )
-        !call nccheck(NF90_PUT_ATT(ncid,NF90_GLOBAL,'date', timeunits) )
 
         !--- Define dimensions
         if (.not. use_sedbypass) then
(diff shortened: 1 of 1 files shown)


TomasTorsvik:
@jmaerz - it has been a while since I looked into this file. Will need some time to test and see what is going on.

TomasTorsvik (approved):
@jmaerz - looks good to me. Not sure why the date was defined as a global attribute in the first place.

hamocc/mo_inventory_bgc.F90
```diff
@@ -827,7 +827,7 @@ subroutine write_netcdf(iogrp)
         !--- Define global attributes
         call nccheck(NF90_PUT_ATT(ncid,NF90_GLOBAL,'title','Global inventory for marine bgc') )
         call nccheck(NF90_PUT_ATT(ncid,NF90_GLOBAL,'history','Global inventory for marine bgc') )
-        call nccheck(NF90_PUT_ATT(ncid,NF90_GLOBAL,'date', timeunits) )
+        !call nccheck(NF90_PUT_ATT(ncid,NF90_GLOBAL,'date', timeunits) )
```
JorgSchwinger, line 830: Could be deleted instead of commented out?

JorgSchwinger (approved):
Looks good, thanks for fixing this.

hamocc/mo_inventory_bgc.F90
jmaerz: Done.

PR #450, opened 2024-12-12

TomasTorsvik:
Same as #448, applied for the `release-1.6` branch.

### DIFF
--- a/cime_config/ocn_in_paramgen.py
+++ b/cime_config/ocn_in_paramgen.py
@@ -1338,6 +1338,22 @@ class OcnInParamGen(ParamGen):
                         line = remove_user_nl_comment(line)
                     #End if
 
+                    #Check if the first character on the line is a hash sign (#):
+                    #This is no longer allowed as a first character, and may indicate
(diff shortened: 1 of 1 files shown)


jmaerz (approved):
Fine with me - thanks for taking care!

jmaerz:
Any ideas, why the continuous integration testing fails? - I had similar issues in #449

jmaerz:
As written in #451 one could handle the line also as user comment line - again, not knowing ad hoc, whether this introduces other side effects...

TomasTorsvik:
> Any ideas, why the continuous integration testing fails? - I had similar issues in #449

@jmaerz - no, not really. Looks like something connected with linking to the MPI library. I see a lot of messages "undefined reference to `mpi_<somethig>' from building mod_xc.F.

jmaerz:
@TomasTorsvik , yes, I saw it, similar in #449 - but there just with the undefined references to mpi functions...

JorgSchwinger (approved):

PR #451, opened 2024-12-12

TomasTorsvik:
Same as https://github.com/NorESMhub/BLOM/pull/448, applied for the `master` branch.

### DIFF
--- a/cime_config/ocn_in_paramgen.py
+++ b/cime_config/ocn_in_paramgen.py
@@ -1338,6 +1338,22 @@ class OcnInParamGen(ParamGen):
                         line = remove_user_nl_comment(line)
                     #End if
 
+                    #Check if the first character on the line is a hash sign (#):
+                    #This is no longer allowed as a first character, and may indicate
+                    #the use of an old user_nl_blom file
+                    if line.strip()[0] == "#":
+                        emsg = f"Line number {line_num+1} in 'user_nl_blom'"
+                        emsg += " starts with a hash sign (#). This may"
+                        emsg += " indicate an outdated format of your"
+                        emsg += "\nuser_nl_blom file. Current versions of"
+                        emsg += " user_nl_blom file conform to the standard"
+                        emsg += " Fortran namelist syntax, indicating comments"
+                        emsg += "\nwith exclamation mark (!). Please check that"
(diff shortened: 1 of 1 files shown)


jmaerz (approved):
Hi @TomasTorsvik , technically approved, but I just realize that one could have potentially handled it also as a comment similar to the lines above (l. 1338) by
```
line=remove_user_nl_comment(line)
```
- not sure, though, whether this would induce other side effects... - similarly for the other cases...

TomasTorsvik:
> Hi @TomasTorsvik , technically approved, but I just realize that one could have potentially handled it also as a comment similar to the lines above (l. 1338) by
> 
> ```
> line=remove_user_nl_comment(line)
> ```
> 
>     * not sure, though, whether this would induce other side effects... - similarly for the other cases...

@jmaerz - True, but the old `user_nl_blom` files use a different format for setting variables, using a `set` keyword. One could also search for `set`, but this could cause issues if a variable name happened to start with `set...`. If the `user_nl_blom` file use # comments, probably something else could be wrong as well.

jmaerz:
Hi Tomas, thanks for clarification. I guess, moving forward in this case is totally fine - from my side, just go ahead with the PR.

JorgSchwinger:
It's fine for me including this, but I think it is actually not necesarry. 

The starting point was that older user_nl_blom files were not compatible with noresm 2.0.9. This was an issue because this model version is still (compatible to) our CMIP6 model, so a user could expect also the old user namelist files to work here.

However moving forward, there is no reason to expect this - format of user_nl_blom has changed (there are also other changes, not only the # vs !...), but that is fine because it is a new model version.

JorgSchwinger (approved):

TomasTorsvik:
> It's fine for me including this, but I think it is actually not necesarry.
> 
> The starting point was that older user_nl_blom files were not compatible with noresm 2.0.9. This was an issue because this model version is still (compatible to) our CMIP6 model, so a user could expect also the old user namelist files to work here.
> 
> However moving forward, there is no reason to expect this - format of user_nl_blom has changed (there are also other changes, not only the # vs !...), but that is fine because it is a new model version.

Yes, hopefully this will not have much use, at least when going to NorESM2.5. At the moment this change will make it into the planned `release-1.7` as well.

PR #452, opened 2024-12-13
Add N2 compset for NOINYOC that still uses isopycnic coordinates

JorgSchwinger:
The NOINYOC compset is used a lot for testing, so we need a N2NOINYOC compset that still selects isopucnic coordinates as a default.

### DIFF
--- a/cime_config/config_compsets.xml
+++ b/cime_config/config_compsets.xml
@@ -43,6 +43,11 @@
     <alias>NOINYOC</alias>
     <lname>1850_DATM%NYF_SLND_CICE_BLOM%HYB%ECO_DROF%NYF_SGLC_SWAV</lname>
   </compset>
+  <compset>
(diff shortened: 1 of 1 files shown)


jmaerz:
Stupid question, wouldn't it be better to keep the old naming for the `old` model version and use a new naming for the new, hybrid or other coordinate system?

JorgSchwinger:
Going ahead, these "old" compsets will be removed at some point.

Also, there are already a couple of other (ocean-only) compsets that use this naming convention.

jmaerz:
Will these compsets and changes be somewhere documented?

JorgSchwinger:
Once we want to scientifically support these compsets, they should be documented. Not sure if this will be the case for release 2.3? For now they only work with  --run-unsopported

TomasTorsvik (approved):

jmaerz (approved):

PR #453, opened 2024-12-18
R1.4 update ci02

TomasTorsvik:
It seems that our CI test system fails with `ubuntu-24.04`. Currently `ubuntu-latest` is in transition from `ubuntu-22.04` to `ubuntu-24.04` (see runner-images [issue 10636](https://github.com/actions/runner-images/issues/10636)), which triggers some mpi error.

* Set ubuntu version to 22.04
* Update meson and python version numbers

If approved, this fix will also be applied to the `release-1.6` and `master` branches.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,7 @@ on: [push, pull_request]
 
 env:
   # This version of Meson should be available as an EasyBuild on Fram and Betzy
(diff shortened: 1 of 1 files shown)


jmaerz (approved):
Looks good to me. Thanks!

JorgSchwinger (approved):

Sopheap-cloud (commented):
Download

PR #454, opened 2024-12-19
Temperature- and O2-dependent sediment remineralization

jmaerz:
with this PR, I make the temperature- and oxygen-dependency of the lumped ammonification&nitrification step in the default version as option available. As already for the water column, it is switched on for the default version by using the switch `lTO2depremin`. 

**Note:** Currently, there is no change in the `disso_poc` value automatically implemented (so it needs to be set in the `user_nl_blom` when using the `lTO2depremin` option (I currently use 6.0e-5 as a value, which might be a bit high in combination with `disso_sil=0.5e-6`). 

Further, in the extended nitrogen cycle, the denitrification rates are also oxygen- and temperature-dependent. I didn't include this thus far (and I am not sure, if it makes sense in the default version - maybe one could consider the T-dependency).


JorgSchwinger (approved):
Looks good to me, thanks for fixing this. I'll do a test run once this is merged.

Moving forward I would remove the option `lTO2depremin` and make this a default, if you agree.

I agree we should update `dissoc_poc`. I currently use `dissoc_poc=5.0e-6`, which gives a stable water column PO4 inventory. I currently doing a few tuning tests for the ecosystem parameterization and wanted to fine-tune `dissoc_poc` towards the end.

jmaerz:
Hi @JorgSchwinger , sure, we can make it a default at some point. Note that the `disso_poc` might need to be different in the current case and when you switch to the T-dep case, since the Q10-dep modifies the sort of average `disso_poc` with regards to the average bottom T (hope you understand, what I am trying to say). Further, with the current sediment in place, it is a bit of a question, which areas one wants to highlight in the tuning (i.e. that it better aligns to shallower or deep ocean regions).

JorgSchwinger:
Ok, I understand, thank you. But I do not necessarily think it's a question of highlighting a region, I think the priority would be to keep the water column PO4 constant (for given river input and the given sediment).

jmaerz:
Hi @JorgSchwinger , ideally yes... (while there might be slightly different needs in different setups). I discussed this with @TomasTorsvik the past days a bit. Coming closer to the tuning phase, we should discuss the default settings and model version to use - I would suggest to do this early next year.

JorgSchwinger:
Yes that's a good idea. See you next year, and have a nice Christmas break!

TomasTorsvik (approved):
@jmaerz - thanks, looks good to me.

PR #460, opened 2025-01-10
Update M4AGO scheme to version v1.1.1 (equivalent to dev-1.1.1)

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , with this PR, the M4AGO scheme tag is updated to an improved, updated M4AGO scheme version. For the updated version, see release notes: https://github.com/jmaerz/M4AGO-sinking-scheme/releases/tag/v1.1.0 and https://github.com/jmaerz/M4AGO-sinking-scheme/releases/tag/v1.1.1

Two notes:
- I have tested this only in the NorESM2.1 version thus far - I am not sure, if anything needs to be done wrt git fleximod
- I have kept the meson.build file backwards compatible to earlier M4AGO versions - not sure, if this should remain like this (it doesn't harm, but could be simplified with the updated version). 

Now steps taken to test with the NorESM development tag `noresm2_5_alpha08d` (for documenting the required steps to test branches)
```
git clone git@github.com:NorESMhub/NorESM.git
cd NorESM
git checkout noresm2_5_alpha08d
gvim .gitmodules
```
In `.gitmodules`, replaced
```
[submodule "blom"]
path = components/blom
url = https://github.com/NorESMhub/BLOM.git
fxtag = dev1.6.1.8
fxrequired = ToplevelRequired
fxDONOTUSEurl = https://github.com/NorESMhub/BLOM.git
```
by (**Note:** `new-m4ago` is just a branch, not a tag)
```
[submodule "blom"]
path = components/blom
url = https://github.com/jmaerz/BLOM.git
fxtag = new-m4ago
fxrequired = ToplevelRequired
fxDONOTUSEurl = https://github.com/NorESMhub/BLOM.git
```
Then, to update the submodules:
```
./bin/git-fleximod update --optional
```
After this call, everything should be/was in place to build and run the model as usual. For help on `git fleximod`, see also https://github.com/NorESMhub/NorESM/discussions/599 and https://github.com/ESMCI/git-fleximod.

However, currently the model version  `noresm2_5_alpha08d` fails, when running due to errors that seem not to be related to the updated scheme, but rather due to the ice mesh (see: https://github.com/NorESMhub/CMEPS/issues/28).

### DIFF
--- a/hamocc/mo_hamocc_init.F90
+++ b/hamocc/mo_hamocc_init.F90
@@ -47,7 +47,7 @@ contains
                               ocn_co2_type, use_sedbypass, use_BOXATM, use_BROMO,use_extNcycle,    &
                               use_nuopc_ndep,lTO2depremin
     use mo_param1_bgc,  only: ks,init_por2octra_mapping
-    use mo_param_bgc,   only: ini_parambgc
+    use mo_param_bgc,   only: ini_parambgc,claydens,calcdens,calcwei,opaldens,opalwei,ropal
     use mo_carbch,      only: alloc_mem_carbch,ocetra,atm,atm_co2
     use mo_biomod,      only: alloc_mem_biomod
     use mo_sedmnt,      only: alloc_mem_sedmnt,sedlay,powtra,burial,ini_sedmnt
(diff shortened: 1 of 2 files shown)


TomasTorsvik (approved):
@jmaerz - thanks, looks fine to me.
Betzy is not available, so I can't set up a NorESM2.5 build now (NorESM2.3 should be the same as NorESM2.1). Think it should be fine to merge this, and then deal with a potential problem if it should emerge with NorESM2.5. Or we could just wait a few days until betzy is availabe again.

TomasTorsvik:
I think it's fine to have the meson.buid file backwards compatible at this stage.

JorgSchwinger (approved):

jmaerz:
Hi Tomas, as summarized in the PR description, I tried a run with the most recent development tag of NorESM. To me, it seemed as if nothing else needed to be changed with respect to `git fleximod` than just the `.gitmodules` file as described (which will be required to be changed anyway, when a new tag is introduced for BLOM). Hence, I would suggest to merge the stuff into `master` by tmw noon so that you can prepare the BLOM tag 1.7, if you agree.

TomasTorsvik:
@jmaerz - Hi, I think it's fine to merge this. The bug in the `noresm2_5_alpha08d` run seems to be unrelated to this PR.

PR #463, opened 2025-01-16
Fix some namelist parameters for tnx2v1 grid

JorgSchwinger:
There were two BLOM namelist parameter that were not set differently from default for the tnx2v1 grid. This PR fixes this since the defaults did not make sense for the 2-degree grid (although it hasn't been explored if the new setting are optimal). The changes due to the new remain relatively small, but notably there is a small reduction in overall temperature biases. Diagnostics can be found here (comparison valid for physics only):

https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/jschwing/NOINYOC_T62_tn21_45/BLOM_DIAG/yrs991to1000-NOINYOC_T62_tn21_42/

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -337,6 +337,7 @@
     <group>limits</group>
     <values>
       <value>5.e-5</value>
+      <value ocn_grid="tnx2v1"    >.75e-4</value>
       <value ocn_grid="tnx1v4" ocn_ncpl="24">.75e-4</value>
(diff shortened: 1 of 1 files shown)


jmaerz (approved):
Hi Jörg, approved. Nevertheless a stupid question: why not just changing the default value and rather only include the exceptions in the namelist definition? But I guess, this is rather something for the BLOM developers.

JorgSchwinger:
Yes, I agree, but I didn't want to mess around with the BLOM namelist. I don't know where the default value comes from and for which configuration it would be suited.

TomasTorsvik (approved):

PR #464, opened 2025-01-16
Increase value of sediment POC degradation constant

### LINKED ISSUE #461: Correct POC remineralization rate

JorgSchwinger:
This increases the default value of disso_poc quite substantially. A larger value for this constant leads to a number of improvements (higher PP, reduced deep sea oxygen biases). Even if this parameter will need fine-tuning, the new value has a much better order of magnitude.

closes #461

### DIFF
--- a/hamocc/mo_param_bgc.F90
+++ b/hamocc/mo_param_bgc.F90
@@ -483,7 +483,7 @@ module mo_param_bgc
   !
   real, protected :: sedict      = 1.e-9          ! m2/s Molecular diffusion coefficient
   real, protected :: silsat      = 0.001          ! kmol/m3 Silicate saturation concentration is 1 mol/m3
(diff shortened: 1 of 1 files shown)


jmaerz:
Hi @JorgSchwinger , just to be sure (and to document): the new value is for the default model (so not for the recently included T- and O2-dependent sediment remin #454), right?

jmaerz (approved):
Thanks for performing runs and for putting this in!

JorgSchwinger:
It is ok-ish for both, the T-dependent remineralization does not have such a huge impact (globally) as far as I have seen. As mentioned above, this is not yet fine-tuned, but the order of magnitude is now much better.

jmaerz:
Hi @JorgSchwinger , thanks for checking - yes, the value is generally closer to what I would expect (and to which I tuned the sediment thus far for the N-cycle).

TomasTorsvik (approved):

PR #466, opened 2025-01-17
add git fleximod support to .gitmodules

### LINKED ISSUE #465: git fleximod support for submodules in BLOM (?)
Hi @TomasTorsvik , @matsbn and @JorgSchwinger , 
I wonder, whether we need to update the `.gitmodules` file in BLOM to fully align to the new `git fleximod` used in NorESM2.5 for the checkout of CVMIX and M4AGO. Currently, the update seems to work (also with `git fleximod`), while my suspicion is that it is rather due to the regular `git submodule` command that `git fleximod` seems somehow to wrap. 

I am about to prepare a branch with the `git fleximod` entries in `.gitmodules`, trying to test the behavior of `git fleximod` in combination with `git submodule`. 

Hence maybe worth to wait a bit with the tag release...

jmaerz:
Hi @TomasTorsvik , @matsbn , @JorgSchwinger ,
This PR add full fleximod support for both, CVMIX and M4AGO submodules. 

**One thing to note:**
 `git fleximod` seems to first check the regular `git submodule update` command. If then a` fx...` (`git fleximod`) entry is found in the `.gitmodules` file, it supersedes the submodule version that is committed via a regular git command (if they are different from each other). Hence, one needs to take care, which command to run when, if the regular git committed change in the submodule is not identical to the tag/branch entry of `git fleximod` (the `fx...` entries). This can lead to divergent versions to be used, if not taken care of manually (I kind of miss a `git fleximod update-to-regular-git-submodule` command to not having to manually edit the `.gitmodules` file, when committing submodule changes...). 
I currently made the entries identical between `git fleximod` and the regular `git submodule`.

Closes #465

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,12 @@
 [submodule "CVMix-src"]
 	path = pkgs/CVMix-src
 	url = https://github.com/CVMix/CVMix-src
+	fxtag = v0.98-beta
+	fxrequired = AlwaysRequired
+        fxDONOTUSEurl = https://github.com/CVMix/CVMix-src
 [submodule "pkgs/M4AGO-sinking-scheme"]
 	path = pkgs/M4AGO-sinking-scheme
 	url = https://github.com/jmaerz/M4AGO-sinking-scheme
+	fxtag = v1.1.1
+	fxrequired = AlwaysRequired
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):
@jmaerz - thanks for taking care of this!

PR #469, opened 2025-01-20
fix long_name attribute length for POC

jmaerz:
A very minor fix to properly write the POC long_name attribute to the restart file (previously, the 'n' of carbon was missing). I haven't checked, if it has some consequences to other variables, though. Just checked via single column setup, since I stumbled over it, when handling the restart file of iHAMOCC.


TomasTorsvik (approved):

JorgSchwinger (approved):
Thanks for fixing this

PR #471, opened 2025-01-24
Add Externals_BLOM.cfg file for use with NorESM2.3

### LINKED ISSUE #470: Update `config_compsets.xml` and `config_pes.xml` on release-1.6 branch?
There is a suggestions to port back relevant changes from the `v1.7` tag to the `release-1.6` branch, which is the current release branch for NorESM2.3. One such issue was #431.

I see that both `config_compsets.xml` and `config_pes.xml` files have evolved quite a lot between 1.6 and 1.7, so I wonder if there are parts of these files that should be ported.

TomasTorsvik:
This PR re-introduces the `Externals_BLOM.cfg` file, so that we can use the current `master` branch (currently `v1.7` tag and potentially the upcoming `v1.8` tag) in NorESM2.3.

See discussion in #470

### DIFF
--- a/Externals_BLOM.cfg
+++ b/Externals_BLOM.cfg
@@ -1,0 +1,16 @@
+[CVMix]
+tag = v0.98-beta
+protocol = git
+repo_url = https://github.com/CVMix/CVMix-src
+local_path = pkgs/CVMix-src
+required = True
+
+[M4AGO]
+tag = v1.1.1
+protocol = git
+repo_url = https://github.com/jmaerz/M4AGO-sinking-scheme
+local_path = pkgs/M4AGO-sinking-scheme
+required = True
+
+[externals_description]
+schema_version = 1.0.0


jmaerz (approved):
Thanks for converting back (and adjust for the most recent version)!

JorgSchwinger (approved):

gold2718 (approved):
Side question: If you had a choice, would you like to have a pure manage-externals like solution (i.e., externals file, does not use submodules to check out externals), git fleximod as it is today, or a hybrid tool that uses an externals file along with submodules?

jmaerz:
@gold2718 thanks for asking. I might be missing the point of `git fleximod`, but to my understanding, it doesn't do anything that the regular `git submodule` cannot handle - I currently feel as if `git fleximod` was written to be able to reference to model versions via written `git tags` in the `.gitmodule` file, while `git submodule` just refers to the checked-in sha number (of a particular tag) and it is not immediately obvious to the user, which tag (if any) is used - while I feel it is a weak motivation for `git fleximod` instead of the more common `git submodule`. The problem that I currently see with `git fleximod` is that it is not necessarily in sync with what `git submodule` has as information - and depending, which command you use, it can have different results. I haven't looked into `git fleximod` routines - how they are set up, etc. but I wish, it would at least provide some `git fleximod sync_with_submodule_information` and `git fleximod check_if_in_sync_with_submodules` functionality (in the future). 
Long story short: I wouldn't want to go back to manage_externals necessarily, but would consider to decrease the overhead even further and only use `git submodule` - but again, I might be missing something that is obvious to others.

gold2718:
> @gold2718 thanks for asking. I might be missing the point of `git fleximod`, but to my understanding, it doesn't do anything that the regular `git submodule` cannot handle - I currently feel as if `git fleximod` was written to be able to reference to model versions via written `git tags` in the `.gitmodule` file, . . .

@jmaerz, You are basically correct. The original motivation was the feeling at NCAR that manage_externals was too time consuming to maintain. manage_externals was written at a time when many externals were in subversion repositories and also when git submodules were not as mature as they are now. However, the community has a strong desire to have explicit documentation and control of the tags used in each external. `git fleximod` is an attempt to combine submodules with this documentation requirement. It seems to be okay with clean checkouts but seems to me that it is missing other functionality including correctly switching branches or tags as well as the items you mention.

Note that BLOM is free to use submodules without including `git fleximod`s `fx` keywords.

I have started a [discussion](https://github.com/NorESMhub/NorESM/discussions/630) on what features we need / want for our external checkout tool. I believe it includes the synchronization features you describe above.

PR #476, opened 2025-02-04
Initial version for iHAMOCC restart remapping

jmaerz:
Hi @TomasTorsvik , @JorgSchwinger ,

this draft PR encompasses the new structure for the BLOM `utils` that I proposed in #472 (also building on #83) and includes an iHAMOCC restart file remapping  `iHAMOCC_restart_remapping.py` that allows to remap iHAMOCC restart  files between

- different grids (hybrid versus isopycnic) 
- different number of pressure level layers (currently 53 in isopycnic versus 56 in hybrid)
- different grid resolutions (e.g. interpolation from lower to higher resolution, currently `tnx2`,`tnx1`,`tnx0.5`,`tnx0.25`) and
- supports the different unit systems (CGS versus MKS - needed to allow for remapping between old/new blom restart files)

and can perform an inventory adjustment, if needed. 

The default restart mapping currently first interpolates horizontally/along isopycnic layers and then performs a vertical interpolation to the center of the isopycnic layers. Only for the diagnostic water column variables, there is a vertical interpolation step carried out that performs some linear inter-/extrapolation for NaN-filling (which could also be applied in general to all water column variables, if it turns out to be needed - tbd). 

As far as I can see, it serves the current needs, while I still need/want to test it via using a remapped restart file in a run. 

**NOTE:** While the script is functional (and tested between different `tnx2` and `tnx1` grids), **it should be used with caution and newly produced restart files should be carefully checked**. 

The script could be further improved and cleaned/simplified - some notes in that respect:

- align dimensions and coordinates at the very beginning to avoid the current re-swapping of dim/coord-names
- currently, there are a number of hard-coded dimensions/coords - particularly in the vertical, which could be changed when aligning them in the beginning
- for the diagnostic variables `hi`,`co2star`,`satoxy` and `co3`, the used time step is still of concern - while I expect that it won't change the outcome dramatically irrespective of which one is used for remapping (can be set and tested via the yml-file)
- potentially perform the initial vertical interpolation step to fill NaN/missing values for all variables (not only the diagnostic ones)
- I still need to test the `tnx0.5` and `tnx0.25` support - are there a 0.5 and 0.25 degree BLOM runs somewhere, for which I can test the regridding?   

If you make use of it and it fails or crashes, please report back or make adjustments that can be pushed as well. Please also push improvements to the script in the future. 

I feel that it would be worth to have some common scripts under git control.


TomasTorsvik:
@jmaerz - thanks for providing the draft PR! I haven't  done any testing with `tnx0.5`, but I'm also not sure that it is needed for the initial code commit, as long as `tnx2` and `tnx1` are working.

jmaerz:
Hi @TomasTorsvik , as far as I tested, it works for the `tnx2` -> `tnx2` , `tnx2` -> `tnx1`  and also between isopycnic and hybrid, but since I am thus far not too knowledgeable wrt restart files and how they typically look (rarely looked into them), I might have missed certain particular cases, which is why I suggest to use the routine currently with a note of caution. There isn't much to add for `tnx0.5` (while I see that it's not yet available on nird - maybe, for simplicity, we could copy the gridfile there as well for such testing, regridding and diagnostic purposes?! - which would make adjusting the routine easier). If wished, I can also add `tnx0.125` support - expecting a 'cascadal' use from coarse to higher to high resolution iHAMOCC restart remapping after some years of running in-between.

jmaerz:
Hi @TomasTorsvik , I now copied `tnx0.5` to nird and added support for `tnx0.5` and `tnx0.25` - for `tnx0.125` full clarity is missing, which version is supposed to be used (`namelist_definition_blom.xml` refers to: `grid_tnx0.125v4_20221013.nc`, but I don't know, if that's the right/intended one, since there are lingering - likely older- other versions available). If needed, I would copy that to nird as well and would update the info in the remapping routine.

TomasTorsvik:
Hi @jmaerz , I checked CICE for `noresm2_3_alpha01`, grep for `tnx0.125`, I find
```
components/cice/cime_config/namelist_definition_cice.xml:      <value hgrid="tnx0.125v4">$DIN_LOC_ROOT/ocn/blom/grid/horiz_grid_tnx0.125v4_20200722.ieeer8</value>                                                                      
components/cice/cime_config/namelist_definition_cice.xml:      <value hgrid="tnx0.125v4">$DIN_LOC_ROOT/ocn/blom/grid/topography_tnx0.125v4_20200722.ieeei4</value>
```
So at least for the NorESM coupled system the 20200722 variant seems to be the one to use.

jmaerz:
Mh, ok, this looks like different variants are used for different purposes... - having also different y-dimensions. Maybe @matsbn  or @AleksiNummelin can comment (and potentially clean up)?

jmaerz:
I now added support for both grid versions... - since they are easily distinguishable through different grid dimensions.

JorgSchwinger (approved):
I have not looked at the details, but I think it is fine to include this

jmaerz:
I performed a run, where I was remapping an isopycnic, relatively well adjusted `tnx2` to an isopycnic `tnx1` iHAMOCC restart file. From visual inspection of trends, it looks as if after ten years, the `tnx1` run is not trend free, but in many variables, it seems to stabilize - even in the deeper ocean (meaning here 2000m). Most sensible seem to be O2, which can be expected to be susceptible to resolution effects. In summary, applying and investigating the remapping strategy further might be worth the effort. I merge this now, while there are certainly improvements to the routine possible (see the header of this PR for possible pathways). 

As written, please use with caution. A good initial check of the generated restart file is `cdo infov my_new_restart_file | grep nan` which shouldn't show anything. Other than that a visual inspection comparing the new and base/coarse restart file.

PR #480, opened 2025-02-05
Filtering of squared buoyancy frequency and Richardson number

matsbn:
Filtering of squared buoyancy frequency and Richardson number has been added to reduced noise in vertical diffusivity when using KPP. Along with the filtering, the critical gradient Richardson number for shear driven vertical mixing has also been adjusted.

This modification has been quite extensively tested in OMIP2 simulations and was also used in the simulation https://github.com/NorESMhub/noresm3_dev_simulations/issues/68 based on the noresm2_5_alpha08d tag. There is no change with isopycnic vertical coordinate.

### DIFF
--- a/phy/mod_difest.F90
+++ b/phy/mod_difest.F90
@@ -1,5 +1,5 @@
 ! ------------------------------------------------------------------------------
-! Copyright (C) 2009-2024 Mats Bentsen, Mehmet Ilicak, Aleksi Nummelin,
+! Copyright (C) 2009-2025 Mats Bentsen, Mehmet Ilicak, Aleksi Nummelin,
 !                         Mariana Vertenstein
 !
 ! This file is part of BLOM.
--- a/phy/mod_difest.F90
+++ b/phy/mod_difest.F90
@@ -1118,13 +1118,11 @@ contains
           rig_i(kk+1) = rig_i(kk)
           bvfsq_i(kk+1) = bvfsq_i(kk)
 
-          do n_iter = 1, 1 !n_smooth_ri
-            rig_i_prev(:) = rig_i(:)
-            bvfsq_i_prev(:) = bvfsq_i(:)
-            do k = 2, kk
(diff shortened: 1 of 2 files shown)


phy/mod_difest.F90
```diff
@@ -1088,7 +1116,17 @@ subroutine difest_vertical_hyb(m,n,mm,nn,k1m,k1n)
 
           ! bottom values for the Ri and N
           rig_i(kk+1) = rig_i(kk)
-          bvf_i(kk+1) = sqrt( max( bfsqi(i,j,kk+1), 0.) )
+          bvfsq_i(kk+1) = bvfsq_i(kk)
+
+          do n_iter = 1, 1 !n_smooth_ri
```
TomasTorsvik: If I understand correctly, this loop will smooth  `rig_i` and `bvfsq_i` N times (default N=1).
I would prefer either (1) to define the loop as `do niter = 1, n_smooth_ri` , and then define the `n_smooth_ri` parameter somewhere above, or (2) remove the loop if it will not be used.

phy/mod_difest.F90
```diff
@@ -1250,6 +1288,15 @@ subroutine difest_vertical_hyb(m,n,mm,nn,k1m,k1n)
           rig_i(kk+1) = rig_i(kk)
           bvfsq_i(kk+1) = bfsqi(i,j,kk+1)
 
+          do n_iter = 1, 1 !n_smooth_ri
```
TomasTorsvik: Same comment as above.

TomasTorsvik (approved):
Looks fine to me, just one thing I suggest to change (see separate comment).

phy/mod_difest.F90
matsbn: Thanks, for seeing this @TomasTorsvik. I have pushed a commit to remove these unnecessary loops.

milicak (approved):

PR #483, opened 2025-02-06
Update buildlib for M4AGO, ecosys dependency

### LINKED ISSUE #482: NOINY.T62_tn14 test fail due to mo_ihamocc4m4ago
Source code: `noresm2_3_alpha02`
BLOM tag: `v1.6.8`
Test: ERP_Ld3_P256.T62_tn14.NOINY.betzy_intel

This test fails at build stage with error
```
/cluster/projects/nn9560k/tomast/NorESM/src/noresm2_3_alpha02/components/blom/pkgs/M4AGO-sinking-scheme/src/mo_ihamocc4m4ago.f90(72): error #7002: Error in opening the compiled module file.  Check INCLUDE paths.   [MO_VGRID]
  use mo_vgrid,         only: dp_min
```

Since this is a NOINY test, I think we should not have any iHAMOCC dependency, and by extension no dependency on M4AGO source code. With `noresm2_3_alpha02` we are including M4AGO `v1.1.1` as external dependency. Note that this dependency is specified by the `Externals_BLOM.cfg` file for NorESM2.3, not git submodules (although, I don't expect that this makes a difference).

TomasTorsvik:
Only include M4AGO source files if building with "ecosys". This fix for `release-1.6` has already been implemented in `master`.

### DIFF
--- a/cime_config/buildlib_2.1
+++ b/cime_config/buildlib_2.1
@@ -66,7 +66,6 @@ def _main_func():
                      os.path.join(comp_root_dir_ocn, "channel"),
                      os.path.join(comp_root_dir_ocn, "single_column"),
                      os.path.join(comp_root_dir_ocn, "pkgs", "CVMix-src", "src", "shared"),
-                     os.path.join(comp_root_dir_ocn, "pkgs", "M4AGO-sinking-scheme", "src"),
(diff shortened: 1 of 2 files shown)


jmaerz (approved):
@TomasTorsvik , thanks for fixing and taking care of this. Approved.

PR #485, opened 2025-02-09
Ensure non-negative tracer values and conservation by virtual flux correction

matsbn:
Tracers are enforced to be non-negative and and any tracer quantity added to ensure non-negative values are compensation for by a global virtual flux correction.

The simulation n1850.ne30_tn14.hybrid_wavparam_vfcor_fatessp.20250204 was started as a hybrid simulation from year 29 of https://github.com/NorESMhub/noresm3_dev_simulations/issues/68 to test the impact of this PR. Diagnostics can be found here: https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/matsbn/n1850.ne30_tn14.hybrid_wavparam_vfcor_fatessp.20250204/

Some erratic biogeochemistry global flux time-series (net downward CO2 flux) and global averages (DIC and alkalinity) seems to be mitigated in the new simulation. It should be noted that recent changes to iHAMOCC included in the code base used for 1850.ne30_tn14.hybrid_wavparam_vfcor_fatessp.20250204 could also contribute to this improvement. Then there are difference to the evolution of global averages of oxygen, phosphate and nitrate that would be useful with comments on.

Global average salinity has a stronger increase in  n1850.ne30_tn14.hybrid_wavparam_vfcor_fatessp.20250204 compared to https://github.com/NorESMhub/noresm3_dev_simulations/issues/68. Apparently and unintentionally, CICE started from the initial condition set in user_nl_cice, and not the year 29 restart of https://github.com/NorESMhub/noresm3_dev_simulations/issues/68. The resulting sea-ice volume build-up might explain the salinity evolution. A new simulation is underway to make sure CICE starts from the year 29 restart of https://github.com/NorESMhub/noresm3_dev_simulations/issues/68.

### DIFF
--- a/ben02/mod_thermf_ben02.F90
+++ b/ben02/mod_thermf_ben02.F90
@@ -1,5 +1,5 @@
 ! ------------------------------------------------------------------------------
-! Copyright (C) 2002-2024 Mats Bentsen, Mehmet Ilicak
+! Copyright (C) 2002-2025 Mats Bentsen, Mehmet Ilicak
 !
 ! This file is part of BLOM.
 !
@@ -36,7 +36,7 @@ module mod_thermf_ben02
                            swa, nsf, hmltfz, lip, sop, eva, rnf, rfi, &
                            fmltfz, sfl, ustarw, surflx, surrlx, &
                            sswflx, salflx, brnflx, salrlx, ustar, &
-                           t_rs_nonloc, s_rs_nonloc
+                           salt_corr, trc_corr, t_rs_nonloc, s_rs_nonloc
   use mod_swabs,     only: swbgal, swbgfc
   use mod_ben02,     only: tsi_tda, tml_tda, sml_tda, alb_tda, fice_tda, &
                            tsi, ntda, dfl, albw, alb, &
@@ -46,13 +46,12 @@ module mod_thermf_ben02
                            fuss, fice_max, tice_m, tsnw_m, hice_nhmn, &
                            hice_shmn, sagets, sice, cwi, cuc
   use mod_seaice,    only: ficem, hicem, hsnwm, ustari, iagem
-  use mod_utility,   only: util1, util2, util3, util4
+  use mod_utility,   only: util1, util2, util3
   use mod_checksum,  only: csdiag, chksummsk
   use mod_intp1d,    only: intp1d
   use mod_tracers,   only: ntr, itrtke, itrgls, trc, trflx
   use mod_diffusion, only: difdia
   use mod_tke,       only: gls_cmu0, Zos, gls_p, gls_m, gls_n, vonKar
-  use mod_tracers,   only: ntr, trc, trflx
   use mod_ifdefs,    only: use_TRC, use_TKE, use_GLS
 
   implicit none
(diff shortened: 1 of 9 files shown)


TomasTorsvik:
@matsbn - Thanks, the changes in oxygen, phosphate and nitrate are likely due to PR #464. I tested the impact of this change, in
https://github.com/NorESMhub/noresm3_dev_simulations/issues/77
https://github.com/NorESMhub/noresm3_dev_simulations/issues/74
See in particular
https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/tomast/n1850.ne30pg3_tn14.2_5_alpha08d.BLOMv1.7.0.hybrid_wavparam_fatessp.20250124/HAMOCC_DIAG/ts1to28-n1850.ne30_tn14.hybrid_fatessp.20241204/set1/set1_ann_avg_2models.png

TomasTorsvik (approved):

JorgSchwinger (commented):
A general question about the time-stepping: Now the salt and tracer correction is calculated but applied one time-step later, correct? This is why it has to be saved in the restart files. When in this sequence is HAMOCC called? Is this _after_ the correction has been applied? HAMOCC cannot be called with negative tracer values. In fact we have the routine `trc_limitc`, which is called before HAMOCC and makes sure no negative values occur. Assuming that the virtual fluxes are the only potential source of negative values, this routine should then be removed?

matsbn:
> A general question about the time-stepping: Now the salt and tracer correction is calculated but applied one time-step later, correct? This is why it has to be saved in the restart files. When in this sequence is HAMOCC called? Is this _after_ the correction has been applied? HAMOCC cannot be called with negative tracer values. In fact we have the routine `trc_limitc`, which is called before HAMOCC and makes sure no negative values occur. Assuming that the virtual fluxes are the only potential source of negative values, this routine should then be removed?

Yes, the tracer added to ensure non-negative values is compensated as a correction of the virtual tracer flux applied the next time-level. Ensuring non-negative tracer values (of the 'k+nn' time-level) happens in routine 'mxlayr' when isopycnic coordinate is used and in routine 'ale_vdifft' otherwise. These routines are called before routine 'updtrc' is called, so there should be no negative tracer values entering HAMOCC (assuming 'k+nn' time-level is used). So I think the limiting in 'trc_limitc' can now be skipped.

JorgSchwinger:
Regarding the `trc_limitc` routine, I would suggest merging this now and then removing the routine later (we can first issue an error if negative tracer values are encountered to make sure everything works as expected). I can do this in a subsequent PR.

JorgSchwinger (approved):

matsbn:
> Regarding the `trc_limitc` routine, I would suggest merging this now and then removing the routine later (we can first issue an error if negative tracer values are encountered to make sure everything works as expected). I can do this in a subsequent PR.

Sounds like a good plan. A simulation https://github.com/NorESMhub/noresm3_dev_simulations/issues/86 with correct CICE restart file for a hybrid run of https://github.com/NorESMhub/noresm3_dev_simulations/issues/68 has soon completed 7+7 years. After the first 7 years, the salinity evolution is very close to the reference case: https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/matsbn/n1850.ne30_tn14.hybrid_wavparam_vfcor_fatessp.20250205/BLOM_DIAG/yrs29to35-n1850.ne30_tn14.hybrid_wavparam_fatessp.20241226/set1/set1_ann_salnga_2models.png. 

Let me verify that the last 7 years also look reasonable, then I'll merge.

matsbn:
The simulation n1850.ne30_tn14.hybrid_wavparam_vfcor_fatessp.20250205 (https://github.com/NorESMhub/noresm3_dev_simulations/issues/86) completed 13 years (crash in final year that seems unrelated to the code changes).

Diagnostics can be found here: https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/matsbn/n1850.ne30_tn14.hybrid_wavparam_vfcor_fatessp.20250205/

Time series of global average salinity and iHAMOCC fields are shown below. Salinity evolution seems fine and also iHAMOCC  averaged variables with the additional iHAMOCC changes taken into account.

image
image

PR #486, opened 2025-02-10
Provide new N-deposition files and used remapping sh-script

### LINKED ISSUE #397: Preparing N-deposition input files for CMIP6 scenarios and historical runs

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik.
with this PR new-generated N-deposition files, holding `nhx`, `noy` and `ndep=nhx+noy`. The latter currently re-gridded AFTER summing up, which might be of an issue, when moving forward, and only using `nhx` and `noy` in the code - so bfb cannot be expected - while this is also the case to former ndep-files, since different scripts and thus slightly different weights are used than before. They are introduced for `tnx2`, `tnx1` and `tnx0.5` (currently trying also for `tnx0.25`, while this seems to create a hickup on betzy (likely memory issues).   

I anticipate to develop the code further towards just making use of the `nhx` and `noy` (abondon `ndep` as used variable at some point for the input files). 

I am not sure, if the download script for the raw data should also be stored in addition to the re-gridding script. 

Closes #397


jmaerz:
Where to best store the used raw data? I would suggest to put raw data into:
`/cluster/shared/noresm/inputdata/ocn/data` there maybe having a folder `observations` for evaluation purposes and `input` (or `bnd`) for boundary conditions such as the N-deposition, etc.

TomasTorsvik:
> Where to best store the used raw data? I would suggest to put raw data into: `/cluster/shared/noresm/inputdata/ocn/data` there maybe having a folder `observations` for evaluation purposes and `input` (or `bnd`) for boundary conditions such as the N-deposition, etc.

Maybe it would be good to involve Lise (MET) and Tyge (NORCE), as they are responsible for data management in INES2 (WP8)? At least, when it comes to putting things under `/cluster/shared/noresm` it would be good to have some common structure for all model components.

jmaerz:
Hi @TomasTorsvik , I am a bit uncertain, how to best handle the storage of input/observational data - seems in general like a long lasting issue, how to best organize those things... - i.e. https://github.com/NorESMhub/NorESM/issues/27 (here for input files). How does the diagnostics package and ESMval-Tool handle this/where are observations stored? Another rabbit hole - I guess, for now, I keep it under my own project folder, where I processed the files.

TomasTorsvik:
For ESMValTool we store data on Nird only, under
`/nird/datalake/NS9560K/ESGF/obsdata/`

It is true that this is a long lasting issue. In general I would think that data that is needed as part of NorESM runs should go into `/cluster/shared/noresm/`, since this storage area is accessible from work nodes, and all other data should go somewhere in nird.

JorgSchwinger:
Regarding the "raw" data: I guess 99.99% of the data we use come with a DOI from some kind of trustworthy repository. So adding the DOI(s) of the raw-data as a global attribute to the netCDF-file should do the job in my opinion.

jmaerz:
I'll add the download script, where all those information are provided.

utils/iHAMOCC/prepare_Ndeposition/prepare_N-deposition_input.sh
```diff
@@ -0,0 +1,292 @@
...
+    #   c = 2250963.597
+    # - considering leap years (probably would be the most correct one)
+    #   s -> yr: 365.25 * 86400
+    #   kg -> kmol: 1/14.01
+    #   c: 2252505.3533190577
+    # - considering leap years, but take mol-weight of 14:
+    #   s -> yr: 365.25 * 86400
+    #   kg -> kmol: 1/14.
+    #   c: 2254114.285714286
+    # - not considering leap years and take mol-weight of 14:
```
JorgSchwinger, line 168: Would be worth noting that not considering leap years is correct, since the input4MIPs files are on a 365-day calendar. And maybe just remove the other options, since this is confusing?

jmaerz: Done. Keeping the explanations - I feel it's good to know, what could be considered - and the choice is clear.

JorgSchwinger (approved):
Looks very nice! My only general comment is that many of the resulting global attributes in the NorESM input files don't make much sense (e.g. the "contact" and "grid" attributes). If there is an easy way to remove such attributes, that might be a good idea?

TomasTorsvik (approved):

jmaerz:
Raw data is a general issue - also for observations that are just useful for evaluations. Maybe worth putting that in an extra issue and thus on the agenda.

Wrt attributes: while I see that one should indeed add some further (e.g. sometimes just the first files attributes are taken over), I wish, such attributes would be present in the other input files that we provide to iHAMOCC... I am not sure, if it makes sense to clean this up, because those are the information that allow better to trace down original input files, if at some point information for that purpose is missing... Other than that, `ncatted -O -a your_global_attribute_name,global,d,, file.nc` should be your friend (untested - use with caution).

I am currently hesitant to re-run all the stuff (a lot of hick-ups, cleaning, re-running involved these days on betzy...). Not sure, if the routine will work for `tnx0.25` and higher resolution, where likely a different approach needs to be taken.

jmaerz:
I opened an issue wrt to the data - since it is a recurring issue. 

@matsbn : with this PR, N-deposition is now available for `tnx0.5` (river & dust/fe still missing - Jörg will provide the riverine input, dust/fe tba)

PR #492, opened 2025-02-11
Abandon usage of ndep in input files, always read NHx and NOy

jmaerz:
with this PR, we abandon the usage of `ndep` in N-deposition input files. From now on, NOy and NHx are always read and used (and in case of the current default, just added together). The newly generated input files for N-deposition are prepared to support the old and new version, while as written before, one cannot expect bfb.

### DIFF
--- a/hamocc/mo_read_ndep.F90
+++ b/hamocc/mo_read_ndep.F90
@@ -184,14 +184,14 @@ contains
     real     :: fatmndep
     logical  :: first_call = .true.
 
-    ! if N-deposition is switched off set ndep to zero and return
+    ndep(:,:,:) = 0.0
+
     if (.not. do_ndep) then
-      ndep(:,:,:) = 0.0
+      ! if N-deposition is switched off return
       return
     endif
 
     if (use_nuopc_ndep) then
-
       ! If  use_nuopc_ndep, nitrogen deposition is ALWAYS obtained from the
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):
Looks fine to me

hamocc/mo_read_ndep.F90
```diff
@@ -269,7 +252,7 @@ subroutine get_ndep(kpie,kpje,kbnd,kplyear,kplmon,omask,ndep,patmnhxdep,patmnoyd
             ndep(i,j,idepnoy) = noydepread(i,j)
             ndep(i,j,idepnhx) = nhxdepread(i,j)
           else
-            ndep(i,j,idepnoy) = ndepread(i,j)
+            ndep(i,j,idepnoy) = noydepread(i,j) + nhxdepread(i,j)
```
JorgSchwinger, line 254: I think the `ndep = 0.0` in line 203 should be moved up to the start of the routine (could then also be removed in line 189). Currently it is only set to zero for `use_nuopc_ndep=.true.`

JorgSchwinger (approved):
Thanks for this - looks good!

hamocc/mo_read_ndep.F90
jmaerz: Done.

PR #494, opened 2025-02-13
Add a CODEOWNERS file to the BLOM repo

TomasTorsvik:
I would like to test having a CODEOWNERS file in the repository, to make this more transparent for non-admin users.

See https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners
for more details about CODEOWNERS file.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,21 @@
+# CODEOWNERS for NorESM-docs repository
+#======================================
+#
+# People with write permissions for the repository can create or edit the CODEOWNERS
+# file and be listed as code owners. People with admin or owner permissions can require
+# that pull requests have to be approved by code owners before they can be merged.
+#
+# A CODEOWNERS file uses a pattern that follows most of the same rules used in gitignore files.
+# The pattern is followed by one or more GitHub usernames or team names using the standard
+# @username or @org/team-name format. Users and teams must have explicit write access to the
+# repository, even if the team's members already have access.
+
+# default owners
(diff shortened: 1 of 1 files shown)


JorgSchwinger:
This means everyone who is code-owner will get a review request for every PR? We get notified anyway, so I'm not so sure this is necessary. I think it is good to be able to select who is the best person to review a specific PR. What is the advantage?

TomasTorsvik:
This would mostly be useful for someone from outside the regular team of contributors, to automatically tag people who are relevant for their PR. You need at least "triage"/"write" access on a repository in order to suggest reviewers.

We could also leave the "default owners" list empty, so that the automated tagging only kicks inn when changes are requested for either the "phy" or "hamocc" directories.

jmaerz:
As far as I understand the text, default owners aren't necessarily asked as reviewers, since the linked github pages states: 'Unless a later match takes precedence'. However, I am not sure, if other reviewers than listed in the owner file will also be allowed (or are sufficient to merge the PR) -  i.e. thinking of a PR that would be good to be e.g. reviewed by Mariana or someone else not in the current owner file. Would the list of potential reviewers remain the same or is it then fixed? Maybe worth a try and up to potential revert.

jmaerz (approved):

TomasTorsvik:
I think the documentation of the CODEOWNERS feature is a bit sparse, so the easiest would be just to enable it and try it out. It is easy to revert if we find that it doesn't work out.

Issue #479, opened 2025-02-05
HAMOCC option lkwrbioz_off = .true.  fails


JorgSchwinger:
Setting the option `lkwrbioz_off = .true.` leads to missing values in most of the fields in the output within the first month.

This test-run was done with hybrid coordinates and extended nitrogen cycle on the tn21-grid (CORE normal year forcing)

jmaerz:
Could you provide the path to your setup and the output?

JorgSchwinger:
/cluster/projects/nn2345k/schwinger/cases/NOINYOC_T62_tn21_49/

I'm going to run a short test without extended N-cycle to see if this gives us a hint. As far as I remember, we had a test run that did not fail technically at least.

JorgSchwinger:
Update: A run with `lkwrbioz_off = .true.`  and extended nitrogen cycle turned OFF is not showing missing values in the output (so far - six years completed). I'll keep this running, but from this we can conclude that something wrong is happening in the extended N-cycle is happening with  `lkwrbioz_off = .true.`.

I don't have time to investigate right now, I'll look further into this next week.

jmaerz:
Hi @JorgSchwinger , thanks for doing this exercise - a bit strange, but I will have a look into the code again to see, if there is something obvious (tmw/start of next week - being in a work shop today).

jmaerz:
Hi @JorgSchwinger , I seem not to have access to `nn2345` - any chance to grant me access so that I can have a look? I am trying to look into the code today.

JorgSchwinger:
You need to ask Mats, I cannot grant access

matsbn:
> Hi [@JorgSchwinger](https://github.com/JorgSchwinger) , I seem not to have access to `nn2345` - any chance to grant me access so that I can have a look? I am trying to look into the code today.

@jmaerz , I have given you access to nn2345k on Fram and Betzy. It might take some time for permissions to propagate in the system.

jmaerz:
@matsbn , many thanks for granting access!

jmaerz:
[...]

PR #495, opened 2025-02-14
fix bug in acclvl

jmaerz:
Hi @JorgSchwinger , I just ran the model with this bugfix and it runs through in `tnx2` without complains now (except that I still have to put a PR for issue #479, where the bug fixed with this PR additionally popped up).


JorgSchwinger (approved):
Thanks a lot for fixing this! Good to hear that it runs through now in debug mode.

jmaerz:
At least until time out (in my case 15min for 1month run time).

PR #497, opened 2025-02-17
Fix OpenMP call in mo_read_shelfmask

TomasTorsvik:
A syntax error in OpenMP calls seems to have slipped by.


JorgSchwinger (approved):
Thanks for fixing this

jmaerz (commented):
Autsch. - Thanks for fixing!

PR #502, opened 2025-02-20
Introducing optional sediment quality-based remineralization

jmaerz:
With this PR, an optional aerobic remineralization scheme is introduced to iHAMOCC that considers sediment POM age and quality to calculate the POC remineralization rate. The main -regression- formulations are based on Pika et al. 2023 while they are adjusted to explicitly account for O2-limitation and T-dependency.

**A first impression** - started from a relatively spun-up sediment state - showing the new, applied remineralization rate after one month run time - variable over time as well (as opposed to a globally fixed one - slightly varied by O2- and T-dependency in the default version - not shown):
sed_qual-first-impression

I have performed some short term simulations (months to years) starting from the relatively spun-up sediment. First seasonal responses look promising, while I still need to perform some 100 year simulations to potentially adjust the rate globally by a linear scaling factor and to have a look into how it affects iHAMOCC with a focus on the ocean shelf regions / depth<2000m.


hamocc/mo_param1_bgc.F90
```diff
@@ -484,7 +488,17 @@ subroutine init_indices()
...
+    !NOTE: The age tracer currently always needs to be the last sedlay tracer
+    !      - otherwise issues in mo_sedshi.F90!
+    if (use_sediment_quality) then
+      i_sed_age   = 1
+      issso12_age = i_sed_base + i_sed_cisonew +1
+    else
+      i_sed_age   = 0
+      issso12_age = -1
+    endif
+    nsedtra = i_sed_base + i_sed_cisonew + i_sed_age
```
TomasTorsvik: Maybe define an additional counter, `nsedtra_noage = i_sed_base + i_sed_cisonew` for use in `mo_sedshi`?

TomasTorsvik (approved):
@jmaerz - thanks, looks promising with a more dynamic sediment remineralization!

hamocc/mo_aufr_bgc.F90
```diff
@@ -26,7 +26,8 @@ module mo_aufr_bgc
 
 CONTAINS
 
-  subroutine aufr_bgc(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,kplyear,kplmon,kplday,omask,rstfnm)
+  subroutine aufr_bgc(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,kplyear,kplmon,kplday,omask,rstfnm,     &
+                     & sed_POCage,prorca_mavg)
```
JorgSchwinger: Is there a reason why `sed_POCage` and `prorca_mavg` are passed as parameters? Everything else is passed via use-statements.

If I'm correct these are also not needed inside this routine - see my comment below

hamocc/mo_aufr_bgc.F90
```diff
@@ -649,6 +677,26 @@ subroutine aufr_bgc(kpie,kpje,kpke,ntr,ntrbgc,itrbgc,trc,kplyear,kplmon,kplday,o
...
+              prorca_mavg2(i,j,2)             = prorca_mavg(i,j)
+              do k=1,ks
+                sedlay2(i,j,k,issso12_age)    = sed_POCage(i,j,k)
+                sedlay2(i,j,ks+k,issso12_age) = sed_POCage(i,j,k)
+              enddo
+            endif
+          enddo
+        enddo
+      endif
+    endif
```
JorgSchwinger, line 695: The copying into the `burial2` and `sedlay2`fields is happening at the end of `hamocc_init`, and I would suggest to move the `prorca_mavg2` part also there. Then this can be deleted here.

hamocc/mo_param1_bgc.F90
JorgSchwinger: or define the age tracer as a separate array?

JorgSchwinger (commented):
In general looks good, but there is an issue how the initialisation is done in `aufr_bgc`(see comment there). Also I don't really like that `sedlay` and `burial`is now initialized in two different routines, but admittedly, `ini_fields_ocean` and `ini_sediment` have been a bit blurry already before.

hamocc/mo_aufr_bgc.F90
jmaerz: Hi @JorgSchwinger , yes, since they are local variables in `mo_hamocc_init`, I didn't want to introduce `sedlay` and `burial` as use statements to avoid confusion - but if that's your preferred choice, I can change it.

hamocc/mo_aufr_bgc.F90
jmaerz: You mean placing all these loops in `mo_hamocc_init` instead? - The copying currently carried out at the end is doing something different than what is done here.

hamocc/mo_param1_bgc.F90
jmaerz: I was thinking about adding an extra field, but it adds a lot of unnecessary code to introduce it. If it's ok, I consider Tomas idea - was also considering it before, but thought a local adjustment in sedshi is easier to follow (which is arguable).

hamocc/mo_aufr_bgc.F90
JorgSchwinger: No, not really. There, the whole `burial` array is copied into bot time-levels of `burial2`, same for `sedlay`, why is that different?

hamocc/mo_param1_bgc.F90
JorgSchwinger: I'm fine with either solution, I don't think this is critical

hamocc/mo_aufr_bgc.F90
jmaerz: Since the one in `mo_aufr_bgc` is at restart (e.g. for a hybrid run), when the information is not there, while the stuff in `mo_hamocc_init` is for cold start, as far as I understand. One could use the field `prorca_mavg`,  as use statements though, that's true (while one would either make the init fields public or include `sedlay` and `burial`, or introduce another field in `mo_sedmnt` just for the initial values, not sure, what to prefer).

JorgSchwinger: The logic applied in HAMMOC is that the initial values are read in any case whether cold start or restart. If restart, `aufr_bgc` overwrites if something is actually in the restart. You initialize the fields in `ini_sediment`. If there is nothing in the restart file, then it is not necessary to repeat the initialization in `aufr_bgc`, and the copying over in the `burial2`, `sedlay2`, ... arrays is done elsewhere

jmaerz: Nope - I thought so, too, but failed - the copying in that way is needed.

jmaerz: Ok, clarified via short in person chat - thanks for the comments!

jmaerz: Done (and fixed a small bug by this) - now using use statements for `sedlay`, `burial` etc. instead of passing a function argument. Current default is/remains that an initialization file is required at cold start and hybrid start without previous usage of sediment quality.

jmaerz:
Ok, I resolved the issues. Anything else?

JorgSchwinger (approved):
Sorry for the lengthy discussion, took me some time to understand this correctly.

jmaerz:
Hi @TomasTorsvik , I would merge this - since there was the discussion wrt github squash - any new preferences in that respect?

jmaerz:
@JorgSchwinger , no worries - sometimes, it is just easier to explain things in person. Glad we were able to clarify things.

TomasTorsvik:
@jmaerz - for the moment I suggest that all merges to the `mater` branch to be made as a regular "merge pull request". When we use git submodules/fleximod in NorESM, the file changes in the components become part of the git commit history (see e.g. https://github.com/NorESMhub/NorESM/pull/642/files). There is a discussion if want to keep this behavior (https://github.com/NorESMhub/NorESM/discussions/630). For now, it is perhaps best if we do the merge commits.

PR #504, opened 2025-02-21
Deallocate arrays in mo_netcdf_bgcrw

TomasTorsvik:
This PR will not make any difference, but maybe it is still good practice to explicitly deallocate variables? At least, I find it confusing  when only some of the temporary arrays are deallocated.

### DIFF
--- a/hamocc/mo_netcdf_bgcrw.F90
+++ b/hamocc/mo_netcdf_bgcrw.F90
@@ -316,8 +316,11 @@ contains
       allocate(istart(ndims))
       allocate(icount(ndims))
       allocate(arr_l(ii,jj,klev))
+      allocate(arr_g1(itdm,jj,klev))
 
       arr_l=0.0
+      arr_g1=0.0
+
       if (klev.gt.1.or.time.gt.0) then
         if (klev.gt.1.and.time.gt.0) then
           istart(3)=1
@@ -353,8 +356,6 @@ contains
         enddo
       enddo
 
-      allocate(arr_g1(itdm,jj,klev))
-      arr_g1=0.0
       call xcgetrow(arr_g1, arr_l, klev)
 
       ncstat=nfmpi_inq_varid(ncid,desc,ncvarid)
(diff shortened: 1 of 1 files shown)


hamocc/mo_netcdf_bgcrw.F90
```diff
@@ -378,7 +378,7 @@ subroutine write_netcdf_var(ncid,desc,arr,klev,time)
       !   stop '(write_pnetcdf_var)'
       ! endif
 
-      deallocate(istart,icount,arr_g1)
+      deallocate(istart,icount,arr_l,arr_g1)
```
jmaerz, line 382: Hi @TomasTorsvik , I am not sure, if I agree here to put in the `arr_l`, since from Fortran, `deallocate` seem to expect the object to be allocated, if I am not mistaken (and you deallocate it above, if I see it correct).

jmaerz (approved):
Hi @TomasTorsvik , do we have a chance to check for memory leaks when running BLOM/iHAMOCC standalone? - to check, whether we really need to go through the code to avoid memory leakages (I vaguely remember that Mariana reported some for NorESM - but how about BLOM?). In any case, thanks for looking into it.

hamocc/mo_netcdf_bgcrw.F90
TomasTorsvik: The allocation of `arr_l` is different depending on IOTYPE==0/1, and the allocation is placed inside an `if ... else if ... endif` block. Hence the scope of the allocation should not reach outside this if-block, so there is a `deallocate`for each of them.

Because the `PNETCDF` CPP flag determines if the `IOTYPE==1` block is populated, there is a potential that the part after `else if`is empty. Hence we could introduce an error of the deallocation of `arr_l` is pulled outside the scope of the if-block in this case.

JorgSchwinger (approved):

TomasTorsvik:
Hi @jmaerz - I haven't checked recently for OMIP-type simulations. I haven't done any long ones in a while.
You can check, if you have a longer run, in the `cpl.log` file, grep for `memory_write` and see if there is an increase in the usage over time.

PR #505, opened 2025-02-21
Add river input file for tn0.5v1 grid

JorgSchwinger:
This PR adds support for a river input file on the tnx0.5v1 grid. The file has been created and copied in the input data directory on betzy. I have also added the scripts used to create the river inputs to the utils-dirctory. These scripts are not optimal (and in matlab/octave...), but I think we agreed that it is better to collect them.


TomasTorsvik (approved):
@JorgSchwinger - thanks, great to have the scripts there as well!

matsbn (approved):
Looks good. Thanks @JorgSchwinger!

jmaerz (approved):
Hi @JorgSchwinger , thanks for putting in the scripts (and data) as well - greatly appreciated! Looks like input data for `tnx0.5` is now fully available for the default settings!

Issue #411, opened 2024-10-02
ndep forcing file is incorrect for most compsets using DATM using MCT


mvertens:
The MCT coupling interface will use the configuration variable BLOM_NDEP_SCENARIO to determine the NDEP forcing file. However, BLOM_NDEP_SCENARIOonly uses CAM60%NORESM and results in 1850 forcing regardless of the first 4 digit forcing date. In particular, a 2000_DATM*_BLOM compset will pick up ndep forcing for 1850. Furthermore CAM60 will no longer be used in noresm2.5 - so again 1850 forcing will result. Although we are moving away from BLOM_NDEP_SCENARIO in noresm2.5 (since ndep will always be obtained from datm or cam) - comparisons will show large differences for incorrect settings of BLOM_NDEP_SCENARIO.

```
<entry id="BLOM_NDEP_SCENARIO">
    <type>char</type>
    <valid_values>UNSET,1850,2000,hist,ssp119,ssp126,ssp245,ssp370,ssp434,ssp460,ssp534os,ssp585</valid_values>
    <default_value>1850</default_value>
    <values>
      <value compset="HIST_CAM60%NORESM.*_BLOM%ECO"    >hist</value>
      <value compset="20TR_DATM.*_BLOM%ECO"            >hist</value>
      <value compset="2000_CAM60%NORESM.*_BLOM%ECO"    >2000</value>
      <value compset="SSP119_CAM60%NORESM.*_BLOM%ECO"  >ssp119</value>
      <value compset="SSP126_CAM60%NORESM.*_BLOM%ECO"  >ssp126</value>
      <value compset="SSP245_CAM60%NORESM.*_BLOM%ECO"  >ssp245</value>
      <value compset="SSP370_CAM60%NORESM.*_BLOM%ECO"  >ssp370</value>
      <value compset="SSP370LOWNTCF_CAM60%NORESM.*_BLOM%ECO"      >ssp370</value>
      <value compset="SSP370REFGHGLOWNTCF_CAM60%NORESM.*_BLOM%ECO">ssp370</value>
      <value compset="SSP434_CAM60%NORESM.*_BLOM%ECO"  >ssp434</value>
      <value compset="SSP460_CAM60%NORESM.*_BLOM%ECO"  >ssp460</value>
      <value compset="SSP534_CAM60%NORESM.*_BLOM%ECO"  >ssp534os</value>
[...]

Issue #404, opened 2024-09-16
new requirements for dust deposition and riverine input for new stream functionality


mvertens:
In order to use the CDEPS stream functionality for forcing data - the time coordinate needs to be of the form
```
	double time(time) ;
		time:long_name = "time" ;
		time:units = "days since 0001-01-01 00:00:00" ;
		time:calendar = "noleap" ;
```
(since this is a climatology - noleap should be used)

However, looking at the file /cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_mhw2006_tnx1v4_20171107.nc
the time coordinate is 
```
	double time(time) ;
		time:standard_name = "time" ;
		time:units = "day as %Y%m%d.%f" ;
		time:calendar = "proleptic_gregorian" ;
```
and the values are all zero
```
$ ncdump -ff -v time /cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_mhw2006_tnx1v4_20171107.nc
 time = 0,   // time(1)
    0,   // time(2)
    0,   // time(3)
    0,   // time(4)
    0,   // time(5)
    0,   // time(6)
    0,   // time(7)
    0,   // time(8)
    0,   // time(9)
    0,   // time(10)
    0,   // time(11)
    0;  // time(12)
    }
```

Also - the format needs to be cdf5 not classic
```
$ ncdump -k /cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_mhw2006_tnx1v4_20171107.nc
classic
```
To change this you simply need to invoke
`nccopy -k cdf5 oldfile newfile`

mvertens:
@monsieuralok - I see that you created the following files:
```
-rw-rw-rw-. 1 agu002 noresm   2224868 May  6  2013 /cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_mhw2006_tnx2v1_20130506.nc
-rw-rw-rw-. 1 agu002 noresm  15524700 Nov  7  2017 /cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_mhw2006_tnx1v4_20171107.nc
-rw-r--r--. 1 agu002 noresm 185957276 Oct  4  2018 /cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_mhw2006_tnx0.25v4_20181004.nc
[...]

PR #506, opened 2025-02-24
Fix N-cycle coupling

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , cc @DirkOlivie ,

with this PR, a bugfix is introduced to the mct-coupling of NH3, NOy and NHx fields (likely introduced during the BLOM code restructuring - code to transfer the fields wasn't reached any more). Further, I renamed the switch for providing the coupler fields for nitrogen deposition to iHAMOCC, since it can be used by both couplers (MCT and NUOPC) as being currently under testing. Hence, with this PR, iHAMOCC `master` is again fully capable to handle NH3,N2O, NHx and NOy gas and deposition fluxes. 
The setup allows to be used in current test/N-cycling coupling setups (not yet fully supported in NorESM(X), while we are working on this issue as well).

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3870,7 +3870,7 @@
     <values>
       <value>.false.</value>
       <value comp_interface="nuopc">.true.</value>
-      <value "hamocc_atmndepc"="yes">.true.</value>
+      <value hamocc_atmndepc="yes">.true.</value>
     </values>
     <desc>if .true., use MCT coupler fields from CAM or NUOPC to obtain NDEP (always imported to model).</desc>
   </entry>
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3863,15 +3863,16 @@
    <desc>File name (incl. full path) for atmopheric N-deposition data</desc>
   </entry>
 
-  <entry id="use_nuopc_ndep">
+  <entry id="use_coupler_ndep" modify_via_xml="HAMOCC_ATMNDEPC">
     <type>logical</type>
     <category>config_bgc</category>
     <group>config_bgc</group>
     <values>
       <value>.false.</value>
       <value comp_interface="nuopc">.true.</value>
+      <value "hamocc_atmndepc"="yes">.true.</value>
     </values>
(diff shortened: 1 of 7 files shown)


cime_config/namelist_definition_blom.xml
```diff
@@ -3863,15 +3863,16 @@
...
 
-  <entry id="use_nuopc_ndep">
+  <entry id="use_coupler_ndep" modify_via_xml="HAMOCC_ATMNDEPC">
     <type>logical</type>
     <category>config_bgc</category>
     <group>config_bgc</group>
     <values>
       <value>.false.</value>
       <value comp_interface="nuopc">.true.</value>
+      <value hamocc_atmndepc="yes">.true.</value>
```
TomasTorsvik, line 3873: How will this work now - the `use_coupler_ndep` is `.false.` by default, but `.true.` if either `comp_interface="nuopc"` or `hamocc_atmndepc="yes"`?
Have you tested what happens if `comp_interface="nuopc"` and `hamocc_atmndepc="no"`? I guess the entry will still be `.true.` in this case.

TomasTorsvik (approved):
@jmaerz - I think this should be fine, although I'm not up to speed with regards to switching between MCT and NUOPC couplers.

cime_config/namelist_definition_blom.xml
jmaerz: Hi @TomasTorsvik , good question. In the setup that I have currently available, I went though all your suggested combinations (via `xmlchange` and `preview_namelist`) and it works as expected - so if `comp_interface='nuopc'` and `hamocc_atmndep='no'`, `use_coupler_ndep=.true.`, if both are set, it results also in `.true.` and if `comp_interface=mct` and `hamocc_atmndep='yes'`, `use_coupler_ndep` is alo `.true.` (the latter now also tested in a short coupled simulation).

jmaerz: In any case, it's currently not really safe using nuopc-coupled nitrogen deposition fields, since there still seems to be a persistent issue: #411 (and similar for dust #404) - I haven't looked into those thus far (and probably won't the next weeks).

jmaerz:
I'll wait until tmw and will merge it then.

PR #507, opened 2025-02-26
Corrected and updated processor layouts for tnx0.5v1 grid.

matsbn:
Fixed issues with the processor layouts for the tnx0.5v1 grid that would lead to an error with message "xcspmd: arctic empty tiles are not twins".


TomasTorsvik (approved):
@matsbn - thanks for updating, I suppose this should be fine.

PR #511, opened 2025-03-10
Fix sediment diagnostics and sediment quality

### LINKED ISSUE #512: Set consistent calendar option in iHAMOCC
In most places where it is relevant, iHAMOCC assumes a 365 day per year or `noleap` calendar, as is used by default in `NorESM`. However, it is possible to select other calendar options as well, in particular a 360 day per year calendar used for some test cases.

It would be useful if iHAMOCC would be able to handle the different calendar options somewhat gracefully, maybe with a warning if someone use an unexpected calendar setting.

jmaerz:
I realized the issue, why diagnosed burial fluxes (and some extended nitrogen cycle diagnostics) were overestimated/not correctly estimated - now setting those at each time step to zero, before performing calculations (i.e. burial fluxes are only calculated once per day and not per time step, which overestimated the burial fluxes by factor 24 - similarly, rates for processes that only happen below certain thresholds, etc.). In that process, I also updated the sediment quality calculations to account for the sediment acceleration scheme and in general fixed the burial POC age (previously underestimated). 

I set up a run - and at least the diagnostics now look reasonable. Wrt sediment quality, it remains still a bit experimental, while promising - we can discuss details in person, if needed.

### DIFF
--- a/hamocc/mo_param_bgc.F90
+++ b/hamocc/mo_param_bgc.F90
@@ -134,9 +134,10 @@ module mo_param_bgc
   !********************************************************************
   ! Time parameters
   !********************************************************************
-  real, parameter :: sec_per_day = 86400. ! seconds per day
-  real, protected :: sec_per_year         ! seconds per year
-  real, protected :: days_per_year        ! simulated days per year
+  ! NOTE: days_per_year and sec_per_year are updated in ini_bgctimes(!)
+  real, parameter :: sec_per_day   = 86400.           ! [s/d]  seconds per day
+  real, protected :: days_per_year = 365.             ! [d/yr] days per year
+  real, protected :: sec_per_year  = 365.*sec_per_day ! [s/yr] seconds per year
 
   !********************************************************************
   ! Stoichiometry and fixed parameters
--- a/hamocc/mo_param_bgc.F90
+++ b/hamocc/mo_param_bgc.F90
@@ -523,32 +523,17 @@ module mo_param_bgc
   real :: beta13, alpha14, d14cat, d13c_atm
 
 contains
+
   !********************************************************************
-  subroutine ini_bgctimes(calendar)
+  subroutine ini_bgctimes(nday_in_year)
 
-    character(len=*),intent(in) :: calendar
+    integer,intent(in) :: nday_in_year
 
-    select case (trim(calendar))
-      ! see mod_calendar.F90 and mod_time.F90
-      case ('noleap','365_day')
-        days_per_year = 365.
-      case ('all_leap','366_day')
-        days_per_year = 366.
-        if (mnproc == 1) then
-          write (io_stdo_bgc,*) 'WARNING: Init iHAMOCC time variables: non-standard 366_day calendar selected'
-        endif
-      case ('360_day')
-        days_per_year = 360.
-        if (mnproc == 1) then
-          write (io_stdo_bgc,*) 'WARNING: Init iHAMOCC time variables: non-standard 360_day calendar selected'
-        endif
-      case default
-        if (mnproc == 1) then
-          write (io_stdo_bgc,*) 'Init iHAMOCC time variables: calendar not supported'
-        endif
-        call xcstop('(init_iHAMOCC_timevars)')
-             stop '(init_iHAMOCC_timevars)'
-    end select
+    days_per_year = real(nday_in_year)
+
+    if (nday_in_year /= 365 .and. mnproc==1) then
+          write (io_stdo_bgc,*) 'WARNING: Init iHAMOCC time variables: non-standard calendar selected with [days] ',days_per_year
+    endif
 
     sec_per_year = days_per_year*sec_per_day
   end subroutine ini_bgctimes
--- a/hamocc/mo_param_bgc.F90
+++ b/hamocc/mo_param_bgc.F90
@@ -138,7 +138,7 @@ module mo_param_bgc
   real, parameter :: sec_per_day   = 86400.           ! [s/d]  seconds per day
   real, protected :: days_per_year = 365.             ! [d/yr] days per year
   real, protected :: sec_per_year  = 365.*sec_per_day ! [s/yr] seconds per year
-
+  logical         :: lini=.true.
   !********************************************************************
   ! Stoichiometry and fixed parameters
   !********************************************************************
@@ -526,13 +526,14 @@ contains
 
   !********************************************************************
   subroutine ini_bgctimes(nday_in_year)
-
+    ! NOTE: called also at run time after initialization
     integer,intent(in) :: nday_in_year
 
     days_per_year = real(nday_in_year)
 
-    if (nday_in_year /= 365 .and. mnproc==1) then
-          write (io_stdo_bgc,*) 'WARNING: Init iHAMOCC time variables: non-standard calendar selected with [days] ',days_per_year
+    if (nday_in_year /= 365 .and. mnproc==1 .and. lini .eqv. .true.) then
+      write (io_stdo_bgc,*) 'WARNING: Init iHAMOCC time variables: non-standard calendar selected with [days] ',days_per_year
+      lini=.false.
     endif
 
     sec_per_year = days_per_year*sec_per_day
--- a/hamocc/mo_param_bgc.F90
+++ b/hamocc/mo_param_bgc.F90
@@ -40,7 +40,7 @@ module mo_param_bgc
                             do_n2onh3_coupled,use_extNcycle,                                       &
                             lkwrbioz_off,lTO2depremin,use_shelfsea_res_time,use_sediment_quality,  &
                             use_pref_tracers,use_coupler_ndep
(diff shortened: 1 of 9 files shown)


TomasTorsvik (approved):
@jmaerz - thanks for fix, looks fine to me!

hamocc/mo_sedshi.F90
```diff
@@ -59,9 +61,15 @@ subroutine sedshi(kpie,kpje,omask)
     real    :: sedlo,uebers,seddef,spresent,buried
     real    :: refill,frac
     real    :: eps=epsilon(1.)
+    real    :: acc_time=0.
 
     sedfluxb(:,:,:) = 0.
 
+    if(do_sedspinup .and. kplyear>=sedspin_yr_s .and. kplyear<=sedspin_yr_e) then
+      ! accumulated time spent due to sediment acceleration
+      acc_time = 86400.*sedspin_ncyc/31104000. ! *dtbgc/dtbgc
```
JorgSchwinger: What is the number `31104000`? A comment explaining this would be nice.

JorgSchwinger: Also, this depends on the fact that `sedshi` is called once per model day, right? Maybe it would be good to add a short comment in `hamocc4bcm` line 364 that there is a dependency on this choice in the sediment code.

JorgSchwinger (approved):

hamocc/mo_sedshi.F90
jmaerz: Hi @JorgSchwinger , maybe the number is incorrect? - it's a model year with 360 days in seconds - but maybe better refer to 365 days? (How is it handled in BLOM/NorESM? - naive question... - shall we maybe better introduce a parameter for this in `mo_param_bgc`?). Yes, it's due to the fact that `sedshi` is only called once per day, as far as I understand from `hamocc4bcm`.

JorgSchwinger: If the number is meant to represent seconds in a model year, then it should be `86400*365`. Is this number used more often in the code? Then it might really be worth to replace it by a parameter in `mo_param_bgc`

TomasTorsvik: It could be useful to include parameters for both 
```
seconds_in_day = 86400.
seconds_in_year = 86400.*365.
```
However, I see also references to different calendars in `phy/mod_calendar.F90`, so I'm not sure if `seconds_in_year` should be a fixed parameter.

TomasTorsvik: The calendar is defined in `phy/mod_time.F90` : ` init_timevars`. The `cesm` case is run with `noleap` = 365.
Other cases use different calendars.

JorgSchwinger: I don't think we will use something different than a 365 day calendar (at least not in the near future). One could hard code it in HAMOCC, query the BLOM calendar in hamocc_init and stop if the calendar type is not 365.

TomasTorsvik: The `fuk95` and `single_column` test cases use `360_day` calendar, so these will not work if the 365 day calendar is hard coded. We usually run these only for short tests, so maybe a warning is sufficient.

I suppose some of the input files assume a 365 day calendar. Not sure what is the best approach here, but I can see that we can run into trouble if we allow different calendar options.

TomasTorsvik: I propose to make a new issue / PR regarding calendar setting in iHAMOCC, not to bundle everything into this PR.

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , I just included some routine to handle the question. The only calendar that I feel issues with is the `standard` calendar, which seem to handle leap versus non-leap years - I wonder, if we need to go to this detail right now or if it is ok to stop the simulation (or set the days to 365 or 365.25). Opinions?

In any case, I see that some parameter settings rely on `sec_per_day` and on `sec_per_year` - for the latter, I don't know, if adjustments are really needed - I expect not. However, for e.g. nitrogen deposition, etc. the adjusted values for the year could be used as well (looks like only a few occasions, where `86400` and `365` appear in the code, which could now be easily exchanged).

hamocc/mo_param_bgc.F90
```diff
@@ -513,6 +522,29 @@ module mo_param_bgc
...
+  !********************************************************************
+  subroutine ini_bgctimes(calendar)
+
+    character(len=*),intent(in) :: calendar
+
+    select case (trim(calendar))
+      ! see mod_calendar.F90 and mod_time.F90
+      case ('noleap','365_day')
+        days_per_year = 365.
+      case ('all_leap','366_day')
```
TomasTorsvik:
```suggestion
      case ('all_leap','366_day')
      if (mnproc == 1) then
         write (io_stdo_bgc,*) 'WARNING: Init iHAMOCC time variables: non-standard 366_day calendar selected'
      endif      
```

TomasTorsvik (approved):

hamocc/mo_param_bgc.F90
TomasTorsvik: I suggest to include a warning message if anythin other than 'noleap' is selected.

jmaerz: I'll do so. Sorry, I overlooked your previous comment on doing all this in a different PR (was already working on it...) - shall I try to split it up or is it ok to proceed?

hamocc/mo_param_bgc.F90
```diff
@@ -128,6 +129,14 @@ module mo_param_bgc
...
           & max_limiter
 
+  ! Time variables
+  public :: sec_per_year,sec_per_day,days_per_year
+  !********************************************************************
+  ! Time parameters
+  !********************************************************************
+  real, parameter :: sec_per_day = 86400. ! seconds per day
+  real, protected :: sec_per_year         ! seconds per year
+  real, protected :: days_per_year        ! simulated days per year
```
JorgSchwinger: Can we set default values also for `sec_per_year` and `days_per_year`, such that something is used even if `ini_bgctimes` is not called?

jmaerz: Propose some - based on a 365 year?

jmaerz: Thus far, we're rather free to start off with some idea, how to best handle it in the future. In any case, `ini_bgctimes` is called early in in the initialization by purpose...

hamocc/mo_param_bgc.F90
TomasTorsvik: Since you are already working on it, it is probably best to proceed, unless the rabbit hole goes too deep ...

hamocc/mo_param_bgc.F90
jmaerz: I.e. I only see very minor issues in the parameter definition - where a few `365` are hard-coded (while I am not seeing a need to change this).

JorgSchwinger: > Propose some - based on a 365 year?

Yes, that would make sense.

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , I would start a short testrun with the current branch. If that results in reasonable results, I would merge it as is (thus far only tested in the single column mode).

Afterward, we can discuss, in how far we should exchange the `365` in the N-deposition, oafx and river-applications.

matsbn:
Note that ``nday_in_year`` can change during a simulation (not the case yet in NorESM), so maybe it should not only be set during initialization?

jmaerz:
@matsbn , what do you mean by "it can change" during a simulation? - with the latest change, I only make use of it once at initialization (when calling `hamocc_init`) to set some values in iHAMOCC. Would that somehow be affected by what you say?

jmaerz:
Ok, you're probably referring to restarts and leap years, etc. - which would then necessitate to update the value throughout run time?

JorgSchwinger:
Not only restarts, but if it is a Gregorian calendar, nday_per_year changes every 4th year.

I don't know if there are plans to support such calendar with NorESM anytime soon. So for now maybe just issue an error message and stop if calendar is set to  Gregorian?

jmaerz:
Hi @JorgSchwinger , while I see an issue for the initial parameter settings (where I wouldn't change the parameter according to year length), I don't see good reasons, why not updating the time accordingly. I now update the time also at runtime (while admittedly the warning is currently not taking always effect in case of non- versus leap years - up to potential change). 
In terms of `days_per_year`: for the current purpose (that might also be up to change), it doesn't make a huge difference. For the deposition reading, etc. we still need to discuss, how to treat it (beyond the scope of this PR - debating currently to branch that time issue completely out, since I would like to have the fix for diagnostics in soon). So technically, we could also hard-set it to 365 at the moment and then debate, if we really need to take different calendars into account.

jmaerz:
since the time-issue developed into a longer discussion on the purpose, I now reverted back some stuff and will open a new PR on this question to address #512 . I merge now these changes that address the header of this PR and will link to this discussion when opening the new PR (after merging this one).

PR #514, opened 2025-03-12
Calendar dependency of HAMOCC processes

### LINKED ISSUE #512: Set consistent calendar option in iHAMOCC

jmaerz:
this PR aims at tackling issue #512 and is a follow-up on the discussion started with #511. 

I guess, we have to clarify a bit, which purpose we have with introducing a dependency on the calendar of BLOM. Related questions:

- do we adjust e.g. deposition input according to the year length?
- do we need to adjust rates (I would expect not)?

The current use case for sediment quality might be up to change though - while I feel it is good to have a general discussion on the question, whether to support different calendar types in iHAMOCC.

Closes #512


jmaerz:
Since #512 evolves into a longer discussion on supported calendar types: do we really envisage to fully support calendars other than 365 days? - if so, we may reconsider input-file units (i.e. towards units of e.g (k)mol/m2/s instead of other units). This would potentially also come at the cost of regenerating input files (again) - as long as CDEPS isn't fully supported.

Another part, where the 365 days may be involved is in the output routines (checking for 365 days frequency, if I am not mistaken... - at least that's how I understand the namelist settings - I suspect that also would need an update).

TomasTorsvik (approved):

TomasTorsvik:
I think there is no immediate need to support calendars other than "noleap" for NorESM configurations. I would prefer to keep things as simple as possible until there is an actual demand for other alternatives.

jmaerz:
@TomasTorsvik and @JorgSchwinger , shall we somehow continue the calendar dependency or shall we scrap this PR? - I wonder, if updating the branch is worth it. It's now lingering around for a while, waiting for some input.

TomasTorsvik:
@jmaerz - I would like to revive this PR, since at least it replaces hard-coded values throughout the code with pre-set parameters, which is an improvement. I don't think we need to solve all calendar issues in this PR.

JorgSchwinger:
In my opinion the need to support different calendars is limited. Maybe in addition to the 365-day calendar the greogian calendar, since this is what happening in reality. For other calendars (e.g. 360-day) I don't see the need.

On the other hand, I am not sure if the treatment of input files needs more thinking, when support for a gregorian calendar is added? 

One option would be to merge this now, but issue an error (instead of a warning) when `days_in_year /= 365`, and then take the rest later (post NorESM3)?

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , ok, I will take care of it the next days (bringing it up-to-date) - a warning is currently already implemented for non-365 day calendars in the subroutine `ini_bgctimes`. I'll re-check what I did and let you know once it is (again) ready for merging.

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , I brought the calendar dependency up-to-date and issue now an error (and a model halt) in case of non-`single_column` runs - in case of the latter, just the error is issued so that the `single_column` is still useful for testing (as in running through without stopping). Not sure, if this exception is needed for other testcases as well (i.e. `fuk95`).

TomasTorsvik:
@jmaerz - it seems all idealized cases (`single_column`, `fuk95`, `channel`) use the `360_day` calendar (see `init_timevars` in `phy/mod_time.F90`). We only run `single_column` and `fuk95` as part of regular tests, so an exception for these cases would be useful.

jmaerz:
Hi @TomasTorsvik , I added now exceptions also for those cases (wasn't sure, if they are at all compiled with iHAMOCC, which is why I didn't add them in first place).

TomasTorsvik:
@jmaerz - Thanks! I don't think these cases are compiled with iHAMOCC by default, but maybe we will make use of them some day.

hamocc/mo_param_bgc.F90
```diff
@@ -520,6 +530,27 @@ module mo_param_bgc
...
+  subroutine ini_bgctimes(nday_in_year)
+
+    use mod_config, only: expcnf
+    ! NOTE: called also at run time after initialization
+    integer,intent(in) :: nday_in_year
+
+    days_per_year = real(nday_in_year)
+
+    if (nday_in_year /= 365 .and. mnproc==1 .and. lini .eqv. .true.) then
+      write (io_stdo_bgc,*) 'Error: Init iHAMOCC time variables: non-standard calendar selected with [days] ',days_per_year
```
JorgSchwinger: Sorry for being picky... but it would be nicer to write "Error: ..." only if it is treated as an error. Other wise it would be "Warning: ...". So maybe make a write statement for both cases using a if-then-else-block?

JorgSchwinger (approved):
Thanks for taking care of this!

hamocc/mo_param_bgc.F90
jmaerz: Done.

PR #515, opened 2025-03-12
Remove CGS units support

matsbn:
Support for CGS units is removed, except for continuation from restart files written in CGS units.

Bit identical results have been confirmed with the NOIIAJRAOC compset and the tnx1v4 grid for the following cases:
- Simulation from initial conditions with BLOM_VCOORD = "cntiso_hybrid".
- Simulation from initial conditions with BLOM_VCOORD = "isopyc_bulkml"
- Continuation from restart in MKS units.
- Continuation from restart in CGS units.


jmaerz:
Hi @matsbn , could you briefly explain how to perform a simulation starting from cgs restart files?

matsbn:
> Hi @matsbn , could you briefly explain how to perform a simulation starting from cgs restart files?

It is exactly the same as starting from a restart file in MKS units (either a branch run, hybrid run or specifying the restart file as the ocean initial condition by setting the ICFILE namelist variable). The unit system used for the restart file is automatically detected. As it is stated in mod_restart.F90:

"The ratio of bottom pressure to bottom geopotential should approximate sea water density that robustly can be tested if MKS or CGS units was used in the restart file."

milicak (approved):

TomasTorsvik (approved):

jmaerz (approved):
Thanks for the explanation. From the iHAMOCC side, it looks good - I haven't checked all details in BLOM, but expect that it has been tested. Thanks!

JorgSchwinger (approved):

PR #513, opened 2025-03-12
Bring rpointer changes to BLOM

mvdebolskiy:
Added rpointer-timestamps handling for BLOM with NUOPC.
MCT and standalone should be backwards compatible.

`ERR_D_Ld5.TL319_tn14.NOIIAJRAOC.betzy_intel` and `ERI_D_Ld5.TL319_tn14.NOIIAJRAOC.betzy_intel`.

Tests to be ran apart from `aux_blom_noresm` test-suite: 
 - [x] standalone
 - [X]  ERR,~~ERI~~ tests with MCT (mct not available  for testing or case-creation with current cime)
    - ERI test does not work for `noresm2_3_alpha03`. I was able to run after unzipping files manually
 - [x]  ERR, ERI tests with multiple blom instances. 
 - [x] Run a hybrid run from an older blom run with old rpointers.
 - [x] Hybrid case with multiple blom instances starting from a case with one instance (old rpointers and new)

@TomasTorsvik @matsbn Can you help with the rest of the tests? Running NINST>0 even with v1.7.1.4. gives me errors

PR #517, opened 2025-03-14
Add sed_denit to namelist

jmaerz:

this PR just adds the `sed_denit` rate to the namelist - making it tunable. It would be great, if we could merge this before the cgs and pointer PR (#513, #515)


JorgSchwinger (approved):
Looks good.

PR #519, opened 2025-03-17
Include additional tests with GNU compiler

TomasTorsvik:
This PR expands the list of default system tests for BLOM, including tests using the GNU compiler.


mvdebolskiy:
The only caviat here is that when we test for `noresm2_3`, you have to specify `intel` as a compiler for the test suite.

TomasTorsvik:
This PR does not touch any source code, so I will go ahead and merge it before the #513 and #515 changes come in, and make a new release tag `v1.7.2`.

PR #522, opened 2025-03-20
change N-cycle tendencies units from mol/m3/s to mol/m2/s by includin…

### LINKED ISSUE #521: Volumetric rate output for the extended nitrogen cycle tendencies
Calculate volumetric rate output (rate*\DeltaV or *dz) for tendencies instead of current rate for extended nitrogen cycle diagnostics.  

This shall improve the issue discussed in https://github.com/NorESMhub/BLOM/pull/498#issuecomment-2740602505 

Ideally, this is accompanied by making some currently extended nitrogen cycle-specific outputs available to the default model version as well to enable e.g better analysis of the sediment processes.

jmaerz:
…g layer thickness information. Addresses: https://github.com/NorESMhub/BLOM/pull/498#issuecomment-2740602505. And fixing a diagnostics issue in case of `kwrbioz_off=True`

I branched from the second last PR before cgs was removed - not sure, if we shall do some v1.7.2 release candidate, where this can be merged in...

(Partially)
Closes #521

### DIFF
--- a/hamocc/mo_extNwatercol.F90
+++ b/hamocc/mo_extNwatercol.F90
@@ -90,7 +90,7 @@ contains
     real    :: Tdepanh4,O2limanh4,nut1lim,anh4new,potdnh4amox,fdetamox,fno2,fn2o,ftotnh4
     real    :: Tdepano2,O2limano2,nut2lim,ano2new,potdno2nitr,fdetnitr,ftotno2,no2fn2o,no2fno2,    &
                no2fdetamox
-    real    :: amoxfrac,nitrfrac,totd,amox,nitr,temp
+    real    :: amoxfrac,nitrfrac,totd,amox,nitr,temp,dz
 
     ! Set output-related fields to zero
     nitr_NH4      = 0.
@@ -102,7 +102,7 @@ contains
     !$OMP PARALLEL DO PRIVATE(i,k,Tdepanh4,O2limanh4,nut1lim,anh4new,potdnh4amox,fdetamox,fno2,    &
     !$OMP                     fn2o,ftotnh4,Tdepano2,O2limano2,nut2lim,ano2new,potdno2nitr,fdetnitr,&
     !$OMP                     ftotno2,amoxfrac,nitrfrac,totd,amox,nitr,temp,no2fn2o,no2fno2,       &
-    !$OMP                     no2fdetamox)
+    !$OMP                     no2fdetamox,dz)
     do j = 1,kpje
       do i = 1,kpie
         do k = 1,kpke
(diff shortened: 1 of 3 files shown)


JorgSchwinger:
To be honest, I do not understand the purpose of this. Why is it better to have the units in per square metre?

jmaerz:
Otherwise, you don't get the tendencies right, when interpolating to levitus - but also in time (due to changing layer thickness)... - see e.g. also water column DMS production terms. Initially, the purpose of the output was slightly different and your comment got me back to this - and since I am also interested in the budgeting of the terms, this is the correct way to do it (as far as I understand). 

The global ocean volume integral for tendencies $T_{l,t,x}$ (the total change/flow) aren't the same for time $t$-varying layer volumes $V_{l,t,x}$:

$$T_\mathrm{total}\mathrm{[mol/s]}=\oint \frac{1}{N}\sum_{t=1}^N T_{l,t,x}V_{l,t,x} ~~~\neq~~~ \oint \bar{V_{l,x}}\frac{1}{N}\sum_{t=1}^N T_{l,t,x}~~~~\mathrm{with}: \bar{V_{l,x}}=\frac{1}{N}\sum_{t=1}^{N}V_{l,t,x}$$

if one only uses the mean volume over time (and similarly, when remapping to levitus depths).  Since areas are not changing, we thus multiply with $dz$ (and post-processing with area).

JorgSchwinger:
I'm not sure (maybe I am misunderstanding something) - when accumulating layer fields, the changing layer thickness is taken into account by weighting with the layer thickness each time step in `acclyr` (and dividing by the average layer thickness in `finlyr` before writing the output).

If a field is interpolated to z-levels each time step this is not necessary. And I do not see a reason why it would be wrong to interpolate from changing isopycnals to z-layers and average the result. The interpolation procedure is conservative and this should apply to tendencies, too.

JorgSchwinger:
Yes, at least for `acclyr` the first part of you equation is implemented. 

So the question is whether doing an the interpolation to constant z-levels and then summing up would change the budget. I somehow don't belive it at least not as long as the interpolation is a linear operation (which is is, I think).

jmaerz:
Hi @JorgSchwinger , I updated my comment by including an equation that briefly shows the issue. The problem is the correlation or lack thereof between changes in rates and changes in volumes of the layer boxes. It is similar to fluxes from one grid box to another and their budgeting, which is why we on the long term would like to have terms like $u\cdot C$, $v\cdot C$ and $w\cdot C$ (and mixing equivalents) as output in BLOM to enable correct budgeting (where we cannot simply use the mean of velocity times mean of concentrations).

jmaerz:
Ok, starting to get your point... - checking tmw.

PR #523, opened 2025-03-27
make N-deposition for SSPs for 1/4 BLOM

YanchunHe:
* make N-deposition for SSPs for 1/4 BLOM
* add global attributes before remmaping N-dep to high resolution to improve performance
* generated N-deposition files are saved under /cluster/shared/noresm/inputdata/ocn/blom/bndcon

related to issue: https://github.com/NorESMhub/NorESM/issues/652

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3842,7 +3842,7 @@
       <value blom_n_deposition="yes" blom_ndep_scenario="1850" ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_1850_CMIP6_tnx0.25v4_20190912.nc</value>
       <value blom_n_deposition="yes" blom_ndep_scenario="2000" ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_2000_CMIP6_tnx0.25v4_20200826.nc</value>
       <value blom_n_deposition="yes" blom_ndep_scenario="hist" ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_185001-201412_tnx0.25v4_20190705.nc</value>
-      <value blom_n_deposition="yes" blom_ndep_scenario="ssp"  ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_201501-210012-${BLOM_NDEP_SCENARIO}_tnx0.25v4_20250325.nc</value>
+      <value blom_n_deposition="yes" blom_ndep_scenario="ssp"  ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndeposition_201501-210012-${BLOM_NDEP_SCENARIO}_tnx0.25v4_20250325.nc</value>
       <value blom_n_deposition="yes" blom_ndep_scenario="1850" ocn_grid="tnx0.125v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_1850_CMIP6_tnx0.125v4_20221013.nc</value>
     </values>
    <desc>File name (incl. full path) for atmopheric N-deposition data</desc>
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3842,6 +3842,7 @@
       <value blom_n_deposition="yes" blom_ndep_scenario="1850" ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_1850_CMIP6_tnx0.25v4_20190912.nc</value>
       <value blom_n_deposition="yes" blom_ndep_scenario="2000" ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_2000_CMIP6_tnx0.25v4_20200826.nc</value>
       <value blom_n_deposition="yes" blom_ndep_scenario="hist" ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_185001-201412_tnx0.25v4_20190705.nc</value>
+      <value blom_n_deposition="yes" blom_ndep_scenario="ssp"  ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_201501-210012-${BLOM_NDEP_SCENARIO}_tnx0.25v4_20250325.nc</value>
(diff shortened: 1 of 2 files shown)


cime_config/namelist_definition_blom.xml
```diff
@@ -3842,6 +3842,7 @@
       <value blom_n_deposition="yes" blom_ndep_scenario="1850" ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_1850_CMIP6_tnx0.25v4_20190912.nc</value>
       <value blom_n_deposition="yes" blom_ndep_scenario="2000" ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_2000_CMIP6_tnx0.25v4_20200826.nc</value>
       <value blom_n_deposition="yes" blom_ndep_scenario="hist" ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_185001-201412_tnx0.25v4_20190705.nc</value>
+      <value blom_n_deposition="yes" blom_ndep_scenario="ssp"  ocn_grid="tnx0.25v4">$DIN_LOC_ROOT/ocn/blom/bndcon/ndep_201501-210012-${BLOM_NDEP_SCENARIO}_tnx0.25v4_20250325.nc</value>
```
jmaerz: Hi @YanchunHe , thanks for putting your changes in! I here personally would prefer to remove the links on betzy and just write here `.../bnd/ndeposition_201501-210012-${BLOM_NDEP_SCENARIO}_tnx0.25v4_20250325.nc`. But maybe @TomasTorsvik and @JorgSchwinger are differently opinionated (for explanation: with the separate `nhx` and `noy` inclusion into those files, I started to name them `ndeposition` instead of purely `ndep`, which gave me some indication of the differences...) Maybe you can comment?

JorgSchwinger: I would also suggest to have the naming here consistently `ndeposition_...`. But maybe the links on betzy are needed for backwards compatibility?

JorgSchwinger (approved):

mvertens:
@JorgSchwinger - I must be missing a key point but I am wondering why we are not using the functionality that I added in https://github.com/NorESMhub/BLOM/pull/410 to obtain the ndep from datm or cam using rather than recreating this forcing data for each resolution. In particular, if you are running fully active cases you always want CLM and BLOM to receive the same ndep - so that would be what CAM is sending.

cime_config/namelist_definition_blom.xml
YanchunHe: I was actually using an older version of BLOM 1.6.9 (NorESM2_3_develop), so it looking for the name `ndep_xxx-xxx*.nc`. So this is mainly for backward compatibility (if we think it is necessary). Or we can use `ndeposition_` here for the latest BLOM version, but commit changes to the but use `ndep_*.nc` for `NorESM2_3_develop`?

YanchunHe:
> @JorgSchwinger - I must be missing a key point but I am wondering why we are not using the functionality that I added in #410 to obtain the ndep from datm or cam using rather than recreating this forcing data for each resolution. In particular, if you are running fully active cases you always want CLM and BLOM to receive the same ndep - so that would be what CAM is sending.

Sounds a good point. I was using NorESM2_3_develop and found the N-dep is missing and end up making these missing files. Maybe in the latest version with NUOPC, we obtain the the N-dep from the coupler.

If so, I can just merge this to the NorESM2_3_develop branch instead.

JorgSchwinger:
> If so, I can just merge this to the NorESM2_3_develop branch instead.

Since we currently still support running with MCT, I think it is fine to merge this into master

For running NorESM3 this would not have been necessary thanks to the new NUPC developments by @mvertens !

jmaerz:
@mvertens , there are still fixes to do for the N-deposition using CDEPS (to my understanding) - see #411 (and fdust - see #404, and the rivers - so iHAMOCC isn't yet fully suported). Second, using sometimes still MCT, it's useful to have those files around. @JorgSchwinger and @YanchunHe , I don't see a reason, why the links should be needed for backwards compatibility (maybe, if you're using a really old iHAMOCC version that handles these things differently) - but at least the line in `namelist_definitions_BLOM.xml` could read `ndeposition` (leave the links, if you see a good reason).

mvertens:
@jmaerz @JorgSchwinger @YanchunHe - thanks for all of your clarifications. My plan was to implement dust and riverine input using CDEPS inline functionality once MCT was no longer supported in the BLOM code in order to have a clean implementation. I'm still not clear if there is consensus as to when and if this will happen. In terms of #411 - I'll look into it. I think for the currently fully coupled simulations it would be good to clarify what ndep is being used by BLOM and what NDEP is being used by CLM and make sure that everything is consistent.

YanchunHe:
> @mvertens , there are still fixes to do for the N-deposition using CDEPS (to my understanding) - see #411 (and fdust - see #404, and the rivers - so iHAMOCC isn't yet fully suported). Second, using sometimes still MCT, it's useful to have those files around. @JorgSchwinger and @YanchunHe , I don't see a reason, why the links should be needed for backwards compatibility (maybe, if you're using a really old iHAMOCC version that handles these things differently) - but at least the line in `namelist_definitions_BLOM.xml` could read `ndeposition` (leave the links, if you see a good reason).

I am using `blom v1.6.9`.

Yes, I agree we can change in the namelist to use `ndeposition_*`. And remove the links.

I will make a new commit.

YanchunHe:
> > @mvertens , there are still fixes to do for the N-deposition using CDEPS (to my understanding) - see #411 (and fdust - see #404, and the rivers - so iHAMOCC isn't yet fully suported). Second, using sometimes still MCT, it's useful to have those files around. @JorgSchwinger and @YanchunHe , I don't see a reason, why the links should be needed for backwards compatibility (maybe, if you're using a really old iHAMOCC version that handles these things differently) - but at least the line in `namelist_definitions_BLOM.xml` could read `ndeposition` (leave the links, if you see a good reason).
> 
> I am using `blom v1.6.9`.
> 
> Yes, I agree we can change in the namelist to use `ndeposition_*`. And remove the links.
> 
> I will make a new commit.

Now I commit a change to use `ndeposition` as file name prefix. However, I don't update these already existing files and also corresponding entries in the namelist  xml file. (I tried to make new ndep for the pi and hist, but looks like the retrieve data script does not download the ndep for these period. So I just gave up).

jmaerz:
Hi @YanchunHe , I think it is fine to keep the old files in place for now. Once you go for a more recent (e.g. master) version, you would need updated files, though, since there, only nhx and noy are read... The respective original input files are under the/my folder that I sent you via email.

jmaerz (approved):
Fine for me as it is now. Many thanks for taking care and putting this in!

PR #524, opened 2025-03-31
add new nuopc stream capability for dust(fedep) forcing

mvertens:
This PR implements new CDEPS/stream functionality input for dust(fedep) forcing when NUOPC/CMEPS is used and at the same time maintains backwards compatibilty with the older fedep input when MCT.
- Using the new stream capability is triggered via the use of the new namelist
 `USE_STREAM_DUST = .true.`
When this is true - the following new namelist is also read
```
&stream_dust
  STREAM_DUST_DATA_FILENAME = '/cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_mhw2006_T42.nc'
  STREAM_DUST_MESH_FILENAME = '/cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_mhw2006_T42_ESMFmesh_cdf5.nc'
  STREAM_DUST_VARNAME = 'dustdep'
  STREAM_DUST_YEAR_ALIGN = 1850
  STREAM_DUST_YEAR_FIRST = 1
  STREAM_DUST_YEAR_LAST = 1850
/
```
- A new file has been added  `nuopc/driver/ocn_stream_dust.F90` that implements this capability (similar to `ocn_stream_sss.F90` and `ocn_stream_sst.F90`).
- the forcing dataset /cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_mhw2006_T42.nc was created by @JorgSchwinger. 
- another new file has been added in` hammocc/mo_output_forcing.F90` which just outputs an instantaneous profile of the forcing data at the first time step. This is just a sanity check that the new stream capability is working as expected.
The following compares the files generated using mo_output_forcing.F90 between the old (MCT) and new (CDEPS/streams) forcing using the test SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel
Screenshot 2025-03-31 at 9 59 12 AM

- Testing:
Verified that the following two tests worked:
SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel
ERS.T62_tn14.NOINYOC.betzy_intel

- Validation:
@JorgSchwinger - you had mentioned wanting to do a longer simulation to validate this new capability given that no history output was produced for the forcing data.  Is this still the case given the output above? How should validation proceed?


jmaerz:
Hi @mvertens , the image above looks quite similar, when taking into account the different value range. For safety, could you provide the plots next to each other with the same range applied?

mvertens:
@jmaerz - thanks for looking at this. Attached are the plots again - where I requested that the range for fedep_old.nc goes from 0 to 0.0016662. However, this does not get displayed in the displayed range value - which is what was produced originally.
Screenshot 2025-04-04 at 6 33 49 PM

mvertens:
@jmaerz- I would expect differences given that I am using a different forcing dataset, possibly different spatial interpolation (bilinear in my case) and a time interpolation at every time step.

TomasTorsvik (approved):

TomasTorsvik:
@jmaerz , @JorgSchwinger - This PR looks fine to me, but I have not worked much on dust deposition and forcing files. Can you also give this another look before we merge this PR?

TomasTorsvik:
@mvertens - sorry, I hope I didn't mess up your feature branch just now. I resolved a conflict in the `namelist_definitions_blom.xml` file, and I didn't realize this would cause `master` to be merged into your feature branch. Looks like the PR contribution is unchanged, so probably it is not a problem.

JorgSchwinger (approved):
Sorry that I didn't manage to look at this earlier - and thanks @mvertens for implementing this! It looks good to me.

jmaerz:
Hi @TomasTorsvik , thanks for pushing this forward - I would suggest to once check again, when dust deposition has been implemented as output - to check, if the year mean dust deposition flux is equal to the input file.

JorgSchwinger:
I agree, that should be done! I hope I can do this in the coming week(s)

PR #528, opened 2025-04-07
Sediment output

jmaerz:

with this PR, output for the sediment aerob, denitrification and 'sulfate reduction' rates are introduced for the default model in order to get more insights into the sediment processes role in nutrient cycling.


JorgSchwinger (approved):
Looks good to me, thank you!

TomasTorsvik (approved):
Nice, thanks!

PR #529, opened 2025-04-07
Fluxes, surface Alk, PO4, NO3 and ODZ volume to i-monitoring files 

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , I added a number of further -useful- global monitoring fields to the monitoring files output. This includes:
- Surface alkalinity, PO4 and NO3
- Various gaseous surface fluxes
- ODZ volume (ODZ volume for oxygen concentrations < 20mumol/L) 
- diffusive sediment pore water tracer fluxes
- riverine input fluxes
- N-deposition fluxes

**Yearly i-monitoring file writing is now default.**

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -5428,7 +5428,7 @@
     <category>diabgc</category>
     <group>diabgc</group>
     <values>
-      <value>0,1,0</value>
+      <value>0,1,2</value>
     </values>
     <desc>Writing the BGC inventory files at BGC frequencies</desc>
   </entry>
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -4911,7 +4911,7 @@
     </values>
     <desc>Sediment: remineralization rate (at reference temperature) (1/(kmol O2/m3 s))</desc>
   </entry>
-  
+
   <entry id="sed_O2thresh_hypoxic">
     <type>real</type>
     <category>bgcparams</category>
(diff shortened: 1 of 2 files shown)


TomasTorsvik (approved):
@jmaerz - thanks! Looks fine to me :)

JorgSchwinger (approved):
Looks good, I do not rely on the inventory output in the log-files.

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , I added further fluxes (diffusive sediment pore water fluxes, riverine input and N-deposition fluxes - the two latter mainly for the purpose of future developments/coupling of the N-cycle) and made yearly output of i-moni files the default. I tested it for a year run. I further fixed some units (i.e. total sediment pore water volume L->m3, similarly pore water concentrations). What I noticed is that daily i-moni files end up with a bit of a strange day date (e.g. 2nd or 3rd of a month - also being written at the 3rd of January...). I won't look into this as it isn't the classical use case, but maybe worth to note down... In any case, I hope that these enhancements will simplify monitoring a bit.

I will merge it soon - hope that this is ok.

JorgSchwinger:
> What I noticed is that daily i-moni files end up with a bit of a strange day date (e.g. 2nd or 3rd of a month - also being written at the 3rd of January...). I won't look into this as it isn't the classical use case, but maybe worth to note down... In any case, I hope that these enhancements will simplify monitoring a bit.

I think at a startup run, the ocean is only starting to run after one day (that's how it used to be, not sure this is still the case). Maybe this behavior is related to this? But I agree, this is nothing with priority right now.

jmaerz:
@JorgSchwinger , possibly - it seems that for the daily files, the 2nd of each month (or 3rd of January in case of startup) is used (eg. xxxx-01-03 for the January), while for the monthly output, the beginning of the next month is used. As you wrote, nothing of importance right now - just something to note down. Merging this now.

PR #531, opened 2025-04-09
Compatible Piecewise Parabolic Method (CPPM) added as option for layer thickness and tracer transport

matsbn:
The Compatible Piecewise Parabolic Method (CPPM) has been added as an option for layer thickness and tracer transport. The interface to the existing transport method, incremental remapping, has been modified to simplify the incorporation of two transport methods. In this process, a little used method, 'eitflx', of applying eddy-induced transport with incremental remapping was removed.

The choice of transport method is controlled by the namelist variable `ADVMTH` than can be either `'remap'` (the old method, still default in NorESM configuration) or `'cppm'`. The limiting method of CPPM can be controlled by the namelist variable `CPPM_LIMITING` that can be either `'monotonic'` or `'non_oscillatory'`. The latter method is default and, at least in idealized advection tests, provide better accuracy of smooth extremas.

Results when using the old transport method are not bit-identical, but initial errors should be at roundoff level. The reason for this is that 1) the halo width has been increased from 3 to 4 grid cells, which impacts roundoff errors in global sums (`xcsum`), and 2) the interface to incremental remapping has been modified to simplify the incorporation of two transport methods.

Diagnostics of OMIP2 simulations for one forcing cycle, only differing by the transport method, can be found here:

Incremental remapping: https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/matsbn/NOIIAJRAOC_TL319_tn14_2.5a9a_20250225/

CPPM: https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/matsbn/NOIIAJRAOC_TL319_tn14_2.5a9a_cppm_nonosc_20250321/


TomasTorsvik:
@matsbn - You need to add the "NorESM Development" project as well as the milestone if it should be visible in the overview :)

JorgSchwinger:
Hi @matsbn, increasing the halo-size from 3 to 4 does that increase the overall data volume to be communicated among processes by ~30%? That would be a lot and would probably affect the scalability? Do you plan to use this as the default method?

JorgSchwinger (approved):

matsbn:
> Hi @matsbn, increasing the halo-size from 3 to 4 does that increase the overall data volume to be communicated among processes by ~30%? That would be a lot and would probably affect the scalability? Do you plan to use this as the default method?

The full halo width of 4 is only exchanged for CPPM (needs a larger stencil for the higher accuracy). Other exchanges are the same and often only a halo width of 1 is actually exchanged. The additional memory allocation could potentially have a performance penalty, but I did not notice a big difference for the tnx1v4 grid. When comparing OMIP2 simulations (with iHAMOCC) on the tnx1v4 grid using 354 processors for ocean, the old transport scheme (with halo size 3) had an ocean throughput of 20.74 myears/wday, while CPPM with non-oscillatory limiter had a throughput of 19.93 myears/wday. This does not test scalability and there could be Betzy issues impacting these timings.

JorgSchwinger:
Thanks for the clarification, sounds good.

TomasTorsvik (approved):

PR #532, opened 2025-04-09
Fix output volume settings

jmaerz:
Hi @JorgSchwinger , @matsbn and @TomasTorsvik ,

this PR fixes some issue with the recently introduced output size settings. It now sets a default value (i.e. the value, where no switch is set) and the spinup-output volume, when the usermods are set (as before). This doesn't change the behavior of the spinup settings, but if those are not set explicitly, the previous default values are now set again. Hence, it also doesn't affect the previously tagged version (i.e. spinup size for output was always(!) chosen).

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -6,8 +6,8 @@
 
   <entry id="BLOM_OUTPUT_SIZE">
     <type>char</type>
-    <valid_values>spinup</valid_values>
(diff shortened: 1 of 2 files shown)


JorgSchwinger (approved):

TomasTorsvik (approved):

matsbn (approved):
Looks good!

PR #539, opened 2025-04-16
Nutrient limitation output

jmaerz:
@TomasTorsvik and @JorgSchwinger , with this PR, we eventually introduce output on the limiting nutrient (in terms of time fractions - range 0-1). 

I now tested the branch with N-cycle on and off - while wasn't using the most tuned model versions. Both show similar large scale nutrient limitation pattern, which indicates that the calculation of time fraction of respective nutrient limitation should be reasonable, while I thus far do not judge on the correspondence to observations (which can be done while further model tuning, I suggest). Nevertheless, please have a closer look.

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -7875,7 +7875,7 @@
       <value>0,2,2</value>
       <value is_test="yes">4,2,2</value>
       <value is_test="yes" empty_hist="yes">0,0,0</value>
-      <value hamocc_output_size="spinup">0,2</value>
+      <value hamocc_output_size="spinup">0,0</value>
     </values>
     <desc>Nutrient limitation Fe - time fraction</desc>
   </entry>
@@ -7888,7 +7888,7 @@
       <value>0,2,2</value>
       <value is_test="yes">4,2,2</value>
       <value is_test="yes" empty_hist="yes">0,0,0</value>
-      <value hamocc_output_size="spinup">0,2</value>
+      <value hamocc_output_size="spinup">0,0</value>
     </values>
     <desc>Nutrient limitation N - time fraction</desc>
   </entry>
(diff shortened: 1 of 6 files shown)


TomasTorsvik (approved):
@jmaerz - thanks, looks good to me.

JorgSchwinger (commented):
Thanks for implementing this @jmaerz . My only concern is a bit the output volume. Could we in addition (and as the default setting) output a field that gives you (vertically averaged over the top 100m or so) the nutrient that was the most limiting nutrient over the averaging period? I'm sure that would be enough information for most case, but would need a factor 210 less storage...

jmaerz:
I am about implementing an euphotic zone nut lim output.

jmaerz:
@JorgSchwinger I now added output for time fraction mean euphotic zone nutrient limitation. I hope this helps for tuning, etc. However, I am a bit confused when comparing the results with Fig. 10 of Tjiputra et al. 2020 - while iron limitation is very similar, N- and P-limitation is different (i.e. only few regions with N-limitation opposing current knowledge). The current nutrient limitation output is, however, supported by e.g. surface N:P ratios (compared to iHAMOCC stoichiometry). I would suggest to push this now and re-check afterward, when once run in a more tuned version. - potentially also re-checking the script used for producing Fig. 10....

JorgSchwinger (approved):
Regarding Fig 10 in Tjiputra et al 2020: it shows mostly iron and N-limitation, only a few patches with P-limitation, is this opposed to current knowledge?

jmaerz:
@JorgSchwinger , the figure locks good to me, but the output of my current test simulations show little N-limitation. This might be a tuning issue (as in: the -surface- year mean N:P ratio supports the nutrient limitation output), but might deserve an eye. My question with respect to Fig. 10 was more, what has changed so that the N-limitation has switched to large areas of P-limitation in iHAMOCC in the potentially not well tuned current test simulations. But if not much has changed with respect to Tjiputra, then the question is, if the figure needs a re-visit. Sorry for the confusion.

JorgSchwinger:
@jmaerz I guess the main change was the extended nitrogen cycle :)

So its good we have the output now, makes it easier to have an eye on it (Jerry's figure was also made on annual mean surface nutrient concentrations, so this might also be an issue)

jmaerz:
@JorgSchwinger , just to clarify further: I was running both configurations - with and without the extended N-cycle, but both probably un-tuned (and both with similar surface N-limitation pattern). Let's see, how it behaves in more tuned versions.

Issue #538, opened 2025-04-16
Single column run fails with floating point exception


jmaerz:
@matsbn Single column run fails after the latest update of BLOM introducing the CPPM method (version before was working with my limits file - tested). I am not sure, if it will be reproducible with any limits file, but I was just replacing:

```  
RMPMTH   = 'eitvel'
```
by
```
ADVMTH   = 'cppm'
CPPM_LIMITING = 'monotonic'
```
(also tested `ADVMTH='remap') in my current limits file and then the run fails in both cases with a floating point exception:
``` 
       0.7880 Time 1 Just before main loop
----------------------------------- stderr -----------------------------------

Program received signal SIGFPE: Floating-point exception - erroneous arithmetic operation.

Backtrace for this error:
#0  0x736f09023e59 in ???
#1  0x736f09022e75 in ???
#2  0x736f08c4532f in ???
	at ./signal/../sysdeps/unix/sysv/linux/x86_64/libc_sigaction.c:0
#3  0x583310a07e21 in __mod_advect_MOD_advect
	at ../phy/mod_advect.F90:79
#4  0x583310a29a34 in __mod_blom_step_MOD_blom_step
	at ../phy/mod_blom_step.F90:167
#5  0x583310ba81d4 in blom
	at ../drivers/nocoupler/blom.F:42
#6  0x583310a07792 in main
	at ../drivers/nocoupler/blom.F:26
```

jmaerz:
Closed via #541.

PR #541, opened 2025-04-21
Bugfix of mod_advect.F90

matsbn:
This fixes a bug in do-loop ranges in mod_advect.F90 that made `expcnf == 'single_column'` fail. Hopefully this resolves
https://github.com/NorESMhub/BLOM/issues/538. No change of result for other experiment configurations.


milicak (approved):

jmaerz (approved):
Hi @matsbn , thanks for fixing!

PR #542, opened 2025-04-22
Bugfix of mod_inigeo.F90

matsbn:
Fixed bug that could trigger failure during floating-point to integer conversion. Should not change results.


TomasTorsvik (approved):
@matsbn - thanks for providing the bug fix!

TomasTorsvik:
@matsbn - I would like to merge this, make a tag on BLOM, and see if we can get this into a quick `noresm3_0_alpha03a` tag.

PR #544, opened 2025-04-25
Write now to netcdf3_64bit

### LINKED ISSUE #543: Include option to output remapping in either netcdf4 or netCDF3_64bit format
I sometimes have encountered errors when trying to read in netCDF4 files as input to HAMOCC. I last had this problem when I tested the remapping tool to read in output from NOINYOC_T62_tn21 into NOIIAJRAOC20TR_TL319_tn14. In this case i solved the problem by converting the netCFD4 file to netCDF3_64BIT.

I think the problem comes from having a mix of netCDF3_64BIT and netCDF4 as input files, since the original restart files were netCDF3_64BIT and only the remapped HAMOCC file was netCFD4. It may be complicated to include support for a mix of these formats in the input files.

Hence, I suggest to include an option to specify the output file format when doing the remapping. This should be fairly straight forward to implement.

jmaerz:
This PR concerns the iHAMOCC restart remapping. The remapping routine now writes to netcdf3_64bit as currently supported by iHAMOCC.

Closes #543

### DIFF
--- a/utils/iHAMOCC/restart_remapping/iHAMOCC_restart_remapping.py
+++ b/utils/iHAMOCC/restart_remapping/iHAMOCC_restart_remapping.py
@@ -480,7 +480,7 @@ def main():
                                    + 'Original file history: '+ds_coarse_hamocc.attrs['history']
  
     # ----------------------------------------- Eventually write the restart file
-    new_restart.to_netcdf(out_new_restart_file)
+    new_restart.to_netcdf(out_new_restart_file,format="NETCDF3_64BIT")
   
     # Clean up temporary files
     clean_tmpfiles(tmpfiles)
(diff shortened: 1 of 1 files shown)


TomasTorsvik:
@jmaerz - Thanks, I will test this for the NOINYOC -> NOIIAJRAOC dataset that I remapped earlier.

jmaerz:
Hi @TomasTorsvik , that would be great - then I leave the testing to you.

TomasTorsvik:
@jmaerz - Now it works with the NOINYOC -> NOIIAJRAOC dataset. I was able to run `NOIIAJRAOC20TR_TL319_tn14` with the remapped `NOINYOC` BGC fields.

TomasTorsvik (approved):

PR #545, opened 2025-05-02
Add compsets and tests for JRA repeat year forcing

TomasTorsvik:
This PR makes it easier to use the JRA repeat year compsets.

I also removed test description specifying "gnu" or "intel", since some tests are done with both compilers and the compiler is specified in the test name anyway.

Tests:
- `aux_blom_noresm`: tests PASS (except the "expected FAIL" tests)


mvdebolskiy:
Was the updated test-suite run?

TomasTorsvik:
> Was the updated test-suite run?

Yes, output for the test is available at: /cluster/work/users/tomast/noresm/nor3.0a03a-blom
I did not make new baselines yet, thinking to do that after the `noresm3_0_alpha03a` tag has been created.

JorgSchwinger (approved):
I cannot say much for the test and compilers, but otherwise it looks good.

PR #546, opened 2025-05-02
Modified shortwave absorption module

matsbn:
A new shortwave radiation absorption method, 'spatial_frac_attlen', has been added where spatially varying spectral band fractions and attenuation lengths from are read from file.

In addition, spectral band fraction and attenuation length for long visible wavelengths has been added, which makes the implementation more consistent with the references cited in mod_swabs.F90. This is currently only used when a vertical coordinate using the ALE method is selected, but then all shortwave radiation absorption methods, except 'top-layer', will have a modified vertical absorption profiles.

### DIFF
--- a/ben02/mod_thermf_ben02.F90
+++ b/ben02/mod_thermf_ben02.F90
@@ -37,7 +37,7 @@ module mod_thermf_ben02
                            fmltfz, sfl, ustarw, surflx, surrlx, &
                            sswflx, salflx, brnflx, salrlx, ustar, &
                            salt_corr, trc_corr, t_rs_nonloc, s_rs_nonloc
-  use mod_swabs,     only: swbgal, swbgfc
+  use mod_swabs,     only: swal2, swfc2
   use mod_ben02,     only: tsi_tda, tml_tda, sml_tda, alb_tda, fice_tda, &
                            tsi, ntda, dfl, albw, alb, &
                            rnfins, rnfres, nrfets, rhowat
@@ -300,7 +300,7 @@ contains
 
           ! Predict temperature change in mixed layer after a leapfrog time step
           ! due to heat fluxes
-          swfac = 1.-swbgfc(i,j)*exp(-hotl/swbgal(i,j))
+          swfac = 1.-swfc2(i,j)*exp(-hotl/swal2(i,j))
           dtml = (swfac*qsww+qnsw)*2.*dt/(cpsw*rhowat*hotl)
 
           if     (totl+dtml < tice_f) then
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -653,6 +653,18 @@
     </desc>
   </entry>
 
+  <entry id="svfile" is_inputdata="yes">
+    <type>char</type>
+    <category>limits</category>
+    <group>limits</group>
+    <values>
+      <value>unset</value>
+    </values>
+    <desc>
+      Name of file containing spatially varying spectral band fractions and attenuation lengths
+    </desc>
+  </entry>
+
   <entry id="trxday">
     <type>real</type>
     <category>limits</category>
--- a/cime_config/ocn_in.readme
+++ b/cime_config/ocn_in.readme
@@ -84,11 +84,13 @@
 ! NIWLF    : Fraction of near-inertial energy dissipated locally beneath
 !            the boundary layer () (f)
 ! SWAMTH   : Shortwave radiation absorption method. Valid methods:
-!            'top-layer', 'jerlov', 'chlorophyll' (a)
+!            'top-layer', 'jerlov', 'chlorophyll', 'spatial_frac_attlen' (a)
 ! JWTYPE   : Number indicating the Jerlov (1968) water type (i)
 ! CHLOPT   : Chlorophyll concentration option. Valid options:
 !            'climatology' (a)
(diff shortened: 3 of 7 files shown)


TomasTorsvik:
@matsbn - thanks for the PR!
I suppose this will be an optional feature, or do you want to enable this as default?
What testing has been done with the new shortwave absorption scheme?

TomasTorsvik (approved):

matsbn:
> @matsbn - thanks for the PR! I suppose this will be an optional feature, or do you want to enable this as default? What testing has been done with the new shortwave absorption scheme?

The method 'spatial_frac_attlen' is optional, while the additional spectral band for the ALE method is non-optional. Currently running a full OMIP2 cycle with the PR modifications and the method 'chlorophyll' that I expect to set as default for BLOM soon.

jmaerz:
@TomasTorsvik and @JorgSchwinger , is this something that we should/are supposed to incorporate as well in iHAMOCC (i.e. replacing the fixed short-wave fraction in phytoplankton growth with the new scheme-derived values?)

TomasTorsvik:
> @TomasTorsvik and @JorgSchwinger , is this something that we should/are supposed to incorporate as well in iHAMOCC (i.e. replacing the fixed short-wave fraction in phytoplankton growth with the new scheme-derived values?)

@jmaerz - seems like a good idea, but I won't have time to look at this. I am also not sure if this should be a high priority at the moment.

JorgSchwinger:
> @TomasTorsvik and @JorgSchwinger , is this something that we should/are supposed to incorporate as well in iHAMOCC (i.e. replacing the fixed short-wave fraction in phytoplankton growth with the new scheme-derived values?)

But then, we also would like to have this self-consistent with HAMOCC-calculated phytoplankton? To me this sounds we would need to put some thinking into it. Agree that we shouldn't prioritize this now.

JorgSchwinger (approved):
Can't judge the details, but looks good

milicak (approved):

jmaerz (approved):
Dear @matsbn , thanks for bringing in the new feature. As far as I understand, the new introduced spatially varying fractions of radiation could potentially affect the photosynthetic active radiation available in iHAMOCC (which we currently set constant for the incoming `swa` - @JorgSchwinger and @TomasTorsvik : through the factor of `0.4` lumped in `pi_alpha`), right? 
@JorgSchwinger ,  technically yes, there could be a phytoplankton feedback on attenuation and thus heat fluxes, while, as far as I understand the code in BLOM, thus far, climatological Chl maps are -optionally- used currently.

matsbn:
A full JRA cycle has been completed with this PR using `swamth = 'chlorophyll'`. It also contains other BLOM namelist changes that I propose as new defaults for the upcoming NorESM3 development releases. The total list is of changes is:
```
PGFMTH = 'dynamic enthalpy'
ADVMTH = 'cppm'
SWAMTH = 'chlorophyll'
BDMC2 = 1.5e-5
BDMLDP = .true.
```
Standard diagnostics of this simulation (NOIIAJRAOC_TL319_tn14_3.0a3_newmix_20250430) can be found here: https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/matsbn/NOIIAJRAOC_TL319_tn14_3.0a3_newmix_20250430/

It is also compared to a simulation (NOIIAJRAOC_TL319_tn14_2.5a9a_20250225) that are using current default BLOM settings. It should be mentioned that the new simulation uses 1850 CO2 level, while the former uses 2000 CO2 level, making interpreting some iHAMOCC differences difficult.

In the new simulation there is a welcome cooling of SST in midlatitudes during summer, which is likely due to the change in shortwave absorption profile. The warmer SST in the tropics would also be interesting to see the impact of in fully coupled simulations (likely a combination of shortwave absorption and vertical diffusivity changes).

To conclude, there are no issues in this simulation that can be attributed to this PR.

matsbn:
> Dear @matsbn , thanks for bringing in the new feature. As far as I understand, the new introduced spatially varying fractions of radiation could potentially affect the photosynthetic active radiation available in iHAMOCC (which we currently set constant for the incoming `swa` - @JorgSchwinger and @TomasTorsvik : through the factor of `0.4` lumped in `pi_alpha`), right? @JorgSchwinger , technically yes, there could be a phytoplankton feedback on attenuation and thus heat fluxes, while, as far as I understand the code in BLOM, thus far, climatological Chl maps are -optionally- used currently.

@jmaerz, the BLOM changes introduced here does not modify the incoming shortwave radiation, `swa`, only how it is absorbed in BLOM beneath the ocean surface. To my understanding, the handling of shortwave flux beneath the ocean surface is independent in BLOM and iHAMOCC. But it does not have to be and coupling shortwave absorption with iHAMOCC fields would be interesting (we have briefly discussed this before, but left it for later). The climatological chlorophyll concentration maps are from satellite imagery, and the parameterization we use (Morel and Antoine, 1994) are developed for this kind of climatology. I guess it might not be straightforward to "translate" an iHAMOCC quantity to such a product and maybe there are more clever ways to make use of the 3D structure of iHAMOCC fields for shortwave absorption.

jmaerz:
Dear @matsbn, thanks for clarification. Yes, I understand that currently light fields are independently handled in BLOM and iHAMOCC. I was wondering with respect to the "Read spatially varying spectral band" - whether this refers to the fraction of incoming radiation at surface (and thus could be made use of in iHAMOCC as well in terms of fraction of photosynthetically active radiation - fPAR)?

PR #548, opened 2025-05-02
Bugfix: typo in module name

TomasTorsvik:


jmaerz (approved):
Thanks for fixing the meson.build system and re-naming the module name.

PR #549, opened 2025-05-04
Set character length to 512 for all filenames

TomasTorsvik:
Some test configurations require more than 256 characters for PATH + filename specifications. I suggest to set all filenames to 512 characters by default.

I have included a new `mo_kind.F90` file for iHAMOCC. I see we sometimes use the BLOM file `mod_types.F90`, but I think it can be useful to define this internally in iHAMOCC as well.

**Tests**: `prealpha_noresm` and `aux_blom_noresm` tests pass, except expected fails

### DIFF
--- a/hamocc/mo_kind.F90
+++ b/hamocc/mo_kind.F90
@@ -18,8 +18,8 @@
 module mo_kind
 
   use mod_types,   only: blom_i1 => i1, blom_i2 => i2, blom_i4 => i4, blom_i8 => i8, &
-                         blom_r4 => r4, blom_r8 => r8, blom_kind_cs, blom_kind_cm,   &
-                         blom_kind_cl, blom_kind_cx, blom_kind_cxx
+                         blom_r4 => r4, blom_r8 => r8, blom_char_s, blom_char_m,     &
+                         blom_char_l, blom_char_x, blom_char_xx
   use mod_utility, only: fnmlen
 
   implicit none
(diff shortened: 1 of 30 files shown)


jmaerz:
Hi @TomasTorsvik , many thanks for this important initiative (long standing issue). Admittedly, I am not entirely sure, how to best handle this, but I would suggest to consider to import the BLOM module in `mo_kind` instead and make use of the BLOM definitions for the different kinds along the lines of
```
r8 = BLOM_R8 (pseudo code)
HAMOCC_KIND_CS=BLOM_KIND_CS
```
. Further, I would limit ourselves to what we actually use. I believe, we currently implicitly use `r8` in iHAMOCC. Then, before stepping forward, define a `wp` ('working precision') instead (or based on ) of `r4`,`r8`,..., which makes it easier to apply this `wp` to real number (potentially similar to integers) through e.g. `parameter_name=1.2_wp` and making use of the `wp` (or `rp` for real precision and `ip` for integer precision) throughout all files and hard-coded numbers (which we can do at some point after your PR - so no need to do it right away). Hence, switching precision will be made much easier and supports e.g. its use in other submodules.

TomasTorsvik:
> Hi @TomasTorsvik , many thanks for this important initiative (long standing issue). Admittedly, I am not entirely sure, how to best handle this, but I would suggest to consider to import the BLOM module in `mo_kind` instead and make use of the BLOM definitions for the different kinds along the lines of
> 
> ```
> r8 = BLOM_R8 (pseudo code)
> HAMOCC_KIND_CS=BLOM_KIND_CS
> ```
> 
> . Further, I would limit ourselves to what we actually use. I believe, we currently implicitly use `r8` in iHAMOCC. Then, before stepping forward, define a `wp` ('working precision') instead (or based on ) of `r4`,`r8`,..., which makes it easier to apply this `wp` to real number (potentially similar to integers) through e.g. `parameter_name=1.2_wp` and making use of the `wp` (or `rp` for real precision and `ip` for integer precision) throughout all files and hard-coded numbers (which we can do at some point after your PR - so no need to do it right away). Hence, switching precision will be made much easier and supports e.g. its use in other submodules.

I have changed `mo_kind.F90` so that type parameters are imported from BLOM, and define `rp` and `ip` as suggeted default precision for real and integer variables, repectively. For now I'm keeping all the definitions from BLOM, we can perhaps reduce this after we go through and set precision levels throughout the iHAMOCC code base.

phy/mod_types.F90
```diff
@@ -36,7 +36,17 @@ module mod_types
       i8 = int64, &
       r4 = real32, &
       r8 = real64
+   integer,parameter :: BLOM_KIND_CS = 80      ! short char
```
JorgSchwinger: I find it confusing to have the `_KIND_`in these names, since that is not what they indicate. They just indicate the length of the character string, not it's KIND type parameter (which could, for example, indicate different character sets (if the compiler supported them). Maybe just `BLOM_CS` etc.?

hamocc/mo_kind.F90
```diff
@@ -0,0 +1,46 @@
...
+! FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License for
+! more details.
+!
+! You should have received a copy of the GNU Lesser General Public License
+! along with BLOM. If not, see https://www.gnu.org/licenses/.
+
+module mo_kind
+
+  use mod_types,   only: blom_i1 => i1, blom_i2 => i2, blom_i4 => i4, blom_i8 => i8, &
+                         blom_r4 => r4, blom_r8 => r8, blom_kind_cs, blom_kind_cm,   &
```
JorgSchwinger: The renaming `=>` of BLOM names here does not make sense, if this is done explicitly below. I would stick with the explicit way below and remove this here.

JorgSchwinger (approved):
Thanks for fixing this! Looks good, but see my two comments below.

phy/mod_types.F90
TomasTorsvik: I have changed the parameter names to `[BLOM,HAMOCC]_CHAR_[S,M,L,X,XX]`, this is perhaps more descriptive of what the parameters represent.

hamocc/mo_kind.F90
TomasTorsvik: I don't think we can do 
```
use mod_types, only: i1, i2, ...

integer, parameter :: i1 = i1
```
Alternatively, we could import `int8`, `int16` ... from `iso_fortran_env`, as is done in `mod_types`.

jmaerz: Hi @TomasTorsvik and @JorgSchwinger , for consistency with BLOM, I would stick to what you have currently as long as there is no good reason to go directly with `iso_fortran_env`.

jmaerz:
Hi @TomasTorsvik , I understand that you want to provide/support the full suite of available BLOM options in `mo_kind`. I am a bit hesitant, though, to then set all available ones to public. Can we exclude the ones not used (setting them private) so that one is forced to use `_ip` and `_rp` (and not mistakenly make use of e.g. `_r8`)? - I would be concerned about the risk of mixing those at some point mistakenly.

hamocc/mo_kind.F90
JorgSchwinger: Ah ok, I misunderstood this. Sorry, this is fine as it was

jmaerz (approved):
Hi @TomasTorsvik , thanks for introducing the new module and fixing the name length issue. It's good to make progress towards consistent, explicit working precision.

TomasTorsvik:
@matsbn , @milicak - Any more comments on this PR? I will leave it until tomorrow before merging.

matsbn (approved):
Thanks for fixing this, @TomasTorsvik! We should maybe collect infrastructure related constants in one or more dedicated modules at some point (they are a bit all over the place now), but that can be done later. Also the data type discussion is important to follow up, but should continue in a separate issue.

PR #554, opened 2025-05-07
Correct and add a CaCO3 dissolution rate constant

JorgSchwinger:
This PR fixes the treatment of CaCO3 dissolution in `mo_carchm`. Before, there was a hard-coded dissolution rate of 0.05, but a multiplication with `dtb` was missing. This means effectively that 5% of CaCO3 was dissolved per timestep under undersaturated conditions. The distribution of alkalinity (large positive biases in the deep ocean) points towards this value being much too large. 

To fix these shortcomings and to facilitate tuning, this PR introduces a CaCO3 dissolution constant `dremcalc` that can be set via namelist and is correctly treated w.r.t. time-step length.

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -4764,6 +4764,16 @@
     <desc>remineralization and dissolution parameter: 1/d Dissolution rate for opal</desc>
   </entry>
 
+  <entry id="dremcalc">
+    <type>real</type>
+    <category>bgcparams</category>
+    <group>bgcparams</group>
+    <values>
+      <value>None</value>
+    </values>
+    <desc>remineralization and dissolution parameter: 1/d Dissolution rate for CaCO3 (applied if Omega_c lt 1)</desc>
+  </entry>
+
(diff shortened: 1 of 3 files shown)


TomasTorsvik (approved):
@JorgSchwinger - Thanks for fixing!

jmaerz (approved):
Hi @JorgSchwinger , thanks for checking and fixing! - did you compare the new, applied rate with recent literature values?

JorgSchwinger:
> Hi @JorgSchwinger , thanks for checking and fixing! - did you compare the new, applied rate with recent literature values?

Hi @jmaerz, no I didn't - I think we are forced to apply a much smaller rate to get rid of the large alkalinity biases. It would be super interesting though to check if the sediment CaCO3 look better or worse after this fix... (and/or compare with sediment trap data). But even if the sediment looks worse, I would keep a lower dissolution value because of the water column biases...

I think there is good reason to believe that the old rate was to large. With 5% dissolution per timestep, you basically dissolve all CaCO3 more or less instantaneously once you get into under-saturated water masses.

Issue #557, opened 2025-05-07
ERS tests fail after merging `dev1.9.0.9` tag


TomasTorsvik:
See regression test output in discussion #556 

After merging [dev1.9.0.9](https://github.com/NorESMhub/BLOM/tree/dev1.9.0.9) I see that the ERS tests fail either at run time or when comparing base vs. restart run.

`dev1.9.0.9` introduces new default namelist definitions. I suspect this may trigger some bug related to restart procedure.

TomasTorsvik:
@matsbn - just to confirm, I ran the test with `remap`method enabled
```
 <entry id="advmth">
    <type>char</type>
    <category>limits</category>
    <group>limits</group>
    <values>
      <!-- <value>cppm</value> -->
      <value>remap</value>
      <value blom_vcoord="isopyc_bulkml">remap</value>
    </values>
    <desc>Advection method. Valid methods: 'remap', 'cppm'.</desc>
  </entry>
```
Now the `ERS_Ld3.T62_tn14.NOINYOC.betzy_intel` test pass without problems. I think this is what you already indicated in the meeting earlier today.

TomasTorsvik:
Fixed with #561

PR #558, opened 2025-05-08
Whole array initialization

matsbn:
Uninitialised variables could be accessed by the CPPM transport method. This should not impact the solution, but could trigger floating point exceptions.

### DIFF
--- a/phy/mod_barotp.F90
+++ b/phy/mod_barotp.F90
@@ -78,21 +78,13 @@ contains
     ! Local variables
     integer :: i,j,k,l
 
-    !$omp parallel do private(i)
-    do j = 1-nbdy,jj+nbdy
-      do i = 1-nbdy,ii+nbdy
-        do k = 1,2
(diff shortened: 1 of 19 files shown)


matsbn:
@TomasTorsvik, I hope this will fix the floating point exception mention in issue https://github.com/NorESMhub/BLOM/issues/557 and discussion https://github.com/NorESMhub/BLOM/discussions/556 (but not the restart reproducibility issue).

TomasTorsvik:
@matsbn - Thanks for the PR. As expected, I now see fewer fails at run time (WW3 + gnu compiler fail consistently), but no improvement on the `COMPARE_base_rest` tests.

TomasTorsvik (approved):

milicak (approved):
I am approving this change, but we should avoid using single letter variables such as "z".

matsbn:
I agree some better variable names can be chosen. Suggest to revisit this after physics changes deadline mid-June.

PR #560, opened 2025-05-08
Offline sediment spinup

jmaerz:
This draft PR provides a first draft of a code for an offline sediment spinup. **The draft PR is thus far for illustrative purposes only.** The question is, if there is the possibility to fully include this into the NorESM-framework or if something else is needed/wanted (e.g. a highly modified NorESM branch, etc.; comments?). The aim is to enable a very fast offline sediment spinup with -hopefully- a few thousand years throughput per day even for high resolution setups. This work is part of a Bjerknes MDA and is a first starting point for collaborative coding-work with @TomasTorsvik (and others). 

**What it does:**
Thus far, the code allows to switch off large parts of BLOM (while I didn't try for CICE and coupler infrastructure thus far, where I/we would appreciate help) so that only the timing of BLOM is used to run iHAMOCC only. In iHAMOCC, a forcing file with climatological deposition fluxes is read, when offline sediment spinup is switched on (currently via a pre-processor flag). Then, all water column processes are switched off in iHAMOCC and only sediment processes are simulated for these climatological deposition fluxes that need to be generated beforehand. Technically, this should lead to a fast sediment spinup capability which will allow to spin-up the sediment even in high(er) resolution setups while minimizing water column drifts when plugging in the spunup sediment back into a full model simulation.

**Missing:**
Currently, the code runs on betzy similarly fast as before with all processes included - I assume that CICE and coupler infrastructure and processor settings limit the throughput (i.e. BLOM/iHAMOCC idling while waiting for CICE and coupler). We still need to find a way to switch off CICE and coupler infrastructure. Further, a better processor layout needs to be found for this purpose (i.e. all/most processors for iHAMOCC). 

In a 1D simulation, the code is much faster than the regular 1D simulation and looks promising.


mvertens:
@jmaerz - it is not difficult to switch off the cice and coupler infrastructure. In fact CAM does this already in its simple model configurations. Here is an example for adiabatic mode. In config_compset.xml you have
```
  <compset>
    <alias>FADIAB</alias>
    <lname>2000_CAM%ADIAB_SLND_SICE_SOCN_SROF_SGLC_SWAV</lname>
  </compset>

```
In the mediator buildnml you skip the mediator if there is a prognostic component and all other components are stub:
```
    # Determine if will skip the mediator and then set the
    # driver rpointer file if there is only one non-stub component then skip mediator
    if len(valid_comps) == 2 and not datamodel_in_compset:
        # skip the mediator if there is a prognostic component and all other components are stub
        valid_comps.remove("CPL")
        nmlgen.set_value("mediator_present", value=".false.")
	nmlgen.set_value("component_list", value=" ".join(valid_comps))
    else:
        # do not skip mediator if there is a data component but all other components are stub
        valid_comps_string = " ".join(valid_comps)
        nmlgen.set_value(
            "component_list", value=valid_comps_string.replace("CPL", "MED")
        )
```
The above will set the nuopc.runconfig config variable `mediator_present = false`.
This can then be accessed in the CAM nuopc cap:
```
    call NUOPC_CompAttributeGet(gcomp, name="mediator_present", value=cvalue, isPresent=isPresent, isSet=isSet, rc=rc)
    if (ChkErr(rc,__LINE__,u_FILE_u)) return
    if (isPresent .and. isSet) then
       read (cvalue,*) mediator_present
       if (mediator_present) then
          call advertise_fields(gcomp, flds_scalar_name, rc)
          if (ChkErr(rc,__LINE__,u_FILE_u)) return
       end if
    else
       call shr_sys_abort(subname//'Need to set attribute mediator_present')
    endif
```
And then there is logic in atm_comp_nuopc to not communicate with the mediator if it is not present. This same logic can easily be introduced in the BLOM ocn_comp_nuopc.F90. So all you need to do is introduce a new compset in BLOM and then add the logic to ocn_comp_nuopc. I'm happy to help with this - let me know.

jmaerz:
@mvertens , many thanks! - even though the offline sediment spinup isn't high up on the priority list, your recipe looks like a good step further. Once a bit more time, I (or Tomas) will test this out - thanks for looking into it and providing guidance!

jmaerz:
Hi @mvertens , first of all: thanks again for the recipe and no rush with this issue! I tried to follow your recipe (not sure thus far, if it works out), but when building the model, I ran into the issue that the model isn't recognizing a PE-layout for the new compset. While building, it throws:
```
Machine is betzy
Pes setting: grid match    is None 
Pes setting: machine match is None 
Pes setting: compset_match is None 
Pes setting: pesize match  is None 
Pes setting: grid          is a%null_l%null_oi%tnx2v1_r%null_w%null_z%null_g%null_m%tnx2v1 
Pes setting: compset       is SCAM_SATM_SLND_SICE_BLOM%ECO_SROF_SGLC_SWAV_SESP 
Pes setting: tasks       is {} 
Pes setting: threads     is {} 
Pes setting: rootpe      is {} 
Pes setting: pstrid      is {} 
Pes other settings: {}
Pes other settings append: {}
setting additional fields from config_pes: {}, append {}
 Compset is: SCAM_SATM_SLND_SICE_BLOM%ECO_SROF_SGLC_SWAV_SESP 
 Grid is: a%null_l%null_oi%tnx2v1_r%null_w%null_z%null_g%null_m%tnx2v1 
 Components in compset are: ['satm', 'slnd', 'sice', 'blom', 'srof', 'sglc', 'swav', 'sesp'] 
```
To me, it looks as if the first step (recognizing the new compset) is successful. Do you have a recommendation for a PE layout? - I suspect that mainly the ocean requires `ntasks` (since only the sediment of iHAMOCC will be running), but setting up the PE-layout isn't anything, I have done before... - so help would be appreciated to define a new one in `config_pes.xml`. No worries, there is no rush with this...

PR #561, opened 2025-05-08
Fix CPPM for tripolar grids

matsbn:
This PR corrects edge value coefficients of CPPM near the Arctic seam of tripolar grids. This will hopefully avoid non-reproducibility after restarts as reported in https://github.com/NorESMhub/BLOM/issues/557.


TomasTorsvik (approved):
@matsbn - Thanks, now the ERS test is passing. If you merge this in I can start to prepare the `alpha03b`tag.

milicak (approved):

PR #564, opened 2025-05-12
Allow longer time for 0.5 degree ERS test

TomasTorsvik:
The 0.5 degree ERS test sometimes times out before completing.


jmaerz (approved):
Hi @TomasTorsvik , makes sense to increase the wallclock time for higher resolution cases. Thanks.

PR #565, opened 2025-05-12
Add "noforcing" experiment configuration

matsbn:
This PR adds experiment configuration "noforcing", that allows configuration, building and running of comprehensive global grids in standalone mode with all surface forcing fluxes set to zero. To a large degree this replaces the functionality of the "noforc" driver that had not been maintained and is removed by this PR. In addition, configuration of the Meson Build system for standalone application now supports tripolar grids.

No change in results are expected for the already existing experiment configurations.


TomasTorsvik:
@matsbn - Instead of putting a new experiment config folder in the root folder, could we perhaps collect all the experiment configurations into a single folder? They start to clutter the root folder, and I think it would be better to only have a few sub-folders there. I would suggest a folder "experiment_config" or "exp_config", and move "channel", "fuk95", "noforcing", "single_column" and "tests" into the new folder structure.

matsbn:
> @matsbn - Instead of putting a new experiment config folder in the root folder, could we perhaps collect all the experiment configurations into a single folder? They start to clutter the root folder, and I think it would be better to only have a few sub-folders there. I would suggest a folder "experiment_config" or "exp_config", and move "channel", "fuk95", "noforcing", "single_column" and "tests" into the new folder structure.

@TomasTorsvik , I absolutely agree, and I think the whole folder structure should be revisited. Would it be OK to leave such infrastructure changes to after 15 June? The ability to test tripolar grid configurations in standalone mode will possibly be very useful the next few weeks as the bugfix in https://github.com/NorESMhub/BLOM/pull/561/files would have been very difficult to track down without this ability.

TomasTorsvik:
> > @matsbn - Instead of putting a new experiment config folder in the root folder, could we perhaps collect all the experiment configurations into a single folder? They start to clutter the root folder, and I think it would be better to only have a few sub-folders there. I would suggest a folder "experiment_config" or "exp_config", and move "channel", "fuk95", "noforcing", "single_column" and "tests" into the new folder structure.
> 
> @TomasTorsvik , I absolutely agree, and I think the whole folder structure should be revisited. Would it be OK to leave such infrastructure changes to after 15 June? The ability to test tripolar grid configurations in standalone mode will possibly be very useful the next few weeks as the bugfix in https://github.com/NorESMhub/BLOM/pull/561/files would have been very difficult to track down without this ability.

Yes, I suppose we can do it after 15 June. I will make an issue on it as a reminder.

TomasTorsvik (approved):

milicak (approved):

PR #567, opened 2025-05-13
Add options for CMIP6 settings through usermods

TomasTorsvik:
This PR makes it possible to enable CMIP6 settings for BLOM/iHAMOCC through usermods.

I start including this on the `release-1.6` branch, since this is what is being tested currently for NorESM2.3. If this is accepted, I will enable these usermods, with necessary modifications, for the `v1.7` and `master` branches as well.

### DIFF
--- a/cime_config/usermods_dirs/cmip6_config_tn14/shell_commands
+++ b/cime_config/usermods_dirs/cmip6_config_tn14/shell_commands
@@ -1,0 +1,2 @@
+# configure BLOM/iHAMOCC with CMIP6 settings
+./xmlchange BLOM_UNIT=cgs
--- a/cime_config/usermods_dirs/cmip6_config_tn14/user_nl_blom
+++ b/cime_config/usermods_dirs/cmip6_config_tn14/user_nl_blom
@@ -1,0 +1,12 @@
+!----------------------------------------------------------------------------------
+! configure BLOM/iHAMOCC with CMIP6 settings  - tn14 grid
+!----------------------------------------------------------------------------------
+
+EDANIS = .false.
+EDDF2D = .false.
+EDSPRS = .true.
+RHSCTP = .false.
+EGC = 0.85
+EGMNDF = 100.e4
+EGMXDF = 1500.e4
+RHISCF = 0.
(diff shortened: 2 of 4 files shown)


matsbn (approved):
As far as I can see, this looks fine!

jmaerz:
@matsbn and @TomasTorsvik , I wonder, if there is any good reason to put this into usermods - and not into the `namelist_definition_blom.xml`? - since to my understanding the isopycnic bulkmixlayer setup isn't been further developed, they are currently mostly run in CMIP6-compatible mode anyway, right?

mvertens:
@TomasTorsvik - I agree with @jmaerz. Here is an alternative proposal:
- A new BLOM xml variable can be defined in `config_component.xml `-  say `BLOM_CMIP6`
- A new variable can be defined in `buildnml` - say `cmip6_setting`
`cmip6_setting = case.get_value("BLOM_CMIP6")`
- then namelist_definition_blom.xml can use this as follows:
```
  <entry id="edanis">
    <type>logical</type>
    <category>diffusion</category>
    <group>diffusion</group>
    <values>
      <value>.true.</value>
      <value "cmip6_setting">.false.</value>
    </values>
    <desc>If true, apply anisotropy correction to eddy diffusivity (l)</desc>
  </entry>
```
The above can modified to also have grid dependent settings.

jmaerz:
@mvertens , thanks, yes, that would be the other alternative, which could also make sense, if the isopycnic wil be further developed/used differently than with CMIP6 settings.

mvertens:
I think that an advantage with this approach is the there is only one xml variable that needs to be set - rather than multiple user_mod directories - potentially for different grids.

TomasTorsvik:
@jmaerz , @mvertens , @matsbn 
I can try to implement the `namelist_definition_blom.xml`, perhaps in a new PR as it is different from the approach in this PR.

jmaerz:
Hi @TomasTorsvik , since we briefly discussed closing stalled PRs: I feel this one is outdated with the settings:
```
./xmlchange BLOM_UNIT=cgs
./xmlchange BLOM_VCOORD=isopyc_bulkml
./xmlchange BLOM_CMIP6=TRUE
```
that can now be used in the latest tagged versions of release-1.6 to re-gain BLOM circulation behavior as in CMIP6 for `tnx1` and `tnx2` resolution (note that there are some bugfixes and extensions done in iHAMOCC, though). If there is need for other grid resolution, the `namelist_definition_blom.xml` would require an update, though.

TomasTorsvik:
@jmaerz - agree, I don't think this will be needed.

PR #571, opened 2025-05-16
DOC code of Jerry

### LINKED ISSUE #570: DOC code
Re-implement DOC code of Jerry.  I'll try to push this forward the next days.

@TomasTorsvik 
@JorgSchwinger 
@tjiputra

jmaerz:
@TomasTorsvik , @JorgSchwinger and @tjiputra,
this is the PR for Jerries DOC code. It implements the DOC model of Lønborg et al. 2018: Large Stimulation of Recalcitrant Dissolved Organic Carbon Degradation by Increasing Ocean Temperatures.  https://doi.org/10.3389/fmars.2017.00436

It currently supports both, `lkwrbioz_off` and the extended N-cycle (`use_extNcycle`), while the full support of the latest `use_river2omip` developments is still lacking (it is unclear, if some riverine OM should enter the new DOC classes, which could be implemented in `mo_apply_rivin.F90`).  

**Still needs to be done:**
- [x] currently, DOC remin could lead to negative O2, since the factor of 0.33 in the limitation is not summing up to one, when e.g. `use_river2omip` is switched on - also in combination with the new DOM-classes code. Potentially introduce pre-factors: `1*0.33` for default, `0.25*0.33` for DOC classes, `0.3*0.33` for `use_river2omip`, `0.16*0.33` for `use_river2omip`&`use_dom`.  Would this be reasonable - or is there a better distribution for the different classes/types of DOM?
- [x] provide some tuning parameters in the namelist
- [x] Read in initial conditions similar to WOA13
- [x] Rename logical switch from `use_dom` to `use_DOMclasses`?

Closes #570


jmaerz:
@JorgSchwinger and @TimotheeBrgs , I prepared the DOC code of Jerry as much as I can with respect to your new river2omip protocol. However, for full support of it when using Jerrys DOM code, it would be good, if you could have a look into the code and provide potentially suggestions, how to best implement it. Thus far, only the default model is fully supported, when switching on Jerrys DOM code. Implementation of the new DOM code is almost done, while some points are still open (see header of the draft PR), which I will try to take care of soon.

jmaerz:
Hi @tjiputra , @TomasTorsvik , @JorgSchwinger and @TimotheeBrgs , I open this PR for review, despite small expected changes to come in. Namely: addition of some tuning parameters to the bgc namelist. @tjiputra: could you provide a list, which parameters you want to have in the namelist so that they can easily be tuned via the `user_nl_blom`? I saw a number of hard-coded values in `mo_ocprod` - any of those that should become tunable? -if so, any propositions for names?

With regards to the river2omip support for the new DOM-classes code by Jerry: 
We could consider to leave this out of this PR and you guys can discuss this in the aftermath? - so that @TimotheeBrgs could implement this, if wanted/needed, if you agree on this?

tjiputra:
Tunable parameters in mo_param_bgc.F90:
      gammap   = 0.10        ! DOC_l exudation rate [day-1]
      gammapsl = 0.02        ! DOC_sl exudation rate [day-1]
      gammaz   = 0.06        ! DOC_l excretion rate [day-1]
      gammazsl = 0.03        ! DOC_sl excretion rate [day-1]
      alphasl  = 0.18        ! fraction of DOC_sl converted to DOC_sr []
      alphasr  = 0.19        ! fraction of DOC_sr converted to DOC_r  []

in mo_ocprod.F90 (for now):
=====
                bacfra=1.7e6*dtb*exp(-38.e3/(8.314*(ptho(i,j,k)+273.15)))*ocetra(i,j,k,idoc) !increased from 2e5 13082024
                 bacfrasl=5.0e7*dtb*exp(-58.e3/(8.314*(ptho(i,j,k)+273.15)))*ocetra(i,j,k,idocsl)
                 bacfrasr = 1.7e17*dtb*exp(-116.e3/(8.314*(ptho(i,j,k)+273.15)))*ocetra(i,j,k,idocsr)
                 bacfrar  = 5.0e26*dtb*exp(-179.e3/(8.314*(ptho(i,j,k)+273.15)))*ocetra(i,j,k,idocr)
=====

Suggest only the first constant above, with the following naming (supposedly to be added in mo_param_bgc.F90):
docl_remin = 1.7e6
docsl_remin = 5.0e7
docsr_remin = 1.7e17
docr_remin =  5.0e26

jmaerz:
Hi @tjiputra , thanks for coming up with the list - I'll implement this. Will you be preparing the input files similar to WOA-data? - this would allow to start from a reasonable spunup state.

jmaerz:
@tjiputra , while implementing, I realized that you provided `docl_remin=1.7e6`, while in your original code it seems to be: `1.5e6`. Not sure, which to implement as default - just let me know.

tjiputra:
> @tjiputra , while implementing, I realized that you provided `docl_remin=1.7e6`, while in your original code it seems to be: `1.5e6`. Not sure, which to implement as default - just let me know.

please use docl_remin=1.7e6

cime_config/namelist_definition_blom.xml
```diff
@@ -4668,6 +4678,16 @@
...
   </entry>
 
+  <entry id="gammapsl">
+    <type>real</type>
+    <category>bgcparams</category>
+    <group>bgcparams</group>
+    <values>
+      <value>None</value>
+    </values>
+    <desc>zooplankton parameter: 1/d -exudation rate</desc>
```
tjiputra: <desc>phytoplankton parameter: 1/d -DOC_sl exudation rate</desc>

cime_config/namelist_definition_blom.xml
```diff
@@ -4678,6 +4698,16 @@
...
   </entry>
 
+  <entry id="gammazsl">
+    <type>real</type>
+    <category>bgcparams</category>
+    <group>bgcparams</group>
+    <values>
+      <value>None</value>
+    </values>
+    <desc>zooplankton parameter: 1/d -excretion rate</desc>
```
tjiputra: zooplankton parameter: 1/d -DOC_sl excretion rate

tjiputra:
> Hi @tjiputra , thanks for coming up with the list - I'll implement this. Will you be preparing the input files similar to WOA-data? - this would allow to start from a reasonable spunup state.

Is there a script available to prepare this? Any suggestions? I could use CDO for horizontal interpolation, but not sure about vertical.

jmaerz:
Hi @tjiputra , there is the less documented CDO command  `intlevel3D` (or `intlevel` - the latter is easier for the regular WOA grid) that should allow for your use case. Otherwise, by providing some BLOM output and tweaked depth information, the remapping routine in `./utils/iHAMOCC/restart_remapping/` could also be used (but I fear that preparing for this in your case requires the same work as preparing it by using a simple combination of bilinear remapping and applying `intlevel3d` or likely just `intlevel`). I am not entirely sure, which input files you'll be using (which would allow to give better advice).

hamocc/mo_param1_bgc.F90
```diff
@@ -432,9 +442,24 @@ subroutine init_indices()
       itdoc_lc  = -1
       itdoc_hc  = -1
     endif
+    if (use_dom) then
+      i_dom=7
+      idocsl=i_base+i_iso+i_cfc+i_agg+i_nat_dic+i_bromo+i_extn+i_pref+i_shelfage+i_r2o+1
+      idocr =i_base+i_iso+i_cfc+i_agg+i_nat_dic+i_bromo+i_extn+i_pref+i_shelfage+i_r2o+2
```
tjiputra: better:
idocsr =i_base+i_iso+i_cfc+i_agg+i_nat_dic+i_bromo+i_extn+i_pref+i_shelfage+i_r2o+2
idocr =i_base+i_iso+i_cfc+i_agg+i_nat_dic+i_bromo+i_extn+i_pref+i_shelfage+i_r2o+3

tjiputra (commented):
Hi @jmaerz ,
I have reviewed all use_dom related changes.
Many thanks again for implementing these changes.

jerry

tjiputra:
> Hi @tjiputra , there is the less documented CDO command `intlevel3D` (or `intlevel` - the latter is easier for the regular WOA grid) that should allow for your use case. Otherwise, by providing some BLOM output and tweaked depth information, the remapping routine in `./utils/iHAMOCC/restart_remapping/` could also be used (but I fear that preparing for this in your case requires the same work as preparing it by using a simple combination of bilinear remapping and applying `intlevel3d` or likely just `intlevel`). I am not entirely sure, which input files you'll be using (which would allow to give better advice).

I plan on just extracting the DOClvls from one of the last annual outputs. Will try to do this when I get back. many thanks,

jmaerz:
Hi @tjiputra , the plan sounds reasonable - potentially consider using December monthly output particularly for the labile DOC (but I guess, it won't make much of a difference for the spinup). It should be straight forward to remap via cdo bilinear remapping and `intlevel` (no `intlevel3d` needed). 

Do you have suggestions on how to best make use of @TimotheeBrgs `river2omip` implementation to support your DOM classes in `mo_apply_rivin`? I wonder a bit, if there is any change at all needed, since with `river2omip`, the riverine component of DOC is anyway differently treated and only labile DOC input would overlap with your DOM classes.

tjiputra:
Hi @jmaerz ,

I put an example restart files here:
/cluster/projects/nn9560k/tjiputra/DOC_codes/restart_files
containing all four DOCs and four respective preformed tracers.

Regarding your mo_apply_riverin question, I will have a chat with Timothee on Friday to get familiar with what he does with river2omip. 

many thanks,

hamocc/mo_ncout_hamocc.F90
```diff
@@ -1770,6 +1858,24 @@ subroutine hamoccvardef(iogrp,timeunits,calendar,cmpflg)
       call ncdefvar3d(SRF_ATMBROMO(iogrp),cmpflg,'p',                           &
            &   'atmbromo','Atmospheric bromoform',' ','ppt',0)
     endif
+    if (use_dom) then
+      call ncdefvar3d(INT_EXUDL(iogrp),cmpflg,'p','exudlint',                   &
+           &   'Integrated phy exudation to DOC-L',' ','mol C m-2 s-1',0)
+      call ncdefvar3d(INT_EXUDSL(iogrp),cmpflg,'p','exudslint',                 &
+           &   'Integrated phy exudation to DOC-SL',' ','mol C m-2 s-1',0)
+      call ncdefvar3d(INT_EXCRL(iogrp),cmpflg,'p','excrlint',                   &
+           &   'Integrated zoo excretion to DOC-SL',' ','mol C m-2 s-1',0)
```
jmaerz: @tjiputra: Should this read DOC-L (l. 1867)

tjiputra: yes. thank you for noticing this.

jmaerz:
@TomasTorsvik , @tjiputra and @JorgSchwinger , I was now running a first test-year simulation with the new DOM code and after some last technical fixes, it looks good/reasonable: `/cluster/work/users/jma105/archive/chk-DOCjerry3_NOINYOC_T62_tn21_NorESM2.1.3-BLOMmaster` - I just started a resubmission of one year to also check, if the restart works as expected.

TomasTorsvik (approved):

TomasTorsvik:
@jmaerz - great that you also test the restart procedure.
I approve this PR now, but suggest that @tjiputra and @JorgSchwinger also provide a review.

jmaerz:
@TimotheeBrgs , thanks for pointing me to an issue in `mo_ocprod` - and `endif` was misplaced. That should be now fixed.

jmaerz:
Hi @tjiputra , @JorgSchwinger and @TomasTorsvik , to my understanding, the PR is now ready for merging. The only one thing that I am uncertain about is the small re-initialization section in `ini_param_biol` for the `use_DOMclasses` - I feel this doesn't really make sense, since some retuning is likely needed anyways. Thoughts on this @tjiputra ? - I would remove it and merge (those parameters can be put in via `user_nml_blom`, when running the model - note that there were some other parameters also changed since your model version) - and would leave the riverine input application to a future PR.  I leave the further tuning/more testing to @tjiputra since he knows what to expect and how to tune better - I suspect that with the hybrid coordinates, some changes become visible.

jmaerz:
Hi @JorgSchwinger and @tjiputra , apart from potentially removing the re-initialization section in `ini_param_biol`, is there anything else left that hinders to approve and merge the PR? I suspect/suggest the riverine input needs a bit of discussion and could come in as a separate PR (while I would be pragmatic and leave it as is - but leave this up to you).

jmaerz:
@JorgSchwinger , @tjiputra , I give this another ping for review - for further code updates, it would be nice being able to merge this soon, latest early next week (but ideally by tmw afternoon).

tjiputra (approved):
Looks good to me @jmaerz . Thank you very much.

hamocc/mo_ocprod.F90
```diff
@@ -418,8 +437,32 @@ subroutine ocprod(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,omask,ptho,pi_ph,psao,pp
...
+                ! R=8.314 J/mol/K (Universal gas constant)
+                ! Ea=38kJ/mol (Activation energy for DOC_L decay)
+                ! A_sl=7.96e9 (day-1) (Theoretical Kc in the absence of Ea)
+                ! R=8.314 J/mol/K (Universal gas constant)
+                ! Ea_sl=58kJ/mol (Activation energy for DOC_L decay)
+                bacfra   = docl_remin*exp(-38.e3/(8.314*(ptho(i,j,k)+tzero)))*ocetra(i,j,k,idoc) !increased from 2e5 13082024
+                !JT if (ocetra(i,j,k,iphosph) >5.e-8) then ! Nutrient thresholds for surface bacteria activity (Anton Salgado, pers. comm.)
+                bacfrasl = docsl_remin*exp(-58.e3/(8.314*(ptho(i,j,k)+tzero)))*ocetra(i,j,k,idocsl)
+                bacfrasr = docsr_remin*exp(-116.e3/(8.314*(ptho(i,j,k)+tzero)))*ocetra(i,j,k,idocsr)
+                bacfrar  = docr_remin*exp(-179.e3/(8.314*(ptho(i,j,k)+tzero)))*ocetra(i,j,k,idocr)
```
JorgSchwinger: For consistency, use the above temperature range (variable `temp`)? There we have `temp = min(40.,max(-3.,ptho(i,j,k)))`

JorgSchwinger (commented):
@jmaerz Thanks for integrating this into the current code! I have two comments:

1) There is now no possibility to start the DOM-code from scratch, the current initial conditions are always used and it won't be possible to start the model without. However, we would want to have this capability.

2) I'm not so happy with the fact that the preformed DOM tracers are handled differently than the other preformed tracers. If I switch off the option for preformed tracers, I would expect that this switches them off altogether. However the DOM preformed tracers are now ALWAYS on, and I do not see a particular reason for this.

jmaerz:
@JorgSchwinger and @tjiputra , wrt 2. (usage of preformed DOM tracers): I had this discussion with Jerry as well, while he was in favor keeping the preformed DOM tracers handled with the regular DOM tracers. Please agree on the treatment among you two. 

Wrt to restarting from scratch: I will have a look, while it is always possible to generate the file and fill them with initial conditions of your choice...

tjiputra:
Regarding point (2), i agree with @JorgSchwinger that it's awkward that they are separated, and would suggest combining them. My point during a discussion with @jmaerz is that i think preformed tracers should all be default turned on. Without proper interior ocean values, they are quite useless. And in order to have reasonable values, they need to be turned on throughout all spin-ups. I do understand that they will slow down the execution time. In CMIP6 version, i ended up having to run an extended spin up and new set of scenarios in order to use the preformed tracers.

jmaerz:
Ok, this means to combine the switches in the way:
```
if (use_pref_tracers .and. use_DOCclassses) then
! perform preformed DOM tracer code
endif
```
That's how I also would have supported/done it - great to agree on this. I'll try to implement this soon.

hamocc/mo_ocprod.F90
jmaerz: Done.

jmaerz:
@JorgSchwinger , I now separated out the preformed DOC tracers.
With regards to DOM initialization: I don't see such a switch or similar for PO4, NO3, etc. either. Not sure, if we really need such a switch (I am not foreseeing long spinup times needed for such simulations - and if so, the memory of the initial distribution currently set is likely anyway gone). If still requested, such switch could come in a separate PR (to make this clean for -all?- tracers). Easiest solution right now is copying the input file, exchange values to desired initial values and use the modified file as input, if needed.

JorgSchwinger:
But the difference for PO4 and NO3 is that we start from observations. The DOM initialization depends on the model physics and tuning (e.g. I actually fear that the current initial file is not so super useful since it is based on simulations with isopycninc coordinates and a quite different tuning (lkwrbiozoff=flase). I would assume that one needs to create new initial conditions with each major update of the model (and at least the DOCR should take quite some time to spinup - at least that was Jerry's point)

jmaerz:
@JorgSchwinger , I now introduced an additional switch to facilitate either starting from former simulations from Jerry or start from scratch. However, since for these DOM simulations long spinup times are needed, while the interest is mainly in the upper 1000m ocean in CMIP-type simulations, I would encourage to start the simulations always from long-spun-up simulations - here Jerrys initial simulations - despite changes in ocean physics, etc. - on the longer run, the shorter lived DOM classes will adjust quickly anyway, while the more refractory classes will likely have lesser drift with further spin-up.  Once the CMIP7 spinup is done with (potentially) these DOM classes, one could write new DOM classes initial files for further future simulations. It would be good to merge this soon.

JorgSchwinger (approved):

jmaerz:
Initialization works now as wished/expected. I merge this now.

PR #572, opened 2025-05-19
Add output for dust and soluble iron deposition fluxes

JorgSchwinger:
This PR adds output for the fluxes of total dust and of soluble iron to the surface ocean. It is a first step for introducing a new iron-deposition input data set, that will be done in a separate PR.

### DIFF
--- a/hamocc/mo_apply_fedep.F90
+++ b/hamocc/mo_apply_fedep.F90
@@ -55,15 +55,18 @@ contains
     real    :: dustinp
 
     ! dust flux from the atmosphere to the surface layer; dust fields are
-    ! monthly mean values (kg/m2/month - assume 30 days per month here)
-    ! dissolved iron is a fixed fraction (typically 3.5%), and immediately released
-
+    ! monthly mean values (kg/m2/month - assume 30 days per month here).
+    ! The iron content of dust is a fixed fraction (typically 3.5%), of 
+    ! which a fixed fraction (typically 1%) is soluble (bioavailable) and immediately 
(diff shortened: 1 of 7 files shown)


TomasTorsvik (approved):
@JorgSchwinger - thanks, looks fine to me!

hamocc/mo_apply_fedep.F90
```diff
@@ -62,9 +62,10 @@ subroutine apply_fedep(kpie,kpje,kpke,pddpo,omask,dust)
     do j = 1,kpje
       do i = 1,kpie
         if(omask(i,j) > 0.5) then
-          dustinp = dust(i,j) / 30. * dtb / pddpo(i,j,1)
-          ocetra(i,j,1,ifdust) = ocetra(i,j,1,ifdust) + dustinp
-          ocetra(i,j,1,iiron) = ocetra(i,j,1,iiron) + dustinp * perc_diron
+          dustflx(i,j,itdust)  = dust(i,j) / 30. * dtb / pddpo(i,j,1)
+          dustflx(i,j,isfe)    = dust(i,j) / 30. * dtb / pddpo(i,j,1) * perc_diron
```
jmaerz: Could be simplified to `dustflx(i,j,isfe)  = dustflx(i,j,itdust)*perc_diron` (and in the next step, when exchanging the input, vice versa). Maybe add a comment on unit convesrion here (i.e. it is percent-iron  * kmol/kg factor, which is included in perc_diron).

jmaerz (approved):
Thanks for bringing this output in.

PR #574, opened 2025-05-20
Make lkwrbioz=true and lto2remin=true default values, new tuning

JorgSchwinger:
This PR sets the options lkwrbioz and lto2remin to .true. by default and introduces a new default tuning. The new tuning roughly works for the new default settings in both 1 and 2 degree resolution with CORE and JRA forcing. Further fine-tuning will be necessary.

### DIFF
--- a/hamocc/mo_param_bgc.F90
+++ b/hamocc/mo_param_bgc.F90
@@ -300,10 +300,10 @@ module mo_param_bgc
   ! Zooplankton parameters
   !********************************************************************
   real, protected :: grami      = 1.e-10          ! kmol/m3 - i.e. 1e-5 mmol P/m3 minimum concentration of zooplankton
-  real, protected :: bkzoo      = 8.e-8           ! kmol/m3 - i.e. 0.08 mmol P/m3 half saturation constant
+  real, protected :: bkzoo      = 1.e-7           ! kmol/m3 - i.e. 0.08 mmol P/m3 half saturation constant
 
   !ik addded parameter definition; taken from OCPROD.F
-  real, protected :: grazra     = 1.2             ! 1/d - grazing rate
+  real, protected :: grazra     = 1.5             ! 1/d - grazing rate
   real, protected :: spemor     = 3.*1.e6         ! 1/d - mortality rate
   real, protected :: gammap     = 0.04            ! 1/d - exudation rate
   real, protected :: gammaz     = 0.06            ! 1/d - excretion rate
@@ -314,7 +314,7 @@ module mo_param_bgc
   !********************************************************************
   ! Shell production (CaCO3 and opal) parameters
   !********************************************************************
-  real, protected :: bkopal     = 5.e-6           ! kmol/m3 - i.e. 4.0 mmol Si/m3 half saturation constant
+  real, protected :: bkopal     = 1.e-5           ! kmol/m3 - i.e. 10 mmol Si/m3 half saturation constant
   real, protected :: rcalc                        ! calcium carbonate to organic phosphorous production ratio
   real, protected :: ropal                        ! opal to organic phosphorous production ratio
   real, protected :: calmax                       ! maximum CaCO3 production fraction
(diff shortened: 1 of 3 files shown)


TomasTorsvik (approved):

TomasTorsvik:
@JorgSchwinger - Thanks for making the PR, I think this should be fine.

Do you think to tune further with CORE and JRA forcing, or leave this until we have coupled simulations? At the moment I have not put in further BLOM/iHAMOCC updates to alpha04, as not to interfere with work for CAM updates.

jmaerz:
Hi @JorgSchwinger , have you checked that changes in `namelist_definition_blom.xml` are enough? - for safety, I would set it also in `mo_control`. Further, I am a bit uncertain, how to best organize the handling of tuning parameters... - via hard-coding as you suggested here or via some intermediate user nmls/usermods. Maybe we can at least keep somehow track of the old settings so that it is easier, re-using the code for other applications (such as CMIP6-like settings)?

JorgSchwinger:
> Hi @JorgSchwinger , have you checked that changes in `namelist_definition_blom.xml` are enough? - for safety, I would set it also in `mo_control`. Further, I am a bit uncertain, how to best organize the handling of tuning parameters... - via hard-coding as you suggested here or via some intermediate user nmls/usermods. Maybe we can at least keep somehow track of the old settings so that it is easier, re-using the code for other applications (such as CMIP6-like settings)?

Hi, @jmaerz yes, `namelist_definition_blom.xml` is enough, but I will nevertheless also update `mo_control`

I have not the time/energy right now to provide a different solution. My take is that if someone wants to run CMIP6-like with this version of BLOM will have to go and make alternative parameter setting via  `namelist_definition_blom.xml`.

jmaerz:
@TomasTorsvik and @JorgSchwinger , any chance to just put the old parameters in a `user_nml_blom_CMIP6`-file in a folder like `./utils/iHAMOCC/tmp/`? That can be removed again later?

JorgSchwinger:
@jmaerz the question is also what you actually want to run. With/without sediment spinup? With/without correction for CaCo3 dissolution? With the old (bad) sediment dissolution rates? Reverting to krwbioz=false (this possibility will soon be removed)? 

I think we are approaching a point where this doesn't make sense any longer.

hamocc/mo_param_bgc.F90
```diff
@@ -656,8 +656,8 @@ subroutine ini_param_biol()
       ropal  = 10.5        ! opal to organic phosphorous production ratio
       calmax = 0.20
     else if (use_WLIN) then
-      rcalc  = 33.         ! calcium carbonate to organic phosphorous production ratio
-      ropal  = 45.         ! opal to organic phosphorous production ratio
+      rcalc  =  8.         ! calcium carbonate to organic phosphorous production ratio
+      ropal  = 70.         ! opal to organic phosphorous production ratio
```
jmaerz, line 660: Hi @JorgSchwinger , just for information this (`ropal`) is a pretty high value compared to literature (while not totally out of range). Typical range likely between ca. 4-51, mean ca. 16, maximum ca. 99 according to Nelson et al. 1995.

JorgSchwinger: Hi @jmaerz , thanks for checking this - I was already wondering. 

It's the only possibility though to reduce Southern Ocean (and northern high latitude) silicate - it is otherwise way too high despite phosphate being okayisch. Probably a consequence of fixed stochiometry?

jmaerz:
... which is why I am uncertain, whether hard-coded values make sense at the moment (and usermods could be used for more flexibility in this, while sharing information among us). Nevermind, I am trying to keep track of all those things for my own purposes - which is increasingly difficult, which is why I was asking to put such thing at least in a tmp folder. I also wondered with respect to the next tag still entering NorESM2.3?() and what it should entail in that respect.

jmaerz (approved):
Thanks for the tuning efforts.

JorgSchwinger:
> ... which is why I am uncertain, whether hard-coded values make sense at the moment (and usermods could be used for more flexibility in this, while sharing information among us). Nevermind, I am trying to keep track of all those things for my own purposes - which is increasingly difficult, which is why I was asking to put such thing at least in a tmp folder. I also wondered with respect to the next tag still entering NorESM2.3?() and what it should entail in that respect.

If we want to run NorESM2.3 CMIP6-like, we can just use the release that is currently foreseen for NorEMS2.3? At least in my head that is/was the plan?

JorgSchwinger:
> Do you think to tune further with CORE and JRA forcing, or leave this until we have coupled simulations? At the moment I have not put in further BLOM/iHAMOCC updates to alpha04, as not to interfere with work for CAM updates.

I think it would be really good, if it is possible, to put in these changes to see how different the coupled model is from JRA-forced runs. It won't affect the coupled model much except through DMS fluxes, but I think the changes here are reasonably small.

jmaerz:
Hi @JorgSchwinger, maybe @TomasTorsvik can clarify, but to my understanding, the next BLOM tag is the last tag entering NorESM2.3.

JorgSchwinger:
> Hi @JorgSchwinger, maybe @TomasTorsvik can clarify, but to my understanding, the next BLOM tag is the last tag entering NorESM2.3.

Oh, okay - I was not aware of this. In this case there is anyway the problem of bug-fixes that have a larger impact. I think so far there have been none that really change things (for CMIP6-like settings), but the CaCO3 dissolution really give you very different deeper ocean alkalinity.

jmaerz:
Hio @JorgSchwinger , can you at least provide, if this was tested/run with the latest master-BLOM settings? - or with something else? Or in other words: are the new iHAMOCC parameters consistent with the current BLOM physics master version?

JorgSchwinger:
Yes, this is all for current master. I'll also run for a bit longer now and make results available

TomasTorsvik:
@jmaerz , @JorgSchwinger - I have not checked backwards compatibility with NorESM2.3 after `v1.8.0`, i.e. if we can get something close to CMIP6 output when running with isopycnic coordinates. We still have the idea to keep MCT for a `v1.10` tag, so that it can technically build with NorESM2.3.

Issue #581, opened 2025-05-24
Add stream functionality for the new GESAMP dust/iron deposition data set


JorgSchwinger:
A new dust/iron deposition data set has been added and tested. So far the new input data is optional only, but going forward this should become the default.

First implementation and tests have been done using the "traditional" reading method, but now we should also add this data to the dust-stream functionality. For this I need some assistance, please @mvertens 

1) We need a mesh file for the data set. The input data is on a regular 1x1 degree grid and can be found here:
 
`/cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_GESAMP2018_r360x180_20250515.nc`

How can I create the required mesh file?

2) The main difference between the old and the new scheme is that there are now two variable in the file/stream that need to be read. Looking at `ocn_stream_dust.F90` it is not entirely clear to me where I can add the 2nd variable? Can I make an array and have 2 names in `stream_dust_varname`? Or will I have to make `stream_dust_varname1` and `stream_dust_varname2`?

mvertens:
@JorgSchwinger - happy to help with this.

1. meshfile - @matsbn has already created a 1 degree mesh - see
`/cluster/projects/nn9560k/matsbn/WOA_mesh/WOA_1.00_degree_ESMFmesh_20250506_cdf5.nc
`You need to have a SCRIP file to start with - and the directions for how he did this is here
`/cluster/projects/nn9560k/matsbn/WOA_mesh/README.txt`
2. variable names - in ocn_stream_dust.F90
           stream_fldlistFile  = (/stream_dust_varname/),          &
           stream_fldlistModel = (/stream_dust_varname/),        &
stream_fldlistFile and stream_fldListModel is just a one dimensional array of variable names. A better implementation would be to replace stream_dust_varname with stream_dust_varnames(:) and read those values in from the namelist.

Let me try doing this based on your #575.  
There are other subtleties here that I can help resolve - particularly the interp routine also needs to be changed to cycle over the input field names if they are more than one, etc.

I've implemented this in a draft PR #582 and am in the process of testing.

jmaerz:
Since #582 has been merged, I am closing this issue.

PR #582, opened 2025-05-25
Add stream functionality for the new GESAMP dust/iron deposition data set 

mvertens:
This PR implements #581 
- I have converted the lons in dustdep_GESAMP2018_r360x180_20250515.nc to be -180->180 and created a new file with this dustdep_GESAMP2018_r360x180_20250526.nc. This permitted the plots below for dustflx_tot and dustflx_sfe to look reasonable.
- Successfully ran using noresm3_0_alpha03c:
```
- ERS_D_Ld3.T62_tn14.NOINYOC.betzy_gnu
- SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel
```

### DIFF
--- a/hamocc/mo_read_fedep.F90
+++ b/hamocc/mo_read_fedep.F90
@@ -114,21 +114,21 @@ contains
       case('mahw2006')
         ! Dust/iron deposition using the Mahowald et al. 2006 model data.
         ! The variable 'DUST' contains the total dust-flux in kg/m2/month; this
-        ! is converted to kmol fe/m2/s of bioavailable (soluble) iron using the molar 
-        ! weight of iron (mw_fe=55.85) and assuming that dust contains a fraction 
-        ! 'frac_ironindust' of iron, and that soluble iron is a fraction 'frac_soliron' of 
+        ! is converted to kmol fe/m2/s of bioavailable (soluble) iron using the molar
+        ! weight of iron (mw_fe=55.85) and assuming that dust contains a fraction
+        ! 'frac_ironindust' of iron, and that soluble iron is a fraction 'frac_soliron' of
         ! total iron. A tuning factor 'fetune' is applied to the soluble fraction of iron.
-        ! Note the conversion kg/m2/month -> kg/m2/s is assuming 30 days per month.	  
+        ! Note the conversion kg/m2/month -> kg/m2/s is assuming 30 days per month.
         call read_netcdf_var(ncid,'DUST',dustflx_tot(1,1,1),12,0,0)
         dustflx_tot(:,:,:) = dustflx_tot(:,:,:)/30.0/sec_per_day
         dustflx_sfe(:,:,:) = dustflx_tot(:,:,:)*frac_ironindust*frac_soliron/mw_fe*fetune
 
       case('GESAMP2018')
         ! Iron deposition using data from the GESAMP atmospheric iron deposition
-        ! model intercomparison study (Myriokefalitakis et al. 2018, 
+        ! model intercomparison study (Myriokefalitakis et al. 2018,
         ! doi:10.5194/bg-15-6659-2018).
-        ! The variables 'TFe' and 'LFe' contain the total and soluble (labile) iron 
-        ! fluxes in kg/m2/s; soluble iron is converted to kmol Fe/m2/s using the molar weight of   
+        ! The variables 'TFe' and 'LFe' contain the total and soluble (labile) iron
+        ! fluxes in kg/m2/s; soluble iron is converted to kmol Fe/m2/s using the molar weight of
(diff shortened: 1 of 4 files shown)


hamocc/mo_read_fedep.F90
```diff
@@ -195,34 +195,42 @@ subroutine get_fedep(kpie,kpje,kbnd,kplmon,dust,dust_stream)
...
             end do
           end do
-       
+
+        case('GESAMP2018')
+          do j=1,kpje
+            do i=1,kpie
+              ! note that the stream is read from a file where units already are in kg/m2/s
+              dust(i,j,itdust) = dust_stream(i,j,1)
+              dust(i,j,isfe)   = dust_stream(i,j,2)*frac_ironindust*frac_soliron/mw_fe*fetune
```
JorgSchwinger: The conversion used her for the GESAMP data has to be different than for the mahw2006 data. The GESAMP data already has the soluble fraction of iron (in the field LFe), and the field TFe is total iron (not dust!), so the dust must be "backcalculated" by assuming a fraction of iron in dust. So these two lines should read:

```
dust(i,j,itdust) = dust_stream(i,j,1)/frac_ironindust
dust(i,j,isfe)   = dust_stream(i,j,2)/mw_fe*fetune
```

(compare to lines 137-138 above)

JorgSchwinger (approved):
Dear @mvertens thanks a lot for implementing this! There is an issue regarding the conversion of the stream fields to the input, otherwise this looks good and I approve this.

hamocc/mo_read_fedep.F90
mvertens: @JorgSchwinger - thanks for these corrections.

mvertens:
@JorgSchwinger - there is a problem in using the following mesh file/data combination:
  STREAM_DUST_DATA_FILENAME = '/cluster/shared/noresm/inputdata/ocn/blom/bndcon/dustdep_GESAMP2018_r360x180_20250515.nc'
  STREAM_DUST_MESH_FILENAME = '/cluster/projects/nn9560k/matsbn/WOA_mesh/WOA_1.00_degree_ESMFmesh_20250506_cdf5.nc'
The issue is that the mesh file goes from -180->180 in longitude whereas the data goes from 0->360.
I think it might be simples to convert the data file to go from -180->180 (this was the case for all the WOA data that @matsbn was using). Would that be simple to do? I can try to contact the ESMF  

@JorgSchwinger - I found an easy way to convert the lon:
cdo sellonlatbox,-180,180,-90,90 dustdep_GESAMP2018_r360x180_20250515.nc dustdep_GESAMP2018_r360x180_20250515_convert.nc
I'm going to try this and see if it gives correct results.

mvertens:
Running the case `SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel`:

Attached are two plots for dustflx_tot using the original and new (using the same range)
Screenshot 2025-05-27 at 11 37 16 AM
Below are two plots for dustflx_sfe using the original and new
Screenshot 2025-05-27 at 11 41 11 AM

mvertens:
@TomasTorsvik - is there a baseline for the lastest master commit? If I run the aux_blom_noresm test suite what baseline should I expect bfb with if I am not using the dustdep_GESAMP2018_r360x180_20250515.nc data.

mvertens:
@JorgSchwinger - are you okay with the above plot comparisons?

JorgSchwinger:
> @JorgSchwinger - are you okay with the above plot comparisons?

Dear @mvertens, thanks for implementing and testing this! 

Just for clarification: the plots above are those from the first time-step? If, so this means that the nuopc-field has been interpolated between December and January, whereas the original is the January field, is this correct?

mvertens:
@JorgSchwinger - yes this is from the first time step. Your interpolation assumption is correct.

JorgSchwinger:
> @JorgSchwinger - yes this is from the first time step. Your interpolation assumption is correct.

Ok, then we don't expect a very close match, and the plots above look totally fine! When this is merged I will also do a bit longer test-run. If the interpolated vs. old methods require different fine-tuning.

JorgSchwinger:
@mvertens 

Was there any other issue with the input file? I have updated my script with the cdo command that you provided above (I will submit this change since we are collecting the scripts used for preparing input data in the BLOM repository)

mvertens:
@JorgSchwinger - there was no other issue with the input file.

mvertens:
@TomasTorsvik @JorgSchwinger - what other testing should I do for this PR?
Do you want me to run the aux_blom_noresm test suite and compare to a baseline?

JorgSchwinger:
I think this can be merged now. It is still not the default setting, and I plan to do a longer test run also including this development next week.

PR #534, opened 2025-04-14
Implement R2OMIP protocol

TimotheeBrgs:
Hello, the aim of this draft PR is to implement the River2Ocean MIP protocol partially described [here](https://zenodo.org/records/14889524) along with the input data that will be used. It seems that the full protocol is not publicly available yet so I am not sure if I can put it here. If the link above is not enough, please ask by email.

I implemented this by keeping the original setup operational, and adding a switch (`use_river2omip`) that activates the R2OMIP implementation. It is suggested that this will eventually fully replace the current riverine input implementation considering the state of the art that such riverine input data represents. To be discussed.

Two state variables related to the terrestrial DOC are created: `tDOC_lc` and `tDOC_hc` with `lc` and `hc` for low and high-carbon content due to the respective C:N:P ratios of 276:25:1 and 2583:103:1 ([Meybeck et al 1982](https://doi.org/10.2475/ajs.282.4.401), [Compton et al., 2000](https://doi.org/10.2110/pec.00.66.0021), [Ma et al., 2025](https://doi.org/10.5194/esd-2024-29)).

The main changes in the distribution of the riverine inputs to the model pools are:

- `rivDOCl -> DIC`  (with `l` for labile). Note that the released N & P from this assumed instantaneous remineralisation are already included in the data. `rivDOCl` stoichiometry is assumed to be the same as the marine model DOC. Alkalinity change from instantaneous remineralisation is ignored. Iron release from this remineralisation is assumed included in the former riverine iron input.
- `riv_DOCsl -> tDOC_hc` (with `sl` for semi-labile)
- `riv_POC -> tDOC_lc` (assumed instantaneous dissolution)

Therefore, the marine DOC and DET model pools are no longer supplied by riverine inputs.

Both tDOC pools degrade into DIC over time following a degradation time scale of 1.5 year so far. This remineralisation releases N and P according to the ratios stated above. It also consume oxygen and changes alkalinity following [Paulmier et al. (2009)](https://doi.org/10.5194/bg-6-923-2009). Iron is also released by tDOC remineralisation using the same iron stoichiometry as the former DOC pool.

The units provided in the riverine input file are unchanged: DOC, slDOC and DET must be expressed in kmol P m-2 yr-1.

The data source for riverine silicate and iron inputs stay unchanged.

PR #583, opened 2025-05-25
map river nutrients to blom mesh using rof->ocn mapping files

mvertens:
This PR implements mapping river nutrients on a  0.5 x 0.5 "runoff grid" and mapping the nutrients to the blom grid using the offline mapping files for mapping rof->ocn.

The river nutrient file is here
/cluster/shared/noresm/inputdata/ocn/blom/bndcon/river_nutrients_GNEWS2000c00_r05_20250220.nc

This has to be tested with #534. @JorgSchwinger - can the data for this new riverine input also be generated on the 0.5x0.5 grid - so that I can test with the updated dataset?


JorgSchwinger:
> This has to be tested with #534. @JorgSchwinger - can the data for this new riverine input also be generated on the 0.5x0.5 grid - so that I can test with the updated dataset?


I have copied a file for the R2OMIP river input data (same format, same grid as GNEWs but contains one additional variable) to 

`/cluster/shared/noresm/inputdata/ocn/blom/bndcon/river_nutrients_R2OMIPpi_r05_20250527.nc`

Please let me know if there are any problems with the file.

drivers/nuopc/ocn_map_river_nutrients.F90
```diff
@@ -0,0 +1,148 @@
...
+      rivflx(:,:,:) = 0.0
+
+      ! Loop over the input fields
+      do nfld = 1,size(fldlist_input)
+         ! Extract the field from fldbun_blom
+         call ESMF_FieldBundleGet(fldbun_blom, fieldName=trim(fldlist_input(nfld)), field=field_blom, rc=rc)
+         if (chkerr(rc,__LINE__,u_FILE_u)) return
+         call field_getfldptr(field_blom, fldptr1=dataptr1d, rc=rc)
+         if (chkerr(rc,__LINE__,u_FILE_u)) return
+         fldlist_input = (/'DIN','DIP','DSi','DIC','Fe ','DOC','DET'/)
```
jmaerz, line 109: Hi @mvertens , why do you repeat the `fldlist_input` here (see l. 54)?

drivers/nuopc/ocn_map_river_nutrients.F90
```diff
@@ -0,0 +1,148 @@
...
+      real(r8), pointer      :: dataptr1d(:)
+      integer                :: nlev
+      integer                :: nfld
+      !-------------------------------------------------------------------------------
+
+      rc = ESMF_SUCCESS
+
+      filename = '/cluster/shared/noresm/inputdata/ocn/blom/bndcon/river_nutrients_GNEWS2000c00_r05_20250220.nc'
+      mesh_input_file = '/cluster/shared/noresm/inputdata/share/meshes/r05_nomask_c110308_ESMFmesh.nc'
+      fldlist_input = (/'DIN','DIP','DSi','DIC','Fe ','DOC','DET'/)
```
jmaerz, line 55: This currently seems to only support the 'old' river input file - for the `river2omip` the `slDOC` seems to be missing.

drivers/nuopc/ocn_map_river_nutrients.F90
```diff
@@ -0,0 +1,148 @@
...
+            index = 2
+         else if (trim(fldlist_input(nfld)) == 'DSi') then
+            index = 3
+         else if (trim(fldlist_input(nfld)) == 'DIC') then
+            index = 4
+         else if (trim(fldlist_input(nfld)) == 'Fe') then
+            index = 5
+         else if (trim(fldlist_input(nfld)) == 'DOC') then
+            index = 6
+         else if (trim(fldlist_input(nfld)) == 'DOT') then
```
jmaerz, line 123: `== 'DET'` ?!

drivers/nuopc/ocn_map_river_nutrients.F90
mvertens: That is correct. @JorgSchwinger just provided a new input dataset that I can use and I'll add that field.

drivers/nuopc/ocn_map_river_nutrients.F90
mvertens: Good catch. I fixed this.

drivers/nuopc/ocn_map_river_nutrients.F90
mvertens: This is now fixed as well. Thanks.

TomasTorsvik (approved):
@mvertens - Thanks, looks good to me!

mvertens:
> > This has to be tested with #534. @JorgSchwinger - can the data for this new riverine input also be generated on the 0.5x0.5 grid - so that I can test with the updated dataset?
> 
> I have copied a file for the R2OMIP river input data (same format, same grid as GNEWs but contains one additional variable) to
> 
> `/cluster/shared/noresm/inputdata/ocn/blom/bndcon/river_nutrients_R2OMIPpi_r05_20250527.nc`
> 
> Please let me know if there are any problems with the file.

@JorgSchwinger - thanks for providing this. I don't see an entry for DET in this file. Yet in mo_read_rivin.F90 the following is called - and DET is always read in.

```
    call ncfopn(trim(rivinfile),'r',' ',1,iotype)
    call ncread('DIN',riv_DIN2d,dummymask,0,0.)
    call ncread('DIP',riv_DIP2d,dummymask,0,0.)
    call ncread('DSi',riv_DSI2d,dummymask,0,0.)
    call ncread('DIC',riv_DIC2d,dummymask,0,0.) ! It is actually alkalinity that is observed
    call ncread('Fe' ,riv_DFe2d,dummymask,0,0.)
    call ncread('DOC',riv_idoc2d,dummymask,0,0.)
    if (use_river2omip) then
      call ncread('slDOC',riv_itdoc2d,dummymask,0,0.)
    else
      riv_itdoc2d = 0
    endif
    call ncread('DET',riv_idet2d,dummymask,0,0.)
    call ncfcls
```
In addition, is BLOM master - what file is used if use_river2omip is .true.? Is this currently just set in user_nl_blom?

JorgSchwinger:
@mvertens, yes, you are right the DET field is incorrectly named 'POC' in the file

`/cluster/shared/noresm/inputdata/ocn/blom/bndcon/river_nutrients_R2OMIPpi_r05_20250527.nc`

@TimotheeBrgs can you please fix this such that we have identical names for the GNEWS and the R2OMIP cases?

(For the interpolated files I fixed this in the interpolation-script, but it would be better to have this correct from the beginning). 

And yes, if you use R2OMIP, you currently you need to specify the input file manually in the user namelist - something that has to be fixed once this option is tested and finalized.

TimotheeBrgs:
Hi @mvertens I have now fixed the wrong naming of the DET variable. I replaced the former file with the one below. Hoping that everything else is fine.

`/cluster/shared/noresm/inputdata/ocn/blom/bndcon/river_nutrients_R2OMIPpi_r05_20250604.nc`

mvertens:
@TimotheeBrgs - Thanks!

PR #589, opened 2025-05-28
Various fixes related to riverine labile DOC and inventories

### LINKED ISSUE #588: Nutrients of irdoc in river2omip?
Hi @TimotheeBrgs and @JorgSchwinger 
sorry to bug on this again, but I have a feeling that at the end, there were too many comments and changes in the files, when PR #534 was merged (as in: I lost sight of which comment was handled, which wasn't).

Correct me, if I am wrong, but I just looked in `mo_rivin_apply.F90` again and I am a bit wondering (or I miss it), where the nutrients of the remineralized `irdoc` are entering? - I would have expected it to also enter in PO4, NO3 and Fe and not only the DIC pool? Please also check `mo_inventory.F90` in addition - there, the fluxes are somewhat depicted, but I feel with wrong `rnit` (l. 484)?!

Maybe re-check with the river2OMIP protocol, what is stated there (as far as I understand, only the alkalinity changes due to this remineralization should not be accounted for, right?)

TimotheeBrgs:
PR linked to issue #588 where Jöran spotted inconsistencies.

Closes #588

### DIFF
--- a/hamocc/mo_inventory_bgc.F90
+++ b/hamocc/mo_inventory_bgc.F90
@@ -481,9 +481,9 @@ contains
              &                   - srivflux(iralk) ! no DIN & DIP substraction because alkalinity
                                                    ! changes due to instantaneous remineralisation 
                                                    ! of riverine DOC are ignored
-        totalnitr   = totalnitr  - (srivflux(irdoc)+srivflux(irtdoc))*rnit_tdochc                  &
-             &                   - srivflux(irdet)*rnit_tdoclc - srivflux(irdin)
-        totalphos   = totalphos  - srivflux(irdoc)-srivflux(irtdoc)-srivflux(irdet)-srivflux(irdip)
+        totalnitr   = totalnitr  - srivflux(irtdoc)*rnit_tdochc-srivflux(irdet)*rnit_tdoclc        &
+             &                   - srivflux(irdin)
+        totalphos   = totalphos  - srivflux(irtdoc)-srivflux(irdet)-srivflux(irdip)
         totaloxy    = totaloxy   - (srivflux(irdoc)+srivflux(irtdoc))*(-49.5)                      &
              &                   - srivflux(irdet)*(-10.5)                                         &
              &                   - srivflux(irdin)*1.5 - srivflux(irdip)*2.                        &
--- a/hamocc/mo_inventory_bgc.F90
+++ b/hamocc/mo_inventory_bgc.F90
@@ -481,10 +481,10 @@ contains
              &                   - srivflux(iralk) ! no DIN & DIP substraction because alkalinity
                                                    ! changes due to instantaneous remineralisation 
                                                    ! of riverine DOC are ignored
-        totalnitr   = totalnitr  - srivflux(irdoc)*rnit-srivflux(irtdoc)*rnit_tdochc               &
+        totalnitr   = totalnitr  - (srivflux(irdoc)+srivflux(irtdoc))*rnit_tdochc                  &
              &                   - srivflux(irdet)*rnit_tdoclc - srivflux(irdin)
         totalphos   = totalphos  - srivflux(irdoc)-srivflux(irtdoc)-srivflux(irdet)-srivflux(irdip)
-        totaloxy    = totaloxy   - srivflux(irdoc)*(-24.) - srivflux(irtdoc)*(-49.5)               &
+        totaloxy    = totaloxy   - (srivflux(irdoc)+srivflux(irtdoc))*(-49.5)                      &
              &                   - srivflux(irdet)*(-10.5)                                         &
              &                   - srivflux(irdin)*1.5 - srivflux(irdip)*2.                        &
              &                   - (srivflux(iralk)+srivflux(irdin)+srivflux(irdip))
--- a/hamocc/mo_inventory_bgc.F90
+++ b/hamocc/mo_inventory_bgc.F90
@@ -484,8 +484,7 @@ contains
         totalnitr   = totalnitr  - srivflux(irtdoc)*rnit_tdochc-srivflux(irdet)*rnit_tdoclc        &
              &                   - srivflux(irdin)
         totalphos   = totalphos  - srivflux(irtdoc)-srivflux(irdet)-srivflux(irdip)
-        totaloxy    = totaloxy   - (srivflux(irdoc)+srivflux(irtdoc))*(-49.5)                      &
-             &                   - srivflux(irdet)*(-10.5)                                         &
+        totaloxy    = totaloxy   - srivflux(irtdoc)*(-49.5)-srivflux(irdet)*(-10.5)                &
              &                   - srivflux(irdin)*1.5 - srivflux(irdip)*2.                        &
              &                   - (srivflux(iralk)+srivflux(irdin)+srivflux(irdip))
 
--- a/hamocc/mo_inventory_bgc.F90
+++ b/hamocc/mo_inventory_bgc.F90
@@ -485,8 +485,9 @@ contains
              &                   - srivflux(irdin)
         totalphos   = totalphos  - srivflux(irtdoc)-srivflux(irdet)-srivflux(irdip)
         totaloxy    = totaloxy   - srivflux(irtdoc)*(-49.5)-srivflux(irdet)*(-10.5)                &
-             &                   - srivflux(irdin)*1.5 - srivflux(irdip)*2.                        &
-             &                   - (srivflux(iralk)+srivflux(irdin)+srivflux(irdip))
+             &                   - srivflux(irdin)*1.5-srivflux(irdip)*2.                          &
+             &                   - (srivflux(iralk)+srivflux(irdoc)*rcar_tdochc+srivflux(irdin)    &
+             &                   + srivflux(irdip))
 
       else
         totalcarbon = totalcarbon- (srivflux(irdoc)+srivflux(irdet))*rcar                          &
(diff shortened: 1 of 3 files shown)


hamocc/mo_inventory_bgc.F90
```diff
@@ -481,10 +481,10 @@ subroutine inventory_bgc(kpie,kpje,kpke,dlxp,dlyp,ddpo,omask,iogrp)
...
                                                    ! changes due to instantaneous remineralisation 
                                                    ! of riverine DOC are ignored
-        totalnitr   = totalnitr  - srivflux(irdoc)*rnit-srivflux(irtdoc)*rnit_tdochc               &
-             &                   - srivflux(irdet)*rnit_tdoclc - srivflux(irdin)
-        totalphos   = totalphos  - srivflux(irdoc)-srivflux(irtdoc)-srivflux(irdet)-srivflux(irdip)
-        totaloxy    = totaloxy   - srivflux(irdoc)*(-24.) - srivflux(irtdoc)*(-49.5)               &
+        totalnitr   = totalnitr  - srivflux(irtdoc)*rnit_tdochc-srivflux(irdet)*rnit_tdoclc        &
+             &                   - srivflux(irdin)
+        totalphos   = totalphos  - srivflux(irtdoc)-srivflux(irdet)-srivflux(irdip)
+        totaloxy    = totaloxy   - (srivflux(irdoc)+srivflux(irtdoc))*(-49.5)                      &
```
jmaerz: Isn't the oxygen for the remineralization for DOC beforehand consumed? - I might still not fully get the protocol, but since the irdoc is beforehand put into the DIN and DIP, why is the oxygen contained here considered?

jmaerz:
Two requests/questions:
- [x] Please provide a comment on iron in `mo_read_rivin` / `mo_apply_rivin`
- [ ] I would suggest to run the model in DEBUG mode with: `use_PBGC_OCNP_TIMESTEP` and  `use_PBGC_CK_TIMESTEP` set to true - and then check the inventory - this should allow to check, if all your calculations are correct. I hope that the default model still does preserve mass within a time step (should be checked from time to time... - not sure, when was the last time, though).

hamocc/mo_inventory_bgc.F90
TimotheeBrgs: Good point, this is fixed.

TimotheeBrgs:
Thanks @jmaerz for your suggestions. By "_run the model in DEBUG mode with: use_PBGC_OCNP_TIMESTEP and use_PBGC_CK_TIMESTEP set to true_", do you mean adding the code below to `user_nl_blom`? Or only the last 2 lines? Thanks.
```
use_DEBUG_ALL = .true.
use_PBGC_OCNP_TIMESTEP = .true.
use_PBGC_CK_TIMESTEP = .true.
```

jmaerz:
Hi @TimotheeBrgs , you'll need to perform a `./xmlchange DEBUG=TRUE` in the case folder and the two mentioned flags should enter the `user_nl_blom` as you proposed. The `./xmlchange DEBUG=TRUE` slows down the model quite a bit - so just run it for real debugging purposes and a short time - and not completely necessary in your case for checking the inventory, but maybe worth a go nevertheless for re-checking the current code (it's a more general check - not related to the inventory - more for code issues). The two flags in the `user_nl_blom` switch on the inventory being calculated at each individual time step - you'll find the output in the log file of BLOM (DEBUG notes should appear in the cesm-log).

`use_PBGC_OCNP_TIMESTEP` checks the inventory between individual processes in `mo_ocprod` - giving a hint, if something goes wrong in those processes of ocean productivity.

`use_PBGC_CK_TIMESTEP` checks the whole `mo_hamocc4bgc` steps - so more than just the ocean productivity routine. 

Just have a look throughout the code, where these two flags appear so that you know, what's happening.

In general, you shouldn't see changes in the inventory in one time step (potentially some minor changes in the numerically expectable 3 last digits) for the total inventories.  Maybe worth running both model setups (for a model month or so) - the default model and your river2omip protocol to be sure about where potential errors come from (and to check the current default model). It would be great, if you could report back.

**For later reference:** ideally switch off the sediment acceleration scheme, when checking the inventory in this way, since the inventory calculations don't account for sediment changes induced by the sediment acceleration scheme (set `do_sedspinup = false` - note: `do_sedspinup` is currently automatically set via the `namelist_definitions_blom.xml` -or actually `config_component.xml`- to `true` for some years, if one is not switching it off explicitly).

Check for:
- `cat my_blom_logfile | grep "of phosph"`
- `cat my_blom_logfile | grep "of nitrogen"`
- `cat my_blom_logfile | grep "of silicate"`
- `cat my_blom_logfile | grep "of carbon"`

where values shouldn't differ within individual time steps, while this might get obscured by the limit function in `mo_hamocc4bcm`.

hamocc/mo_inventory_bgc.F90
```diff
@@ -481,11 +481,10 @@ subroutine inventory_bgc(kpie,kpje,kpke,dlxp,dlyp,ddpo,omask,iogrp)
...
-        totalnitr   = totalnitr  - srivflux(irdoc)*rnit-srivflux(irtdoc)*rnit_tdochc               &
-             &                   - srivflux(irdet)*rnit_tdoclc - srivflux(irdin)
-        totalphos   = totalphos  - srivflux(irdoc)-srivflux(irtdoc)-srivflux(irdet)-srivflux(irdip)
-        totaloxy    = totaloxy   - srivflux(irdoc)*(-24.) - srivflux(irtdoc)*(-49.5)               &
-             &                   - srivflux(irdet)*(-10.5)                                         &
+        totalnitr   = totalnitr  - srivflux(irtdoc)*rnit_tdochc-srivflux(irdet)*rnit_tdoclc        &
+             &                   - srivflux(irdin)
+        totalphos   = totalphos  - srivflux(irtdoc)-srivflux(irdet)-srivflux(irdip)
+        totaloxy    = totaloxy   - srivflux(irtdoc)*(-49.5)-srivflux(irdet)*(-10.5)                &
              &                   - srivflux(irdin)*1.5 - srivflux(irdip)*2.                        &
```
TimotheeBrgs: @jmaerz I'm interesting in getting your advice on the factors to use for the oxygen inventory.

The facts:
- `-49.5` refers to a CNP ratio of 2583:103:1 (used for riverine labile DOC and terrestrial semi-labile DOC)
- `-10.5` refers to a CNP ratio of 276:25:1 (used for riverine DET).
- Riverine DIN and DIP includes riverine DIN, riverine DIP, and an unknown part representing the DIN and DIP released from direct remineralisation of riverine labile DOC.
- We agreed to not account in the oxygen inventory for the O2 consumption of riverine labile DOC remineralisation due to the fact that this process is considered happening beforehand.

My issue:
The factors `1.5` and `2` used for riverine DIN and DIP are based on the model Redfield ratio 122:16:1. An unknown part of DIN and DIP are based on the ratio 2583:103:1 from the riverine labile DOC.

I am not sure how to handle this.

TimotheeBrgs: @jmaerz Despite the fact that the code below does not lead to double counting (rDIN and rDIP are corrected), I feel like it does not make sense because no oxygen was consumed for rDOC remineralisation.

```
totaloxy    = totaloxy   - (srivflux(irdoc)+srivflux(irtdoc))*(-49.5)                      &
     &                   - srivflux(irdet)*(-10.5)                                         &
     &                   - (srivflux(irdin)-srivflux(irdoc)*rnit_tdochc)*1.5               &
     &                   - (srivflux(irdip)-srivflux(irdoc))*2.                            &
     &                   - (srivflux(iralk)+srivflux(irdin)-srivflux(irdoc)*rnit_tdochc    &
     &                   + srivflux(irdip)-srivflux(irdoc))
```

TimotheeBrgs: @jmaerz This one makes more sense. I need to treat rALK the same way I treat rDOC.
```
totaloxy    = totaloxy   - srivflux(irtdoc)*(-49.5)-srivflux(irdet)*(-10.5)                &
     &                   - srivflux(irdin)*1.5-srivflux(irdip)*2.                          &
     &                   - (srivflux(iralk)+srivflux(irdoc)*rcar_tdochc+srivflux(irdin)    &
     &                   + srivflux(irdip))
```

TimotheeBrgs: The code above fixes the oxygen inventory when river2omip is activated 👍

jmaerz: Sounds good - thanks for re-checking and fixing! Do I understand correct that now C,N,Si,P inventories are consistent within time steps and O2 inventory calculations has the known issues, while the river2omip doesn't induce further ones?

JorgSchwinger (approved):

jmaerz:
Hi @TimotheeBrgs , I see that you merged this now. Correct me, if I am wrong, but we were discussing that there were still changes in the inventory due to the new riverine input. Was this now fixed or are there still issues left? Would be good to clarify this...

TimotheeBrgs:
Hi @jmaerz , there is still indeed a "minor" issue regarding the calculation of the oxygen inventory which was not closely related with the initial issue of this PR. I wanted to merge this to prevent blocking future PR on the hamocc code. I can reopen issue #588  to keep this alive or create another issue with a copy-paste of the protocol you wrote for inventory checking. Please let me know what is more appropriate.

jmaerz:
Hi @TimotheeBrgs , since not everything was addressed in #588 what you changed here, I would suggest to open a new issue, where you briefly describe the remaining issue related to river2omip and its effect on the oxygen inventory (leaving aside the other discussion we have internally). It helps to keep track of remaining issues with the code and helps others to potentially look into it. For convenience, I suggest to also link to the old issue and this merged PR in the new issues description so that it becomes easier to follow up (but this would be my personal taste - I think, we haven't noted such things down anywhere).   Wrt the DEBUG procedure: you can easily link to individual comments as well (so no need to copy the little recipe).

PR #590, opened 2025-05-28
Change name convention for inventory output file

### LINKED ISSUE #586: Naming convention for inventory output
@JorgSchwinger , @jmaerz 

We found that the current naming convention for inventory output `*.hbgcy.i.*` was interfering with the implementation of timeseries diagnostics. A fix has been proposed for HAMOCC_DIAG that circumvents this problem by searching for integer values after the `.hbgc?.` prefix.

As a precautionary measure, to avoid similar problems in the future, I suggest to change the naming convention for the inventory files. Options include

1. `.hbgcyi.`; `.hbgcmi.`; `.hbgcdi.`
2. `.hbgciy.`; `.hbgcim.`; `.hbgcid.`

I have a slight preference for option 2, as it avoids "false" hits if someone searches for e.g. `hbgcy`. Please suggest if you have other options/preferences.

TomasTorsvik:
In this PR I assume we only want to write netcdf output at a single time interval, and prioritize the highest frequency, i.e. 'daily' > 'monthly' > 'yearly', and therefore that we only have one set of output files.

I have not tried to implement writing netcdf output at a time step frequency at this stage.

Closes #586

### DIFF
--- a/hamocc/mo_bgcmean.F90
+++ b/hamocc/mo_bgcmean.F90
@@ -809,6 +809,9 @@ CONTAINS
       endif
     enddo
 
+    ! Only allow inventory output to netcdf file at one interval
+    call check_glb_inventory(glb_inventory,nbgc)
+
     ! Re-define index variables according to namelist
     i_bsc_m2d=0
     do n=1,nbgc
@@ -2731,5 +2734,35 @@ CONTAINS
     !$OMP END PARALLEL DO
     !
   end subroutine bgczlv
+
+  subroutine check_glb_inventory(glb_inventory,nbgc)
+    !
+    ! --- ------------------------------------------------------------------
+    ! --- Description: check that we only write invetory to netcdf file at
+    ! ---              a single time interval, set by glb_inventory(n) == 2.
+    ! ---              If there are multiple instances where glb_inventory
+    ! ---              is 2, keep only the first instance.
+    ! --- ------------------------------------------------------------------
+
+    ! Arguments
+    integer, intent(inout) :: glb_inventory(nbgcmax)
+    integer, intent(in)    :: nbgc
+    !
+    ! Local variables
+    integer :: i,maxval_glb_inventory, maxloc_glb_inventory
+
+    maxval_glb_inventory = maxval(glb_inventory)
+    if (max_glb_inventory == 2) then
+       maxloc_glb_inventory = maxloc(glb_inventory)
+       if (maxloc_glb_inventory < nbgc) then
+          ! Set subsequent instances of glb_inventory to max 1
+          do i=maxloc_glb_inventory+1,nbgc
+             if (glb_inventory(i) == 2) then
+                glb_inventory(i) = 1
+             endif
+          enddo
+       endif
+    endif
+  end subroutine check_glb_inventory
 
 end module mo_bgcmean
--- a/hamocc/mo_bgcmean.F90
+++ b/hamocc/mo_bgcmean.F90
@@ -2738,7 +2738,7 @@ CONTAINS
   subroutine check_glb_inventory(glb_inventory,nbgc)
     !
     ! --- ------------------------------------------------------------------
-    ! --- Description: check that we only write invetory to netcdf file at
+    ! --- Description: check that we only write inventory to netcdf file at
     ! ---              a single time interval, set by glb_inventory(n) == 2.
     ! ---              If there are multiple instances where glb_inventory
     ! ---              is 2, keep only the first instance.
--- a/hamocc/mo_bgcmean.F90
+++ b/hamocc/mo_bgcmean.F90
@@ -2752,8 +2752,8 @@ CONTAINS
     integer :: i,maxval_glb_inventory, maxloc_glb_inventory
 
     maxval_glb_inventory = maxval(glb_inventory)
-    if (max_glb_inventory == 2) then
-       maxloc_glb_inventory = maxloc(glb_inventory)
+    if (maxval_glb_inventory == 2) then
+       maxloc_glb_inventory = maxloc(glb_inventory, 1)
        if (maxloc_glb_inventory < nbgc) then
           ! Set subsequent instances of glb_inventory to max 1
           do i=maxloc_glb_inventory+1,nbgc
--- a/hamocc/mo_inventory_bgc.F90
+++ b/hamocc/mo_inventory_bgc.F90
@@ -743,8 +743,8 @@ contains
                                nf90_put_att, nf90_put_var, nf90_unlimited, nf90_write
       use mod_types,     only: r8
       use mod_config,    only: expcnf, runid, inst_suffix
-      use mod_time,      only: date0, time0, date, time, nstep, nday_of_year,nstep_in_day,calendar &
-                               blom_time
+      use mod_time,      only: date0, time0, date, time, nstep, nday_of_year, nstep_in_day,        &
+                               calendar, blom_time
       use mo_bgcmean,    only: filefq_bgc, fileann_bgc, filemon_bgc,glb_fnametag
       use mo_param1_bgc, only: idicsat,idms,ifdust,iiron,iprefalk,iprefdic,iprefo2,iprefpo4,       &
                                iadust,inos,ibromo,icfc11,icfc12,isf6,icalc13,icalc14,idet13,       &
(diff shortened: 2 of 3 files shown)


hamocc/mo_inventory_bgc.F90
```diff
@@ -901,8 +901,7 @@ subroutine write_netcdf(iogrp)
         endif
         write(tstamp,'(i4.4,a1,i2.2,a1,i2.2)')                                    &
              &    date%year,sep2,date%month,sep2,date%day
-        fname_inv(iogrp) = prefix//sep1//trim(glb_fnametag(iogrp))//sep1//        &
-             &    'i'//sep1//trim(tstamp)//'.nc'
+        fname_inv(iogrp) = prefix//sep1//'hbgci'//sep1//trim(tstamp)//'.nc'
```
jmaerz, line 907: Should likely read:
```
call blom_time(ymd,tod)
write(tstamp,'(i4.4,a1,i2.2,a1,i2.2,a1,i5.5)') date%year,sep2,date%month,sep2,date%day,sep2,tod
fname_inv(iogrp) = prefix//sep1//'bgci'//sep1//trim(tstamp)//'.nc'
```
(requires `integer :: ymd,tod` and `use mod_time, only: blom_time`

hamocc/mo_bgcmean.F90
```diff
@@ -2732,4 +2735,34 @@ subroutine bgczlv(pddpo,kin,ind1,ind2,weights)
     !
   end subroutine bgczlv
 
+  subroutine check_glb_inventory(glb_inventory,nbgc)
+    !
+    ! --- ------------------------------------------------------------------
+    ! --- Description: check that we only write invetory to netcdf file at
```
jmaerz: inventory (typo)

hamocc/mo_bgcmean.F90
```diff
@@ -2732,4 +2735,34 @@ subroutine bgczlv(pddpo,kin,ind1,ind2,weights)
...
+       maxloc_glb_inventory = maxloc(glb_inventory)
+       if (maxloc_glb_inventory < nbgc) then
+          ! Set subsequent instances of glb_inventory to max 1
+          do i=maxloc_glb_inventory+1,nbgc
+             if (glb_inventory(i) == 2) then
+                glb_inventory(i) = 1
+             endif
+          enddo
+       endif
+    endif
```
jmaerz, line 2765: When looking at the year length and default month length, I started to ponder, when things are actually get written out... - i.e 12*30=360 (!=365).... - so will the monthly also cover the year (end) output? - I suspect so, so the question is more on the 360 !=365...

TomasTorsvik: That is a good question, I haven't considered this. This is getting more complicated than I anticipated ...

jmaerz: Admittedly, I would have just changed `mo_inventory` (as you did now). But maybe it's a better way, how you implemented it - I am not fully getting, how this iopgrp stuff works behind the scenes... (I never fully looked into it - and I guess, once I would have understand, I would try to change the whole thing to make model developers life easier with respect to output ... - back to the old rumbling on output implementation, sorry).

TomasTorsvik: @jmaerz - I think you are right, the changes in `mo_bgcmean` are not needed, especially with the extended time stamp. With the NF90_CLOBBER option in the file cration command, we will overwrite any existing file. Alternatively, we could set this to NF90_NOCLOBBER, which would return a NF90_EEXIST error (and halt the simulation) if there is an existing file. Would this be a better option?

jmaerz: Hi @TomasTorsvik , Ok, I start seeing the consequences that daily output could enter the monthly (and potentially yearly files and so forth)... - admittedly, I have never perceived the files as averaged output - and from my side, I was fine with the previously implemented version. Maybe just adding the time stamp to the previous version (and potentially `hbgc?.i.YYYY` -> `hbgc?i.YYYY`) would suit that distinguishing - while it then counteracts your `ls`/`grep` preferences? I will be happy with whatever comes up (as long as we keep the inventory output). Admittedly, I understand the wish (and need) for the time stamp, while not really for changes with respect to the diagnostics (these will be changed anyway soon and I would assume that the recent fix that we introduced, should work). But maybe @JorgSchwinger has a dedicated opinion on it?

jmaerz: @TomasTorsvik , I just tried it out in single column mode. Maybe you could try to run your branch once for one year with all inventory output on (daily, monthly and yearly) and re-check, if that works? - I think that would help to understand the consequences a bit better (in single column, it seems as if daily output is written the second day at the beginning of the month, while monthly right at the beginning).

jmaerz: Another thing: for daily inventory, the time stamp doesn't make sense, when all is written in a monthly file...

TomasTorsvik: @jmaerz - I will make a test run with inventory output for all time frequencies. Now I went back to the original implementation, just removed the separator `hbgc?.i.YYYY` -> `hbgc?i.YYYY`

jmaerz: Hi @TomasTorsvik , sorry for all this back and forth - I feel that primarily @JorgSchwinger needs to comment - as written, I am (likely) happy, with whatever solution you're coming up (as long as we keep writing the inventory files).

TomasTorsvik: It seems that the output is ignoring 30 day intervals per month, when running fully coupled.
I see output for the last day of the month, but written with a time-stamp for the first day of the following month, i.e. `hbgcmi.0001-02-01 = day 31`, `hbgcmi.0001-02-02 = day 59`, etc.
The yearly output is duplicated by the first monthly out[put of the year, i.e. 
```
hbgcmi.0002-01-01 = hbgcyi.0002-01-01
```
The daily output get a time-stamp for the 2nd day of the month, except the first file which gets time-stamp for the 3rd day of the month.

JorgSchwinger:
Hi just a few comments: Since it is instantaneous values, I do not see the purpose of the qualifiers "d", "m", and "y" in the file names. The only relevant information is the time-stamp that specifies for which time the values in the files are valid. Also, if you write a t higher frequencies, it never makes sense to include files for lower frequencies in the output (i.e. if you write monthly, you will not need annual, if you write daily, you will not need monthly and annual). 

The 30-day frequency is, as far as I am aware, treated as "monthly" so it intentionally always uses whole month.

TomasTorsvik:
> Hi just a few comments: Since it is instantaneous values, I do not see the purpose of the qualifiers "d", "m", and "y" in the file names. The only relevant information is the time-stamp that specifies for which time the values in the files are valid. Also, if you write a t higher frequencies, it never makes sense to include files for lower frequencies in the output (i.e. if you write monthly, you will not need annual, if you write daily, you will not need monthly and annual).
> 
> The 30-day frequency is, as far as I am aware, treated as "monthly" so it intentionally always uses whole month.

@JorgSchwinger , @jmaerz - This is basically the solution I implemented as commit No. 2 (https://github.com/NorESMhub/BLOM/pull/590/commits/aecd4d907736e0b1fbe8986101a29a6e5776cd1f) in this PR, before I rolled back the changes. It assumes that if multiple `glb_inventory` values are set to 2, we only use the frequency associated with the first instance, i.e. a `GLB_INVENTORY` setting of "2,2,2" will be changed to "2,1,1" internally in `mo_bgcmean` before these values are used. 

If this solution is acceptable, I propose to roll back this PR to commit No. 2 and do one more test to make sure the code doesn't break on multiple "2" values in `GLB_INVENTORY`.

jmaerz:
@TomasTorsvik and @JorgSchwinger , I am happy with any solution you come up with (haven't had issues with it before). Will the time stamp for daily output be the time stamp at first day or last day of writing? - while I anticipate that daily output isn't the typical use mode...

TomasTorsvik:
@jmaerz - From the previous test I got the following output for daily inventory:
```
*.hbgcdi.0001-01-03.nc :  day 2-31
*.hbgcdi.0002-01-02.nc :  day 32-59
*.hbgcdi.0003-01-02.nc :  day 60-90
...
```
This was a NOINYOC "startup" run, so day 1 was not included.

I will proceed with rolling back to (https://github.com/NorESMhub/BLOM/commit/aecd4d907736e0b1fbe8986101a29a6e5776cd1f) and run a new test.

jmaerz:
@TomasTorsvik , erh, I fear, I don't understand the time stamp. But yes, fine for me to roll back to your initial solution and give it a try.

TomasTorsvik:
@JorgSchwinger , @jmaerz - The time stamp for daily output is not very sensible in the current setup, but it is consistent with how it has been until now. I updated the archive setting to include the inventory file, so it should be possible to merge this now.

jmaerz (approved):
Hi @TomasTorsvik , thanks for diving into this so deep. For me, it looks reasonable - hope it helps for future simulations and diagnostics tools.

PR #596, opened 2025-06-05
Reference potential density adaptation to evolving model state

matsbn:
This PR introduces adaptability of the reference potential densities to the evolving model state for the hybrid vertical coordinate. This is realised by the following new capabilities:
- The namelist variable `sigref_spec` can now be set to `'function'` so that the reference potential densities are defined by a functional form, specified by the namelist variable `sigref_fun_spec`. The `sigref_fun_spec` is a derived data type with parameters to define a cubic Bezier curve, suitable to provide a reference potential density distribution for the bulk of the water masses. For very light and dense water masses, the Bezier curve is complimented with parabolic distributions, also specified in `sigref_fun_spec`.
- If the new namelist variable `sigref_adaption` is set to true, the parameters of `sigref_fun_spec` are adjusted to the evolving model state. The adjustment seeks to keep the mass fraction of the densest parabolic distribution to a prescribed fraction and to adjust some of the Bezier points to optimally occupy the mixed layer with constant pressure levels.  The e-folding time scale of the parameter adjustment is set by the namelist variable `sra_ts`, currently defaulting to 5 years.

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -978,6 +978,136 @@
     <desc>Pressure level specification: Valid specs.: 'inflation', 'namelist'</desc>
   </entry>
 
+  <entry id="sigref_fun_spec%dsdz_bot">
+    <type>real</type>
+    <category>vcoord</category>
+    <group>vcoord</group>
+    <values>
+      <value>1.125</value>
+    </values>
+    <desc>Derivative of sigma with respect to z at Bezier point 4 (z = 1) [kg m-3]</desc>
+  </entry>
+
+  <entry id="sigref_fun_spec%sp1">
+    <type>real</type>
+    <category>vcoord</category>
+    <group>vcoord</group>
+    <values>
+      <value>23.70</value>
+    </values>
+    <desc>Sigma value at Bezier point 1 (z = 0) [kg m-3]</desc>
+  </entry>
+
+  <entry id="sigref_fun_spec%zp2">
+    <type>real</type>
+    <category>vcoord</category>
+    <group>vcoord</group>
+    <values>
+      <value>0.21</value>
+    </values>
+    <desc>z value at Bezier point 2 []</desc>
+  </entry>
+
+  <entry id="sigref_fun_spec%zp3">
+    <type>real</type>
+    <category>vcoord</category>
+    <group>vcoord</group>
+    <values>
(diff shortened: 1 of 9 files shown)


TomasTorsvik (approved):
@matsbn - superficially, this looks fine to me. Could you please comment
- what kind of testing has ben carried out with these code changes?
- is this intended to be turned on as default in NorESM3.0?
- if so, should we expect to see significant changes in the physics (and where would this be)?

matsbn:
> @matsbn - superficially, this looks fine to me. Could you please comment
> 
> * what kind of testing has ben carried out with these code changes?
> * is this intended to be turned on as default in NorESM3.0?
> * if so, should we expect to see significant changes in the physics (and where would this be)?

@TomasTorsvik,

In terms of testing I have confirmed that results are identical with different processors counts and that it reproduce results after restart in OMIP2 configuration (2 + 1 years identical to 3 years).

If longer simulations give satisfactory results, I would want this as default in NorESM3.0. It is a solution to the inability of NorESM2 to run e.g. the 4XCO2 experiment towards stabilisation (model crashes since available density classes are unsuitable). It might also be beneficial for scenario simulations or other simulations where the model state evolves significantly.

I do not expect very large changes in the physics, but cannot say for sure before analysing long simulations. A full OMIP2 cycle is underway that should give some insight. Also coupled simulations will be started shortly. The biggest challenge found so far is mixing in the Mediterranean outflow, which is very sensitive to model vertical resolution.

TomasTorsvik:
@matsbn - thanks for clarification. I don't mind if you merge this when you think it is ready, but perhaps it would be useful for documentation to include a link to the case description and/or diagnostics for the long OMIP2 in this PR before doing so.

matsbn:
A OMIP2 forcing cycle has been completed as documented in https://github.com/NorESMhub/noresm3_dev_simulations/issues/142.

The diagnostics contains a comparison to NOIIAJRAOC_TL319_tn14_3.0a3_newmix_20250430, which should be consistent with the BLOM tag in noresm3_0_alpha03b. In addition to the reference potential density adaption, the new simulation starts from WOA climatology and uses 72 model layers.

Rather small differences are found. The new simulation has reduced warm bias in low latitude upper ocean, except in Indian Ocean. Somewhat insufficient mixing is found in the Mediterranean outflow, but maybe acceptable.

PR #597, opened 2025-06-05
fix ODZ volume calculation

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , this is a bugfix for the ODZ volume calculation.


JorgSchwinger (approved):
Thanks for fixing this

TomasTorsvik (approved):
@jmaerz - good catch, thanks!

Issue #594, opened 2025-06-04
Dust/iron deposition increases with time when using stream-functionality


JorgSchwinger:
When using the new dust/iron stream functionality, deposition seems to increase over time (blue: old MCT based dust/iron, red/yellow: new stream based dust/iron)

![Image](https://github.com/user-attachments/assets/a8846ac0-c05f-4e69-8650-1af2294da269)

Below is the spatial pattern of the increase (the difference between year 100 and year 1 of the simulation), which follows the spatial pattern of dust deposition itself, so it seems to be some kind of factor/percentage.

![Image](https://github.com/user-attachments/assets/61ad0a13-588d-44cb-919e-46385a4aeed8)

@mvertens could you check if you can spot any possible reason for this?

mvertens:
@JorgSchwinger - what compset is this? Is this a pre-industrial compset? Can you point me to your CASEROOT.
Also - are you using mahw2006 or GESAMP2018? I'm really perplexed with this result.

JorgSchwinger:
@mvertens This is the NOINYOC compset (on the tnx2v1 grid). My caseroot is here:

`/cluster/projects/nn2345k/schwinger/cases/NOINYOC_T62_tn21_60`

JorgSchwinger:
It's the old mahw2006 option.

mvertens:
@JorgSchwinger - Thanks. I have a few other questions - when would you be free to chat for 5 minutes?

mvertens:
@JorgSchwinger - I have looked at this really carefully and I just don't understand how this can be occurring. So a quick chat would be super helpful.

JorgSchwinger:
@mvertens I am on a conference, but I possibly could have a chat Friday morning (or now, if you are still working)

JorgSchwinger:
Has been fixed in #598 (thanks to @mvertens !). I close this issue now.

PR #598, opened 2025-06-06
fixed dust issue where it was growing each year

mvertens:
These changes fixes the problem found by @JorgSchwinger in  issue #594. 
The key problem was the settings for stream_dust_year_first and stream_dust_year_align in namelist_definition_blom.xml.
Testing: I have verified that running his case for 2 years gives the same field for tfluxdst in month 3 of year 2 versus month 3 of year 1.

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -10081,7 +10081,7 @@
       First year in stream to use
     </desc>
     <values>
-      <value>1</value>
+      <value>1850</value>
     </values>
   </entry>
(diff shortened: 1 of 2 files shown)


TomasTorsvik (approved):
@mvertens - thanks for fixing this issue!

mvertens:
@TomasTorsvik - it would be good to get this into noresm3_0_alpha05 - since using the stream capability for dust is the default.

TomasTorsvik:
@mvertens - I have merged this, so it should be ready for `noresm3_0_alpha05`.

PR #599, opened 2025-06-07
Fixes an issue with truncated model depth when using WOA initial conditions.

matsbn:
This PR fixes a bug that caused the model depth to be truncated when the model depth was deeper than the World Ocean Atlas depth range (5500 m).

### DIFF
--- a/phy/mod_inicon.F90
+++ b/phy/mod_inicon.F90
@@ -467,7 +467,6 @@ contains
         ! Create source arrays of practical salinity (sp_src), potential
         ! temperature (pt_src) and source interface depths bounded by model
         ! depth (z_src). Fill remaining missing values from above.
-        z_src(1) = z_src_ref(1)
         do k = 1, kdm_woa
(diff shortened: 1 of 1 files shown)


phy/mod_inicon.F90
```diff
@@ -481,13 +480,16 @@ subroutine inicon_woa_file
...
         enddo
+        z_src(kdm_woa+1) = - depths(i,j)
 
         ! Bound destination interface depths by model depth.
-        do k = 1, kk+1
+        do k = 1, kk
           z_dst(k) = max(z_dst_ref(k), - depths(i,j))
         enddo
+         z_dst(kk+1) = - depths(i,j)
+
```
TomasTorsvik:
```suggestion
        enddo
        z_dst(kk+1) = - depths(i,j)
```
whitespace indentation

TomasTorsvik (commented):
@matsbn - thanks for fixing!

TomasTorsvik (approved):

PR #603, opened 2025-06-10
Eddy diffusivity modifications

matsbn:
This PR adds the option to smooth eddy diffusivity (isopycnal and interface). It also modifies the estimation of diffusivity with the ALE method to avoid large diffusivity values near the surface.

The current default diffusivity with hybrid vertical coordinate is typically like this:
NOIIAJRAOC_TL319_tn14_3 0a3_newmix_20250430
Quite noisy and overall rather small diffusivity values. By enabling lateral smoothing we get this:
NOIIAJRA_TL319_tn14_3 0a3b_eddff_20250609
Better, but still rather small diffusivities. In discussions with Mehmet and Aleksi it was found that the effect of incorporating topographic beta has a strong effect on the diffusivity strenght and noisiness. By not using topographic beta we now get this:
NOIIAJRA_TL319_tn14_3 0a3b_eddff_rhsctp0_20250609
Including topographic beta effect, but changing RHISCF from 0.5 to 5.0 we get this diffusivity:
NOIIAJRA_TL319_tn14_3 0a3b_eddff_rhiscf5_20250609
In this PR the default is to apply lateral smoothing. Other settings, like topographic beta effect is still unchanged, but based on the figures above, they might be adjusted later.


TomasTorsvik (approved):
@matsbn - Thanks, looks interesting. I agree that "lateral smoothing" as a first default should be rather conservative relative to earlier noresm3 test simulations, so probably good to start there.

milicak (approved):
Maybe we should create an issue about this to investigate later. The lateral diffusivity is important for coarse resolution models.

matsbn:
I have created issue https://github.com/NorESMhub/BLOM/issues/604 for further discussions on parameter settings.

PR #606, opened 2025-06-11
Include parenthesis to ensure correct evaluation of if statement

TomasTorsvik:


hamocc/mo_param_bgc.F90
```diff
@@ -584,7 +584,7 @@ subroutine ini_bgctimes(nday_in_year)
 
     days_per_year = real(nday_in_year)
 
-    if (nday_in_year /= 365 .and. mnproc==1 .and. lini .eqv. .true.) then
+    if ((nday_in_year /= 365) .and. (mnproc == 1) .and. (lini .eqv. .true.)) then
       lini=.false.
       if (.not. (expcnf == 'single_column' .or. expcnf == 'fuk95' .or. expcnf == 'channel' .or. expcnf == 'noforcing')) then
```
jmaerz, line 589: Hi @TomasTorsvik , I wonder about this line as well...

jmaerz (approved):

hamocc/mo_param_bgc.F90
JorgSchwinger: According to my fortran book, the operators `==`, `/=` etc. have a HIGHER priority than `.and.` and `or` so these changes should not be necessary.

This is in contrast to `.eqv` which has a LOWER priority than `.and.` and `or` (somewhat counterintuitively...)

TomasTorsvik: Yes, in principle it is only the `(lini .eqv. .true. )` that needs the parenthesis, but for consistency I included for the two first statements as well.

PR #609, opened 2025-06-11
HAMOCC working precision for real values

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , this is the PR to introduce explicit real values working precision in iHAMOCC. 

@TomasTorsvik , could you perform a regression testing based on the branch? - I won't merge without the successful testing. 

Further: I wasn't fishing (much) for real values that haven't had a dot (`.`). So, I expect that there are a few real numbers, written as integer values, left, which I might attempt to fix in a second step.


hamocc/mo_bgcmean.F90
```diff
@@ -2206,10 +2207,10 @@ subroutine wrtsrf(pos,frmt,sfac,offs,cmpflg,vnm)
     integer, intent(in) :: frmt         ! format/precision of output
                                         !  0=field is not written
                                         !  2=field is written as int2 with scale factor and offset
-                                        !  4=field is written as real4
-                                        !  8=field is written as real8
-    real,    intent(in) :: sfac         ! user defined scale factor to be applied
-    real,    intent(in) :: offs         ! user defined offset to be added
+                                        !  4=field is written as real(rp)4
+                                        !  8=field is written as real(rp)8
```
JorgSchwinger: Here the (rp) doesn't make sense...

Same couple of times below.

hamocc/mo_bgcmean.F90
```diff
@@ -2650,7 +2651,7 @@ subroutine msksrf(pos,idepth)
     !
     ! Local variables
     integer :: i,j,l
-    real, parameter :: mskval=nf90_fill_double
+    real(rp), parameter :: mskval=nf90_fill_double
```
JorgSchwinger, line 2654: Here is a bit the question how to handle this. 

If `rp==4` one would need to select also ` nf90_fill_float` 

Maybe issue and error if rp is neither 4 nor 8?

hamocc/mo_bgcmean.F90
```diff
@@ -2676,13 +2677,13 @@ subroutine msklvl(pos,depths)
...
     integer, intent(in)  :: pos                                     ! field position in level buffer
-    real,    intent(in)  :: depths(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) ! bathymetry field
+    real(rp),intent(in)  :: depths(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) ! bathymetry field
     !
     ! Local variables
     integer         :: i,j,k,l
     logical, save   :: iniflg=.true.
     integer, save   :: kmax(idm,jdm)
-    real, parameter :: mskval=nf90_fill_double
+    real(rp), parameter :: mskval=nf90_fill_double
```
JorgSchwinger, line 2686: Same as above

hamocc/mo_get_cfc.F90
```diff
@@ -28,136 +28,137 @@ subroutine get_cfc(kplyear,atm_cfc11_nh,atm_cfc12_nh,atm_sf6_nh,
                      atm_cfc11_sh,atm_cfc12_sh,atm_sf6_sh)
 
     !***********************************************************************************************
-    !     Jerry Tjiputra         *BCCR*          05.12.2012
+    !     Jerry Tjiputra         *BCCR*          05.12_rp.2012
```
JorgSchwinger: Mhm, doesn't make sense here...

hamocc/mo_netcdf_bgcrw.F90
```diff
@@ -318,8 +320,8 @@ subroutine write_netcdf_var(ncid,desc,arr,klev,time)
       allocate(arr_l(ii,jj,klev))
       allocate(arr_g1(itdm,jj,klev))
 
-      arr_l=0.0
-      arr_g1=0.0
+      arr_l=0.0_rp
+      arr_g1=0.0_rp
```
JorgSchwinger, line 324: Further down, the routine `nfmpi_put_vara_double_all` is used, so double precision is hard-coded (if PNETCDF is used)

hamocc/mo_bgcmean.F90
jmaerz: Done.

hamocc/mo_get_cfc.F90
jmaerz: Done.

hamocc/mo_netcdf_bgcrw.F90
```diff
@@ -39,6 +39,7 @@ subroutine netcdf_def_vardb (kcid,kshort,yshort,kdims,kcdims,kcvarid,
     use netcdf,  only: nf90_double,nf90_noerr,nf90_put_att,nf90_def_var
     use mod_xc,  only: mnproc,xchalt
     use mod_dia, only: iotype
+    use mo_kind, only: rp
```
JorgSchwinger, line 42: In this routine/module the usage of DOUBLE is hard-coded a lot of places. Not sure how to handle this

hamocc/mo_read_fedep.F90
```diff
@@ -42,8 +42,8 @@ module mo_read_fedep
   character(len=64),         public :: fedep_source=''
 
   ! Array to store iron deposition fluxes after reading from file
-  real, allocatable,  private :: dustflx_tot(:,:,:)
-  real, allocatable,  private :: dustflx_sfe(:,:,:)
+  real(rp), allocatable,  private :: dustflx_tot(:,:,:)
+  real(rp), allocatable,  private :: dustflx_sfe(:,:,:)
```
JorgSchwinger, line 46: If I'm not mistaken, the routine `read_netcdf_var` expects a double array

hamocc/mo_read_ndep.F90
```diff
@@ -59,8 +59,8 @@ module mo_read_ndep
   public :: ndepfile
 
   character(len=bgc_fnmlen)  :: ndepfile=''
-  real,  allocatable  :: noydepread(:,:)
-  real,  allocatable  :: nhxdepread(:,:)
+  real(rp),  allocatable     :: noydepread(:,:)
+  real(rp),  allocatable     :: nhxdepread(:,:)
```
JorgSchwinger, line 63: Again, the reading routine expects double.

hamocc/mo_read_oafx.F90
```diff
@@ -71,10 +71,10 @@ module mo_read_oafx
 
   character(len=128), protected, public  :: oalkscen   =''
   character(len=bgc_fnmlen), protected, public  :: oalkfile   =''
-  real,allocatable,   protected          :: oalkflx(:,:)
+  real(rp),allocatable,   protected          :: oalkflx(:,:)
```
JorgSchwinger: The reading routine expects double

hamocc/mo_read_rivin.F90
```diff
@@ -64,17 +64,17 @@ module mo_read_rivin
...
-  real, dimension(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) :: riv_idoc2d
-  real, dimension(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) :: riv_itdoc2d
+  real(rp), dimension(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) :: riv_DIN2d
+  real(rp), dimension(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) :: riv_DIP2d
+  real(rp), dimension(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) :: riv_DSI2d
+  real(rp), dimension(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) :: riv_DIC2d
+  real(rp), dimension(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) :: riv_DFe2d
+  real(rp), dimension(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) :: riv_idet2d
+  real(rp), dimension(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) :: riv_idoc2d
+  real(rp), dimension(1-nbdy:idm+nbdy,1-nbdy:jdm+nbdy) :: riv_itdoc2d
```
JorgSchwinger, line 77: This module uses a different routine to read the data from file, I have not checked but it could well be that a certain data type is assumed as well.

hamocc/mo_read_sedpor.F90
```diff
@@ -53,12 +54,12 @@ subroutine read_sedpor(kpie,kpje,ks,omask,sed_por)
...
     integer, intent(in)    :: ks
-    real,    intent(in)    :: omask(kpie,kpje)
-    real,    intent(inout) :: sed_por(kpie,kpje,ks)
+    real(rp),intent(in)    :: omask(kpie,kpje)
+    real(rp),intent(inout) :: sed_por(kpie,kpje,ks)
 
     !local variables
     integer :: i,j,k
-    real    :: sed_por_in(kpie,kpje,ks)
+    real(rp):: sed_por_in(kpie,kpje,ks)
```
JorgSchwinger, line 62: Reading routine assumes double

hamocc/mo_read_shelfmask.F90
```diff
@@ -56,13 +57,13 @@ subroutine ini_read_shelfmask(kpie,kpje,kbnd,pbath,omask)
...
-    real,    intent(in) :: pbath(1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd) ! bathymetry fields - water depth [m]
-    real,    intent(in) :: omask(kpie,kpje)                         ! land/ocean mask.
+    real(rp),intent(in) :: pbath(1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd) ! bathymetry fields - water depth [m]
+    real(rp),intent(in) :: omask(kpie,kpje)                         ! land/ocean mask.
 
     ! Local variables
     logical :: file_exists=.false.
     integer :: i,j,errstat,ncid,ncstat
-    real,allocatable  :: mask(:,:)
+    real(rp),allocatable  :: mask(:,:)
```
JorgSchwinger, line 66: reading routine assumes double

jmaerz:
Hi @JorgSchwinger , maybe it's more useful to tell what's the consequence of the current procedure (as in what happens internally currently without the explicit working precision) to be able to fix this...

JorgSchwinger (commented):
@jmaerz thanks a lot for bringing this in. As I have commented, there is a problem with the netcdf reading/writing, where a DOUBLE data type is hard-coded many places (I do not think that I have caught all places...)

I think this might be ok for now, as long as we are not planning to use something else than 8byte reals, but it should be commented in `mo_kind` that a different setting is currently not supported.

In the longer run it would be good to streamline the netcdf I/O (there is also no reason that HAMOCC partly uses different routines than BLOM...)

jmaerz:
Hi @TomasTorsvik and @JorgSchwinger , with last commits, I added a note to `mo_kind` on support of real values (only `r8`) in iHAMOCC. I further added a few more `_rp` on some real numbers. In single column mode, I got bfb with latest `master`. @TomasTorsvik needed to generate a new base line (due to the latest N-cycle and O2-lim changes in remineralization) - many thanks for testing!. I hope that a new regression test will show also bfb, but I will certainly wait with merging until such test is successful (or discussed otherwise).

TomasTorsvik:
After updating, all tests PASS (except those that have been failing earlier). The tests include basic run test (SMS), restart test (ERS) and reading from restart files (ERR).

jmaerz:
@JorgSchwinger and @TomasTorsvik , anything else to consider for this PR for working precision?

JorgSchwinger:
If we b-f-b results (which we have as I understand) I fine with merging this. Thanks again @jmaerz for bringing this in!

TomasTorsvik:
@jmaerz - I have only tested iHAMOCC with default settings. I will run 2 more hammock-specific tests with optional settings.

It could be useful in this case to visit parts of the code that we normally do not activate.

TomasTorsvik:
To clarify, the 2 standard tests are usermods defined in
`BLOM/cime_config/testdefs/testmods_dirs/blom/hamocc/`

**hamocc1:**
```
./xmlchange HAMOCC_CISO=TRUE
use_cfc = .true.
use_wlin = .true.
use_natdic = .true.
```

**hamocc2:**
```
./xmlchange HAMOCC_CISO=TRUE
./xmlchange HAMOCC_SEDBYPASS=TRUE
use_cfc = .true.
use_wlin = .true.
use_natdic = .true.
```

I'm not sure these are the most sensible test settings, we should perhaps update and expand these.

TomasTorsvik:
The `hamocc1` and `hamocc2` tests also pass without errors.

TomasTorsvik (approved):
@jmaerz - This PR is too long to review in detail, but so far the tests have gone through, so it _should_ be reasonably safe to merge.

jmaerz:
Hi @TomasTorsvik thanks for clarification. I will come up with some further tests (filing a PR), currently, I see a test for:
```
use_M4AGO
use_AGG
use_extNcycle (just realizing that it still defaults to false)
use_river2omip
```
Those are the once that immediately come to my mind at the moment - not sure about combinations. I guess, it would be good to check those in addition to the current hamocc tests, before merging - just to be on the safe side.

jmaerz:
@JorgSchwinger , if you've additional tests in mind, let me know and I can include those in the upcoming PR.

JorgSchwinger:
No, I think that's fine. I would even consider to skip the `use_AGG` - this hasn't been turned on since a long time and I wouldn't consider this a supported option currently.

TomasTorsvik:
@jmaerz - The new tests PASS 6/8, with2 expected FAIL, as discussed in #613 .
I think this can be merged.

JorgSchwinger (approved):

PR #617, opened 2025-06-13
detritus remin also in DOM classes case T- and O2-dependent

### LINKED ISSUE #616: iHAMOCC DOM classes
Hi @JorgSchwinger , @TomasTorsvik and @tjiputra ,

I just realized that the DOM classes code doesn't use the T- and O2-dep remineralization in `mo_ocprod`. Shall we change this to be consistent with the rest of the model?

jmaerz:

this PR makes the detritus remineralization also in case of the DOM classes T- and O2-dependent.

Closes #616


JorgSchwinger (approved):
Thank you for catching this - your changes definitely make sense

PR #618, opened 2025-06-15
Adding heat flux due to material enthalpy flux of water exchange with atmosphere

matsbn:
With this PR, enthalpy flux as required for energy consistency with atmospheric water exchange is added to the non-solar heat flux field. Various methods for distributing the enthalpy flux components are available, and the current (hard-coded) method distributes the total enthalpy flux evenly over the sea-ice free ocean domain. This is based on an implementation of Tomas Toniazzo. 

This PR depends on CAM PR https://github.com/NorESMhub/CAM/pull/227/files and CMEPS PR https://github.com/NorESMhub/CMEPS/pull/36 and must be tested in coordination.

### DIFF
--- a/cesm/mod_cesm.F90
+++ b/cesm/mod_cesm.F90
@@ -28,10 +28,11 @@ module mod_cesm
    use mod_constants,  only: pi
    use mod_time,       only: nstep
    use mod_xc
-   use mod_forcing,    only: trxday, srxday, swa, nsf, lip, sop, eva, rnf, rfi, &
-                             fmltfz, sfl, ztx, mty, ustarw, slp, abswnd, &
-                             lamult, lasl, ustokes, vstokes, atmco2, atmbrf, &
-                             flxdms, flxbrf, &
+   use mod_forcing,    only: trxday, srxday, swa, nsf, hmat, &
+                             lip, sop, eva, rnf, rfi, fmltfz, sfl, &
+                             ztx, mty, ustarw, slp, abswnd, &
+                             lamult, lasl, ustokes, vstokes, &
+                             atmco2, atmbrf, flxdms, flxbrf, &
                              atmn2o, atmnh3, atmnhxdep, atmnoydep, &
                              use_stream_relaxation
    use mod_ben02,      only: initai, rdcsic, rdctsf, fnlzai
@@ -66,6 +67,7 @@ module mod_cesm
    real(r8), dimension(1 - nbdy:idm + nbdy,1 - nbdy:jdm + nbdy, 2) :: &
       swa_da, &          ! Solar heat flux [W m-2].
       nsf_da, &          ! Non-solar heat flux [W m-2].
+      hmat_da, &         ! surf.mat.enth.flx
       hmlt_da, &         ! Heat flux due to melting [W m-2].
       lip_da, &          ! Liquid water flux [kg m-2 s-1].
       sop_da, &          ! Solid precipitation [kg m-2 s-1].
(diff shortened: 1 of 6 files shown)


TomasTorsvik (approved):
@matsbn - I suppose this should be fine. I see that methods 2 and 4 take into account sea ice cover. Will there be a separate enthalpy flux to the sea ice as well, or is it assumed that enthalpy is only transferred between atmosphere and ocean?

matsbn:
> @matsbn - I suppose this should be fine. I see that methods 2 and 4 take into account sea ice cover. Will there be a separate enthalpy flux to the sea ice as well, or is it assumed that enthalpy is only transferred between atmosphere and ocean?

To my understanding, enthalpy flux is not passed to any other components. This approach tries to deal with the enthalpy flux over the ocean domain in an integrated sense to reduce what the CAM "energy fixer" otherwise would have to redistribute for total energy consistency. Enthalpy flux over land domain is still folded back into CAM and the "energy fixer". The average enthalpy flux over ocean is on the order 1 W m-2 into the ocean. Due to worries of systematic impacts on sea-ice, we choose for now only to apply the flux over open ocean.

mvertens:
@matsbn -  I think it would be a good idea for BLOM to receive enthalpy fluxes from the mediator using either the new scheme **OR** the old scheme in the mediator (which is being used by MOM6). The current implementation of this new scheme in CMEPS does not provide for straightforward backwards compatibility to use the old scheme. Furthermore, the budget diagnostics CMEPS using the new scheme are not fully implemented and will take some work. 
BLOM can implement use of the older scheme by advertising the fields Foxx_hrain, Foxx_hsnow, Foxx_hevap and Foxx_hcond (as is done in MOM6). In addition Foxx_hrofl, Foxx_hrofi, Foxx_hrofl_glc, Foxx_hrofi_glc are also advertised in MOM6. It's not clear to me how these translate exactly to what is currently being sent by CAM.
Just to add - a new config variable available to CAM, CMEPS and BLOM can be created (say use_new_enthalpy_scheme) to determine if the new or old enthalpy scheme should be used by these components - and this would coordinate the behavior across all three of these components.

mvertens:
@matsbn - I have another question: 
BLOM advertises that it wants to receive Faxa_hlat
   `call fldlist_add(fldsToOcn_num, fldsToOcn, 'Faxa_hlat'    , index_Faxa_hlat)`
but then never uses it.
In reviewing the CMEPS code for the new implementation - I see the following:

```
logical, parameter :: separate_varlat=.false.

! hack: replace full mat.enth.flux with variable lat.heats part only in pointer to ocean export
! might add another coupling field later but may not be strictly necessary
if(separate_varlat) then
       do n = 1,size(tocn)
          Faxa_hmat(n)=Faxa_hlat(n)
       end do
else
       do n = 1,size(tocn)
          Faxa_hmat(n)=0._r8 ! avoid applying twice in BLOM
       end do
endif
```
So it seems that what BLOM should be using is Faxa_hlat. It would be helpful to look at the coupler history files that are produced in the coupled run to see what is actually being sent to the ocean.

mvertens:
@matsbn - more comments on the BLOM enthalpy implementation
```
               ! Heat flux components due to material material enthalpy flux of
               ! water exchange with atmosphere [W m-2]. Here, index_Faxa_hmat
               ! is related to enthalpy flux of evaporation and index_Faxa_hmoa
               ! related to the ocean average of all other enthalpy flux
	       ! components.
               util1(i,j) = fldlist(index_Faxa_hmat)%dataptr(n)*afac
               util2(i,j) = fldlist(index_Faxa_hmoa)%dataptr(n)*afac
```
In looking at the CMEPS code - Faxa_hmat_oa is the ocean average of Faxa_hmat before it is set to 0 if separate_varlat is .false. (which is the case). If we want to be consistent with CMEPS - I think that BLOM should be using Faxa_hlat above.

matsbn:
Thanks for looking into this, @mvertens! Thomas and I experimented with various methods to distribute the enthalpy flux into BLOM with the limitations that BLOM currently does not accept a mass surface flux. These methods are selectable with the `hmat_method` flag (which should eventually be a namelist option if we cannot determine a clearly preferred method). Some methods, like the `hmat_method = 4` in the PR, are cautious and distributes the global average flux over a large area, while others apply the flux more locally. Actually, in the simulation https://github.com/NorESMhub/noresm3_dev_simulations/issues/202 we went with `hmat_method = 2` and `separate_varlat = .true.` via SourceMods and we were happy to see that it seemed to work well. We should keep the exchange fields that allows this method to work.

PR #619, opened 2025-06-16
New default configuration for iHAMOCC

JorgSchwinger:
This PR makes the following changes to the default iHAMOCC configuration
* extended N-cycle is activated by default
* the GESAMP2018 dust/iron deposition is used by default (and iron-cycle tuning parameters are changed accordingly)
* Adjustments to sediment spinup default settings (resulting in a faster spinup)
* `dremcalc` and `disso_poc` are adjusted to give a balanced simulation with JRA-forcing on the tnx1v4 grid

This configuration is used in https://github.com/NorESMhub/noresm3_dev_simulations/issues/147

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -131,7 +131,7 @@
   <entry id="HAMOCC_FEDEP_SOURCE">
     <type>char</type>
     <valid_values>mahw2006,GESAMP2018</valid_values>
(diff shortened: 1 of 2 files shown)


TomasTorsvik (approved):
@JorgSchwinger - thanks! Looks fine to me.

PR #622, opened 2025-06-25
Fix restart for river2omip

TimotheeBrgs:
Small PR fixing a trivial issue related to the restart capability with `use_river2omip` activated. This PR works (compiles/runs) and leads to similar outputs when using or not restarts in a same simulation period.


JorgSchwinger (approved):

JorgSchwinger:
Looks good, thank you for fixing this. I would recommend to run a test with 1x2 years vs 2x1 years to make sure the run with restart is bit-for-bit compared with the run without restart.

jmaerz (approved):
Looks good to me. Thanks for fixing.

TimotheeBrgs:
Hi @JorgSchwinger , when I wrote "leads to similar outputs when using or not restarts in a same simulation period", I meant exactly what you recommend (1x2 years vs 2x1 years). `tDOC` species have exactly the same values with/without restart. Thank you @jmaerz .

PR #623, opened 2025-06-25
Changed namelist parameter RHISCF

matsbn:
This is the parameter change explored in simulation https://github.com/NorESMhub/noresm3_dev_simulations/issues/155. It reduces the eddy diffusivity suppression due to topographic beta effect.


jmaerz (approved):
Looks good to me. Thanks for evaluating.

TomasTorsvik (approved):

PR #631, opened 2025-08-08
Enhanced v1.6.9 - nitrification N2O/NO2 yield parameters available through nml

jmaerz:
Hi @TomasTorsvik , similar to master, I make two nitrification N2O/NO2-yield parameters available via the nml.


TomasTorsvik (approved):
Looks fine to me. Do you want a new `v1.6.x` tag now, or do you have further enhancements in mind?

jmaerz:
Hi @TomasTorsvik , I am currently running on the enhanced-v1.6.9 branch, so no real need for a new tag. Only thing at the moment is that the post-processing doesn't work for the hbgci-files (they aren't moved to the archive automatically; I don't know why, while also haven't looked deeper into it). I guess, it would be good to fix this before tagging.

PR #632, opened 2025-08-14
Update CFC/SF6 concentrations for historical simulation

JorgSchwinger:
With this PR we extend the CFC/SF6 time-series up to 2022 (Bullister et al. 2017). Data before 2008 are largely unchanged (round-off differences for SF6 are present), and moderately updated for 2009-2014. Data from 2015 onward are new.

This PR does neither change BLOM/HAMOCC simulation results nor the coupled model behavior, since CFCs/SF6 are passive tracers with no interactions to any other parts of the model. I have verified that results are bit-idetical with and without the flag `use_cfc` set, compared to a simulation without this modification.


hamocc/mo_get_cfc.F90
```diff
@@ -41,127 +41,130 @@ subroutine get_cfc(kplyear,atm_cfc11_nh,atm_cfc12_nh,atm_sf6_nh,
...
-    integer :: i
-    integer :: yr_dat(105)
-    integer :: start_yr
-    real(rp):: cfc_11_nh(105),cfc_12_nh(105),sf_6_nh(105)
-    real(rp):: cfc_11_sh(105),cfc_12_sh(105),sf_6_sh(105)
-    integer, save :: kplyear_old = 0
+    integer, parameter   :: nyears = 113
+    integer              :: i
+    integer              :: yr_dat(nyears)
+    integer              :: start_yr
```
TomasTorsvik: Would it make sense to make `start_yr` a parameter as well?

TomasTorsvik (approved):
@JorgSchwinger - thanks, this looks fine. I had one question about `start_yr`, but it's aso fine if you leave it as it is.

jmaerz (approved):
Hi @JorgSchwinger , thanks for updating! - looks good to me.

hamocc/mo_get_cfc.F90
JorgSchwinger: Yes, that makes sense, done

PR #634, opened 2025-08-14
New N-cycle parameter suggestion

jmaerz:

this is an update of new parameter suggestions for the N-cycle based on some own runs and Jörgs default setup run in https://github.com/NorESMhub/noresm3_dev_simulations/discussions/171 , where denitrification was too low and N2O sea-air fluxes were also too low. With the tuning, I focus on the water column.

**For explanation:**
- `bkamoxn2o` is lowered to allow for higher N2O production by first nitrification step also at low NH4 concentrations (changes the partitioning between N2O and NO2 pathway at low NH concentrations towards N2O)
- `rano3denit` is increased to allow for stronger first denitrification step (NO3 -> NO2) - providing more NO2 to DNRA, 2nd denitrification step and anammox, likely it will be still a bit too low
- `rdnra` is lowered to avoid potential too high loss of NO2 to NH4 (allowing for a stronger direct re-circulation from NO2 to NO3 by nitrification) - aiming at keeping the NO3 inventory stable
- `ran2odenit` is lowered to have lower final denitrification step in order to circulate N2O a bit longer (slightly reduced sink of N2O in ODZs) - allowing for higher N2O fluxes from atop of ODZs


TomasTorsvik (approved):
@jmaerz - thanks, looks fine to me!

JorgSchwinger (approved):

PR #635, opened 2025-08-14
Adjust default BLOM namelist parameters and number of vertical levels

matsbn:
This PR includes the following namelist changes:
- Minimum background vertical diffusivity is now available as namelist variable and increased from 1.e-6 m2 s-1 to 2.e-6 m2 s-1.
- Default eddy diffusivity parameters have been adjusted: EGC 3.0 -> 2.0; EGIDFQ 1.0 -> 0.8; EGMNDF 1.0 m2 s-1 -> 50.0 m2 s-1.
- Default mixed layer stratification method is changed from Fox-Kemper et al. (2008) to Bodner et al. (2023).
- Adaption of reference potential densities is enabled as default.

Consistent with testing of parameters related to the potential density adaption, the number of vertical layers for hybrid coordinate is increased from 56 to 72.

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -92,7 +92,7 @@
     <valid_values/>
     <default_value>53</default_value>
     <values>
-      <value compset="_BLOM%HYB">56</value>
+      <value compset="_BLOM%HYB">72</value>
       <value compset="_BLOM%LVL">56</value>
     </values>
     <group>build_component_blom</group>
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -1607,7 +1607,7 @@
     <category>diffusion</category>
     <group>diffusion</group>
     <values>
-      <value>1.e-6</value>
+      <value>2.e-6</value>
     </values>
     <desc>Minimum background diapycnal diffusivity (m**2/s) (f)</desc>
   </entry>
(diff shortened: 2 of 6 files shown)


JorgSchwinger (approved):
Thanks, Mats for putting this is. I will put in a few small changes of HAMOCC defaults very soon (Jöran put already in some parameter updates for the N-cycle), and then the new BLOM tag is probably good to go(?)

TomasTorsvik (approved):

TomasTorsvik:
For some reason, tests with `mpi == 'false'` fail on this PR.

Looks like we get a failing test for `fuk95`.

```
BLOM:fuk95 / run fuk95                 FAIL             0.43s   killed by signal 8 SIGFPE
```

Maybe this gets too resource intensive with 72 layers? I can do a "dummy" PR with 56 layers and see if this makes a difference.

matsbn:
> For some reason, tests with `mpi == 'false'` fail on this PR.
> 
> Looks like we get a failing test for `fuk95`.
> 
> ```
> BLOM:fuk95 / run fuk95                 FAIL             0.43s   killed by signal 8 SIGFPE
> ```
> 
> Maybe this gets too resource intensive with 72 layers? I can do a "dummy" PR with 56 layers and see if this makes a difference.

The number of layers used in the `fuk95` test case (25 layers) should be independent of what we set for NorESM configurations. I suspect the problem is that `./tests/fuk95/limits` has not been updated with the new `nubmin` namelist variable and a default has not been set. I see if I can correct it later today.

milicak (approved):

PR #637, opened 2025-08-15
For the CPPM transport method, correct incompatibility and improve accuracy for nonuniform grids

matsbn:
The enhancement of CPPM is two-fold:
1. The current implementation is not fully compatible and thus cannot guaranty monotonicity of tracers. The lack of compatibility also impacts the accuracy of tracer transport in the case of a highly varying thickness field. This PR provides a fully compatible implementation, but keeps the current partially compatible method as an option.
2. The order of accuracy of edge value reconstructions are now guaranteed also for nonuniform grids.

The fully compatible option is computationally more complex than the partial compatible option, but the additional computational complexity does not depend on the number of tracers. Benchmarks has shown that for NorESM applications with a high number of biogeochemical tracers, the run time is only increased by ~3% comparing full with partial compatibility. It should be mentioned that the lack of strict monotonicy of the partial compatible option might not be crucial for typical NorESM applications and the method will ensure positive definite tracer transport. Still, the fully compatible options is set as default.

The improved accuracy for nonuniform grids should not impact run time, as grid dependent coefficients are only computed during initialisation. This improvement is also expected to have a minor impact with the ocean grids used in NorESM, as grid cell sizes are generally smoothly varying.

BLOM diagnostics for years 11-20 of a OMIP2 type simulation, comparing enhanced CPPM using full compatibility option with current implementation (partial compatibility and edge value reconstruction assuming uniform grid), can be found here: https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/matsbn/NOIIAJRARYF8485OC_TL319_tn14_sra_cppm_fix_20250812/BLOM_DIAG/yrs11to20-NOIIAJRARYF8485OC_TL319_tn14_sra_cppm_pc_20250812/indexnew.html

BLOM diagnostics for years 11-20 of a OMIP2 type simulation, comparing full and partial compatibility options of the enhanced CPPM implementation, can be found here: https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/matsbn/NOIIAJRARYF8485OC_TL319_tn14_sra_cppm_fix_20250812/BLOM_DIAG/yrs11to20-NOIIAJRARYF8485OC_TL319_tn14_sra_cppm_pc_20250812/indexnew.html

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -404,6 +404,16 @@
     <desc>Advection method. Valid methods: 'remap', 'cppm'.</desc>
   </entry>
 
+  <entry id="cppm_compatibility">
+    <type>char</type>
+    <category>limits</category>
+    <group>limits</group>
+    <values>
+      <value>full</value>
+    </values>
+    <desc>CPPM compatibility. Valid options: 'full', 'partial'.</desc>
+  </entry>
+
   <entry id="cppm_limiting">
     <type>char</type>
     <category>limits</category>
--- a/phy/mod_cppm.F90
+++ b/phy/mod_cppm.F90
@@ -31,7 +31,7 @@ module mod_cppm
                      halo_ps, halo_us, halo_vs, halo_uv, halo_vv, &
                      lp, mnproc, nproc, jpr, nreg, itdm, i0, &
                      xctilr, xcstop
-   use mod_grid, only: scp2i
+   use mod_grid, only: scpx, scpy, scp2i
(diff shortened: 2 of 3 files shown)


JorgSchwinger (approved):

TomasTorsvik (approved):

milicak (approved):

jmaerz (approved):
Thanks for bringing this in - this is good news - strict monotonicity in tracer transport opens the option to remove/re-think certain iHAMOCC code parts in the future.

PR #638, opened 2025-08-15
New default parameters for HAMOCC

JorgSchwinger:
This PR sets a few new parameter defaults for HAMOCC. The goal is to reduce the vertical gradient in phosphate and nitrate, which has been consistently too strong in recent simulations. This is achieved by

* increasing the exponent b of the Martin-curve formulation from 1 to 1.3 by setting `wlin=60/3120`, which slows down the sinking of POC above 3120 m depth (beyond which the maximum sinking speed is reached)
* increasing `epsher` (which controls the efficiency of export by grazing) back to 0.8 (this parameter was decreased in the course of tuning when the split between euphotic zone and deep ocean was removed, but it turns out it was too much)

`ropal` is also slightly increased to keep the opal export approximately stable with the above changes. I expect some final tunings of sediment dissolution rates will be necessary to eliminate model drift, but this has to be done once a longer simulation (~200 years) is available.

@jmaerz I'm slightly puzzled by the fact that our b-exponent apparently needs to be that large (isn't otherwise 0.9 a ballpark-estimate?). It is probably reflecting the fact that the model has a low diffusivity, so maybe it is ok, or would you be worried about that?

### DIFF
--- a/hamocc/mo_param_bgc.F90
+++ b/hamocc/mo_param_bgc.F90
@@ -521,7 +521,7 @@ module mo_param_bgc
   real(rp), protected :: wdust_const                    ! m/d   Sinking speed of dust
   real(rp), protected :: wmin        =  5._rp           ! m/d   minimum sinking speed
   real(rp), protected :: wmax        = 60._rp           ! m/d   maximum sinking speed
-  real(rp), protected :: wlin        = 60._rp/2400._rp  ! m/d/m constant describing incr. with depth, r/a=1.0
+  real(rp), protected :: wlin        = 60._rp/3120._rp  ! m/d/m constant describing incr. with depth, r/a=1.3 (r=0.025)
   real(rp), protected :: dustd1      = 0.0001_rp        ! cm = 1 um, boundary between clay and silt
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):

jmaerz:
Hi @JorgSchwinger , I made some -rough- estimates on numerical diffusivity of the sinking scheme implemented by (i)HAMOCC for MPIOM and there, about 0.1 of 'effective' b (b measured by means fitting a power law curve to the fluxes modelled in MPIOM - if I remember correclt between 100 - 1000m) - and then compared to the 'analytically' set b value was just due to the numerical diffusion in the simple vertical sinking scheme. The increasing grid size with depth contributes additionally significantly to the diffusivity (known for the simple implicit 1st order scheme - or probably any scheme). I suspect that the T-dependency now further contributes to this effect at the global scale - so I am not really surprised by the fact that one needs to increase b (and thus imposes shallower remin - compensating for the high numerical diffusion of the implicit 1st order scheme). Thanks for providing the PR.

Issue #639, opened 2025-08-15
Bug in intitial conditions


TomasTorsvik:
A bug in the initial conditions has been discovered during testing of the `v1.11.0` tag intended for `noresm3_beta02`.

The error message was:

```
2143: At line 481 of file /cluster/projects/nn9560k/tomast/NorESM/src/nor3_b02-rc1/components/blom/phy/mod_inicon.F90
2143: Fortran runtime error: Index '0' of dimension 1 of array 'sp_src' below lower bound of 1
```

Looking into `mod_inicon.F90` I find:

```
    do j = 1, jj
      do l = 1, isp(j)
      do i = ifp(j,l), ilp(j,l)


        ! Create source arrays of practical salinity (sp_src), potential
        ! temperature (pt_src) and source interface depths bounded by model
        ! depth (z_src). Fill remaining missing values from above.
        do k = 1, kdm_woa
          if (t_woa(i,j,k) /= t_woa_fval .and. &
              t_woa(i,j,k) /= t_woa_mval .and. &
              s_woa(i,j,k) /= s_woa_fval .and. &
              s_woa(i,j,k) /= s_woa_mval) then
            sp_src(k) = s_woa(i,j,k)
            p_midlev = gsw_p_from_z(- depth_woa(k), plat(i,j))
            sa = gsw_sa_from_sp(sp_src(k), p_midlev, plon(i,j), plat(i,j))
            pt_src(k) = gsw_pt0_from_t(sa, t_woa(i,j,k), p_midlev)
          else
481:        sp_src(k) = sp_src(k-1)
            pt_src(k) = pt_src(k-1)
          endif
``` 

I'm not sure exactly what is going on, but it s aparent that we try to access `sp_src(0)` in this line.

This code was introduced in commit https://github.com/NorESMhub/BLOM/commit/faef4814f100eb02338cd8b6c6d096b070d4ece5 , and merged in [dev1.9.0.18](https://github.com/NorESMhub/BLOM/releases/tag/dev1.9.0.18)

matsbn:
@TomasTorsvik, I was not able to reproduce this issue with the NOIIAJRARYF8485OC case and T319_tn14 grid, compiled with `DEBUG=true`. The intended functioning is that `sp_src(0)` should never be accessed, since `t_woa(i,j,1)` and `s_woa(i,j,1)` should be free of missing values and thus preventing line 481 to be reached for `k=1`. Which grid and compiler are used in the failed test?

[...]

PR #640, opened 2025-08-17
Bug fix of WOA initialization

matsbn:
Fixes a bug in WOA initialization where grid cells in the exchange halo were erroneously accessed.

This should correct issue https://github.com/NorESMhub/BLOM/issues/639.


TomasTorsvik (approved):
@matsbn - great! Thanks for fixing this.

PR #642, opened 2025-08-18
Select fox08 ML restratification method for isopycnic coordinates

### LINKED ISSUE #641: BLOM fails on test with N2N1850 compset
The test case `N2N1850` is failing during testing for `noresm3_0_beta02`.

The error message points to line 1782 in `mod_eddtra.F90`. I believe the `N2N1850` compset should use `fox08` instead of `bod23`, so we need to set `fox08` when using this compset.

```
      ! Resolve mixed layer restratification options.
      select case (trim(mlrmth))
         case ('none')
            mlrmth_opt = mlrmth_none
            if (vcoord_tag == vcoord_isopyc_bulkml) then
               ce = 0._r8
            endif
         case ('fox08')
            mlrmth_opt = mlrmth_fox08
         case ('bod23')
            if (vcoord_tag == vcoord_isopyc_bulkml) then
               if (mnproc == 1) &
                  write (lp,'(3a)') &
                     ' init_eddtra: mlrmth = ', trim(mlrmth), &
                     ' is unsupported with vcoord = ''isopyc_bulkml''!'
1782:          call xcstop('(init_eddtra)')
                      stop '(init_eddtra)'
            endif
            mlrmth_opt = mlrmth_bod23
```

TomasTorsvik:
Fixes #641

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -430,6 +430,7 @@
     <group>limits</group>
     <values>
       <value>bod23</value>
+      <value blom_vcoord="isopyc_bulkml">fox08</value>
     </values>
     <desc>Mixed layer restratification method</desc>
   </entry>


jmaerz (approved):
Looks good to me - while maybe waiting for Mats to comment. In any case: the question is anyways, in how far the isopycnic bulkmixlayer is still supported parameter-wise?

JorgSchwinger (approved):
Looks reasonable to me, thanks for fixing @TomasTorsvik

matsbn (approved):
Thanks, @TomasTorsvik. Indeed, the `bod23` parameterization is not supported for `blom_vcoord="isopyc_bulkml"`.

With respect to @jmaerz comment, I must admit I have not had the capacity to carefully check the parameters selected for `blom_vcoord="isopyc_bulkml"` lately.

TomasTorsvik:
> Looks good to me - while maybe waiting for Mats to comment. In any case: the question is anyways, in how far the isopycnic bulkmixlayer is still supported parameter-wise?

We keep a `N2N1850` compset in the standard test-suite to make sure the isopycnic version builds and runs, but I don't think anyone has been running this to check if it makes sense parameter-wise, at least since early alpha testing.

jmaerz:
Hi @matsbn , no worries. For those interested, the last isopycnic bulkmixlayer version completely compatible with CMIP6 in tnx2 and tnx1 setups is release 1.6  (latest commits) - which is enhanced with most iHAMOCC features currently available in master (only exception: the new dust deposition). One non-fixed culprit, though, is that the iHAMOCC inventory files need to be transferred manually after running ...

PR #650, opened 2025-08-27
Add new PE layouts for tnx1v4 grid

JorgSchwinger:
This PR adds two new `patch.input` files with PE layouts for the tnx1v4 grid (using 4 and 5 nodes on betzy) with JRA forcing. The corresponding PE layouts are added in `config_pes.xml`.

The performance gain on betzy for for the larger PE layouts is somewhat disappointing with an increasing gap between BLOM only timings and the achieved total runtime (see comments in `config_pes.xml`)

I have verified that runs with different PE layouts for BLOM result in bit-for-bit identical results as long as the PEs for CICE are not changed. I have also seen that results are NOT bit-for-bit IF the number of CICE PEs is changed. The new PE-layouts in this PR all use 128 PEs for CICE.


TomasTorsvik (approved):
@JorgSchwinger - thanks for making this PR.

About the issue on CICE not being bit-for-bit with different PEs, is this something new or has this been the case earier (e.g. for NorESM2)?

JorgSchwinger:
> About the issue on CICE not being bit-for-bit with different PEs, is this something new or has this been the case earier (e.g. for NorESM2)?

I have never tested this myself, but according to @matsbn , the problem is present already in NorESM2 (or even earlier)

jmaerz (approved):
Thanks for providing.

PR #653, opened 2025-09-02
Add functionality to limit the local vertical layer thickness variation

matsbn:
This functionality applies to the ALE method when using `regrid_method = 'nudge'` and mitigates the issue of unsatisfactory layer structure that may develop after multidecadal integration times. A namelist variable `dpvar_fac` controls the layer thickness variation limiting and should be in the range [0, 1], reducing variation with increased value. Default value is `dpvar_fac = 0.75`.

NOINYOC_T62_tn21_20250823_atl_temp_full_0200-12

An Atlantic section showing model layers and potential temperature after 200 years with CORE normal year forcing. This simulation is without limiting of layer thickness variation.

NOINY_T62_tn21_dpvarlim_20250829_atl_temp_full_0200-12

Same as above, but with limiting of layer thickness variation and `dpvar_fac = 0.7`.

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -1306,7 +1306,7 @@
     <values>
       <value>.1</value>
     </values>
-    <desc>add desc</desc>
+    <desc>Minimum interior layer thickness [m]</desc>
   </entry>
 
   <entry id="regrid_method">
@@ -1336,7 +1336,7 @@
     <values>
       <value>86400.</value>
     </values>
-    <desc>add desc</desc>
+    <desc>e-folding time scale of interface nudging during regridding [s]</desc>
   </entry>
 
   <entry id="stab_fac_limit">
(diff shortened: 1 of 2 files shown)


JorgSchwinger (approved):
I cannot judge the details here, but on a general basis and the figures shown, I'm happy to approve this

TomasTorsvik (approved):

milicak (approved):

jmaerz (approved):
Similar to Jörg - I can't really judge the details. The only thing I am wondering about is, in how far spurious vertical (de-)mixing is introduced.

matsbn:
> Similar to Jörg - I can't really judge the details. The only thing I am wondering about is, in how far spurious vertical (de-)mixing is introduced.

The limiting of thickness variation is part of the regridding algorithm, so the remapping from old to new vertical grid happens as frequent as before with the same high-order method. This is just to say that there is no adjustment of layer structure that needs an additional remapping, which can be a source of spurious mixing. In my view, the main potential impact on vertical structure/mixing is that desired sharp vertical density gradients (e.g thermocline), which require large thickness variation, might be represented less accurately. On the other hand, without the limiting of thickness variation, we have seen that the layer structure has slowly creeped towards undesired density gradients, which is likely a feedback in the repeated loop of reconstructing high-order temp./saln. profiles, seeking target density intersections on these profiles, and remapping.

PR #656, opened 2025-09-09
Set types for real and integer variables

TomasTorsvik:
I came across some instances that used `integer*2`, `real*4` and `real*8` instead of the type settings `i2`, `r4` and `r8`. I'm not sure if these were intentionally set this way, in particular for interface with external libraries and MIP, so I'm setting this as a draft PR for now.

The `aux_blom_noresm` tests pass with this PR, except for expected fails.

### DIFF
--- a/ben02/mod_ben02.F90
+++ b/ben02/mod_ben02.F90
@@ -321,8 +321,8 @@ contains
     character :: filename*(*)
 
     integer :: status,ncid,varid,i,j
-    real*4 r4lon(atm_idm),r4lat(atm_jdm)
-    integer*2 i2field(atm_idm,atm_jdm)
+    real(r4) :: r4lon(atm_idm),r4lat(atm_jdm)
+    integer(i2) :: i2field(atm_idm,atm_jdm)
 
     ! --- Open netcdf file.
     status = nf90_open(filename,nf90_nowrite,ncid)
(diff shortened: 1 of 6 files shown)


JorgSchwinger (approved):

jmaerz (approved):
Hi Tomas,

thanks. I wonder a bit, if there is a good reason, not to include the `mo_kind` of iHAMOCC for the iHAMOCC routines? So using `mo_kind` instead of `mod_types` for `mo_aufr_bgc` and `mo_aufw_bgc`.

TomasTorsvik:
@jmaerz - I suspect what we really want is something like `kind=MPI_INTEGER` when we are getting data from the mpi library. However, it seems that `MPI_INFO` may be an object that should have a TYPE declaration, so I'm not sure what is correct.

TomasTorsvik:
It seems I was mistaken about the TYPE declaration, this only applies if we invoke the `mpi_f08` library explicitly (i.e. `use mpi_f08` instead of `use mpi`). With the traditional way we still declare the MPI communication variables as INTEGER. However, it seems `mpi_f08` is the recommended library for new development.
```
- Modern Fortran Standard Compliance:
mpi_f08 leverages features like assumed-type and assumed-rank arguments, derived types for handles (e.g., TYPE(MPI_Comm) instead of INTEGER), and explicit interfaces, which align with modern Fortran standards.

- Compile-time Argument Checking:
Unlike INCLUDE 'mpif.h', mpi_f08 guarantees compile-time argument checking for most arguments, helping to catch errors early in the development process.

- Improved Type Safety:
By using derived types for MPI handles, mpi_f08 enhances type safety and prevents accidental misuse of handles of different types.

- Support for Nonblocking Operations:
mpi_f08 provides the necessary mechanisms to fully address optimization problems associated with nonblocking MPI calls, often utilizing the ASYNCHRONOUS attribute where supported by the compiler.
```

matsbn (approved):
Thanks, @TomasTorsvik. I think this should be fine.

PR #660, opened 2025-09-16
Updated and optimized checksum functionality

matsbn:
The checksum functionality has been rewritten and optimised, particularly for better parallel performance. This should not impact results and computing checksums remains off by default. The checksum/hash is still CRC32 (Cyclic Redundancy Check 32-bit), but the C code (crc.c, crc.h) used for the CRC32 computation has been replaced by a Fortran module (mod_crc32.F90).


TomasTorsvik (approved):

jmaerz (approved):
Thanks for increasing the parallel performance!

PR #665, opened 2025-09-18
Add expected fail entry for ERS_Ld3.TL319_tn05.NOIIAJRA

TomasTorsvik:
For documentation:
I would like to add an entry for `ERS_Ld3.TL319_tn05.NOIIAJRA` in the list of expected fail tests, as I do not expect we will resolve issues with the `TL319_tn05` grid for the `noresm3_0_beta03` tag.


jmaerz (approved):
Good to keep a record of failing tests - to tackle the issues at given time. Thanks for keeping the books up-to-date.

PR #666, opened 2025-09-20
Set woa_nuopc_provided FALSE by default

TomasTorsvik:
I would like to disable the ` woa_nuopc_provided` option for now. I still get tests that hang on creation of the `woa18_s_an.nc` file. See e.g. https://github.com/NorESMhub/NorESM/issues/722

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -151,7 +151,7 @@
     <group>limits</group>
     <values>
       <value>.false.</value>
-      <value comp_interface="nuopc">.true.</value>
+      <value comp_interface="nuopc">.false.</value>
       <value comp_interface="nuopc" blom_vcoord="isopyc_bulkml">.false.</value>
     </values>
     <desc>if .true., use NUOPC capability to read WOA climatology</desc>


mvertens:
@TomasTorsvik - I'm really surprised this is happening. Can I please make an alternative PR where I simply turn off the creation of these files with a debug flag?

TomasTorsvik:
@mvertens - Yes, that should be fine. I tested several times with the woa_nuopc enabled, as by default, and once more where I disabled this option through the namelist.

See `/cluster/work/users/tomast/noresm/nor3_b03-rc2/` where the tests passed, compared with `/cluster/work/users/tomast/noresm/nor3_b03-rc1/`.

It is curious that the `ERI.ne16pg3_tn14.N1850fates-nocomp.betzy_intel` test first reported error in COMPARE_base_hybrid, and on later attempts failed on woa_nuopc.

TomasTorsvik:
Alternative solution suggested by @mvertens

mvertens:
@TomasTorsvik - I have created a new PR where the plots are turned off by default. Having the new mapped woa will change answers compared to baselines. I did run the aux_blom_noresm tests with the plots being generated and did not have problems with my changes - so this problem is very surprising.

TomasTorsvik:
I re-open this PR, since #667 did not solve the problem with failing tests.

mvertens (approved):

mvertens:
@matsbn @TomasTorsvik - did you experience hangs in writing the salinity file when running tests or also when just doing a production run?
I am having a hard time duplicating this behavior - so it would be good to know if it's associated with testing or can also occur when just doing a single run.

TomasTorsvik:
@mvertens - as far as I can remember, I only had this problem when running tests. I have not been running so many single runs lately.

TomasTorsvik:
Could it make a difference that I am running multiple tests at the same time? In principle, I would not expect that for read access, but I don't know how the data is handled internally on the system (data caching and IO bottlenecks).

matsbn:
@mvertens, I believe I had hangs with NOIIAJRA* experiments in addition to tests.

matsbn (approved):

mvertens:
@matsbn @TomasTorsvik - thanks for your feedback.

mvertens:
@TomasTorsvik - using woa_nuopc_provided = .true.
ERI.TL319_tn14.NOIIAJRAOC.betzy_intel passes  COMPARE_base_hybrid 
whereas
ERI.ne16pg3_tn14.N1850fates-nocomp.betzy_intel fails COMPARE_base_hybrid
This phase of the ERI tests is a branch run from the ref2 restart created.

PR #667, opened 2025-09-20
turn off mapping plots that cause problems with hangs in testing

mvertens:
This PR turns off the default creation of woa18_t_an.nc and woa18_s_an.nc. This is disabled via a module variable plot_woa_data = .false. that can be modified to be .true.
I will come up with a more sophisticated method that only turns this off for testing and turns it on for longer production runs. But for now I think this is the simples way forwards.

### DIFF
--- a/drivers/nuopc/ocn_map_woa.F90
+++ b/drivers/nuopc/ocn_map_woa.F90
@@ -30,6 +30,8 @@ module ocn_map_woa
 
    public :: map_woa
 
+   logical :: plot_woa_data = .false.
+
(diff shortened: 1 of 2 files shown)


matsbn (approved):
I have also experienced model hangs recently due to the writing of the WOA climatology, so I support this fix for now.

mvertens:
@TomasTorsvik - I am seeing the same failure you saw with ERI.ne16pg3_tn14.N1850fates-nocomp.betzy_intel - so since this is not an expected failure I think we need to go back to setting `woa_nuopc_provided = .false.` by default for now.

PR #668, opened 2025-09-20
Initialize arrays that where accessed with uninitialized values

matsbn:
With the updated checksum functionality (https://github.com/NorESMhub/BLOM/pull/660), halo update (call to `xctilr`) is carried out for a copy of the array to be checksummed. This led to access of uninitialized array elements, which this PR corrects. 

The call to `xctilr` in subroutine `map_woa` also led to access to uninitialized array elements, which is also corrected. I do not think this should impact the issues in https://github.com/NorESMhub/BLOM/pull/666 and https://github.com/NorESMhub/BLOM/pull/667, but might be good be aware of, @mvertens.

The PR should not change any results. Note also that these accesses to uninitalized array elements are not trapped by the regular NorESM debug options. For the Intel compiler, the compiler option `-init=snan,arrays` is needed to trap these accesses and we should consider to make them default for BLOM debug options.

### DIFF
--- a/drivers/nuopc/ocn_map_woa.F90
+++ b/drivers/nuopc/ocn_map_woa.F90
@@ -131,6 +131,8 @@ contains
          call xchalt('(ocn_map_woa)')
          stop '(ocn_map_woa)'
       endif
+      t_woa(:,:,:) = t_woa_fval
+      s_woa(:,:,:) = s_woa_fval
 
       ! ---------------------------
       ! Create input data mesh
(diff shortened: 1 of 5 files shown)


TomasTorsvik (approved):
@matsbn - thanks for fixing!

TomasTorsvik:
@matsbn - I will make a new BLOM tag for `noresm3_0_beta03`. I suppose we want this PR in as well?

JorgSchwinger (approved):

matsbn:
> @matsbn - I will make a new BLOM tag for `noresm3_0_beta03`. I suppose we want this PR in as well?

I think the PR is good to include in noresm3_0_beta03 as checksum verification will be more robust. It should not change answers and not impact the regular tests.

TomasTorsvik:
@matsbn - I will go ahead and merge this, and proceed to re-run tests for `noresm3_0_beta03`.

PR #669, opened 2025-09-24
speed up eri tests

mvertens:
An ERI test has 4 phases:
- ref1case
Do an initial run for 3 days writing restarts at day 3. ref1case is a clone of the main case. Short term archiving is on.

- ref2case (Suffix hybrid)
Do a hybrid run for default 19 days running with ref1 restarts from day 3, and writing restarts at day 10. ref2case is a clone of the main case. Short term archiving is on.

- case
Do a branch run, starting from restarts written in ref2case, for 9 days and writing restarts at day 5. Short term archiving is off.

- case (Suffix base)
Do a restart run from the branch run restarts for 4 days. Compare component history files ‘.base’ and ‘.hybrid’ at day 19. Short term archiving is off.

In the above cases BLOM writes out daily hd and hbgcd - whereas all other components do not write out these files UNLESS the `testmods="allactive/defaultio"`. This PR brings BLOM closer to behaving like the other components, decreases the time for ERI tests and also decreases the number of files created in these tests. This is particularly problematic when file system problems occur (such as has been the case with Betzy latetly).

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -277,7 +277,8 @@ def buildnml(case, caseroot, compname):
         if is_test:
             testcase = case.get_value("TESTCASE")
             config["is_test_pfs"] = "yes" if testcase == "PFS" else "no"
-            config["empty_hist"] = "yes" if testcase == "PFS" else "no"
+            config["is_test_eri"] = "yes" if testcase == "ERI" else "no"
+            config["empty_hist"] = "yes" if (testcase == "PFS" or testcase == "ERI") else "no"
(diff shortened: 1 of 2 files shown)


TomasTorsvik (approved):
@mvertens - thanks for making the PR, looks reasonable to me.

TomasTorsvik:
Will tag this as BLOM `v1.12.5`, since it does not include code changes.

mvertens:
@TomasTorsvik - thanks!

Issue #662, opened 2025-09-16
Different checksums for some iHAMOCC fields after restart


matsbn:
@JorgSchwinger, @jmaerz, @TomasTorsvik,

We have been struggling with a failing restart test as discussed here https://github.com/NorESMhub/NorESM/issues/716. In an effort to understand the failing test, I wrote detailed checksums of BLOM fields in the time step following a restart and compared with checksums of a run that did not restart (to be able to do this, the updated checksum functionality in https://github.com/NorESMhub/BLOM/pull/660 was needed).

A bit surprisingly, with the BLOM updates (BLOM v1.12.1 including https://github.com/NorESMhub/NorESM/issues/716), the previously failing test (ERS.ne30pg3_tn14.N1850fates-nocomp.betzy_gnu.allactive-defaultio) now passed. However, when inspecting the checksums, there were different checksums for tracer field numbers 18, 19 and 23-27. This can be seen when comparing the ocean log files:
```
/cluster/work/users/matsbn/noresm/ERS.ne30pg3_tn14.N1850fates-nocomp.betzy_gnu.allactive-defaultio.nor3_b2_uia_ocn_import_v13/run/ocn.log.*
```
Here is a screenshot of a comparison I did, highlighting the first checksum mismatches found after calling `ale_regrid_remap`:
<img width="864" height="882" alt="Image" src="https://github.com/user-attachments/assets/cd7699fa-1a17-47d0-bec8-9cb4f8bd1d06" />
For subsequent checksums (created the first time step of a model day), mismatches prevail in these tracer fields and there is an occurrence of `dustflx2000` having a different minimum value during writing of output file.

I can do furter tests, for example checking the tracer array before entering `ale_regrid_remap`, but wanted first to check if this is expected behaviour.

jmaerz:
Hi @matsbn many thanks. I was just a bit looking into your output and tried to figure out, which tracers are affected (i.e. mapping the tracer number to tracer names). It looks to me as if `idicsat` and `ibromo` correspond to Trac. 18+19. 23-26 to the preformed tracers (PO4,alk,dic,silica, while 22 to O2) and I am confused about tracer 27. 

Wrt bromoform: I was/am a bit confused now with the changes made in https://github.com/NorESMhub/BLOM/pull/658 - where bromoform was set differently than the other fluxes (I understand N2O and NH3, which are likely so far not supported, while bromoform should be coupled, or not?)

TomasTorsvik:
[...]

PR #670, opened 2025-09-25
Bugfix of non-reproducibility after restart for lateral diffusivity smoothing

matsbn:
This PR corrects the non-reproducibility after restart described in https://github.com/NorESMhub/BLOM/issues/662. The lateral smoothing of diffusivity led to roundoff difference between directly computing diffusivity at the Arctic seam of tripolar grids (non-restart) and halo update (restart). This PR will change results at roundoff level when smoothing of diffusivity is requested.

Other changes (minor code clean-up, memory saving) that do not impact results are also included in this PR.

### DIFF
--- a/phy/mod_ale_regrid_remap.F90
+++ b/phy/mod_ale_regrid_remap.F90
@@ -43,7 +43,7 @@ module mod_ale_regrid_remap
                             extract_polycoeff, regrid, &
                             prepare_remapping, remap, &
                             hor3map_noerr, hor3map_errstr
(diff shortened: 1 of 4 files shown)


TomasTorsvik:
@matsbn - thanks for fixing! I suggest this can be tagged as `v1.12.6` as a bugfix, since changes are only expected at roundoff level.

TomasTorsvik (approved):

JorgSchwinger (approved):

PR #672, opened 2025-09-26
fixes for eri test failures in prealpha testing

mvertens:
With these changes verified that ERI.ne3pg3_tn21.N1850fates-nocomp.betzy_intel now passes.


TomasTorsvik (approved):
@mvertens - thanks for providing the bug fix!

PR #673, opened 2025-10-06
Use high-order vertical reconstructions for remapping to diagnostic z-levels.

matsbn:
In the PR, the already computed high-order vertical reconstructions for ALE regridding/remapping are reused to remap potential temperature, salinity, ideal age and velocity components to diagnostic z-levels.

To illustrate the impact, monthly average potential temperature in a Pacific South-North section is shown below. The upper panel is with the current 1. order remapping the z-levels, the middle panel is using the proposed high-order vertical reconstructions, and the lower panel is the difference. With the 1. order remapping, piecewise constant imprints of layers are visible in the z-level diagnostics.

Screenshot 2025-10-06 at 08 52 36
Screenshot 2025-10-06 at 08 52 39
Screenshot 2025-10-06 at 08 52 11

Also, when using the ALE method, the vertical dimension of layer diagnostic is changed from `sigma` to `layer`. This is because `sigma` may vary in time and is therefore not well suited as a dimension, causing issues when concatenating layer diagnostic variables in e.g. xarray. This will require some changes to the NorESM standard diagnostics and I have a workaround for this that I will discuss with @YanchunHe.

In the implementation of using high-order reconstructions for remapping to diagnostic z-levels, changed results beyond the modified diagnostics are expected. This is due to roundoff differences in the computation of the column pressure range, that reduces code complexity and improves performance.

I have not detected any consistent performance degradation with the code changes introduced in this PR. However, during testing on Betzy, computational performance with and without these changes have been quite variable, so any performance impact should be further monitored.

### DIFF
--- a/phy/mod_ale_regrid_remap.F90
+++ b/phy/mod_ale_regrid_remap.F90
@@ -33,7 +33,8 @@ module mod_ale_regrid_remap
                             vcoord_cntiso_hybrid, vcoord_plevel, &
                             sigmar, plevel
    use mod_eos,       only: sig, dsigdt, dsigds
-   use mod_state,     only: u, v, dp, dpu, dpv, temp, saln, sigma, p, pu, pv
+   use mod_state,     only: u, v, dp, dpu, dpv, temp, saln, sigma, p, pu, pv, &
+                            ub, vb
    use mod_hor3map,   only: recon_grd_struct, recon_src_struct, remap_struct, &
                             hor3map_plm, hor3map_ppm, hor3map_pqm, &
                             hor3map_monotonic, hor3map_non_oscillatory, &
@@ -46,8 +47,15 @@ module mod_ale_regrid_remap
    use mod_diffusion, only: ltedtp_opt, ltedtp_neutral, difmxp
    use mod_ndiff,     only: ndiff_prep_jslice, ndiff_uflx_jslice, &
                             ndiff_vflx_jslice, ndiff_update_trc_jslice
+   use mod_dia,       only: ddm, nphy, alarm_phy, &
+                            depthslev_bnds, pbath, ubath, vbath, phylvl, &
+                            acc_templvl, acc_salnlvl, &
+                            acc_uvellvl, acc_vvellvl, &
+                            acc_idlagelvl
    use mod_checksum,  only: csdiag, chksum
-   use mod_tracers,   only: ntr, trc
+   use mod_tracers,   only: ntr, itriag, trc
+   use mod_ifdefs,    only: use_TRC, use_IDLAGE
+   use mod_utility,   only: util1
 
    implicit none
    private
@@ -96,7 +104,7 @@ module mod_ale_regrid_remap
                                       ! construction of Bezier curves.
    integer, parameter :: &
         regrid_method_direct = 1, &   ! Regrid method (vcoord_tag ==
-                                      ! vcoord_cntiso_hybrid): On the basis of 
+                                      ! vcoord_cntiso_hybrid): On the basis of
                                       ! reconstructed potential density, regrid
                                       ! interface pressures so interface
                                       ! potential densities match target values.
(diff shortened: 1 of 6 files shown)


TomasTorsvik:
@matsbn - thanks for this update. Does this PR turn on ALE regridding/remapping  by default or as an optional feature for further testing? I think it would be fine to include it as optional, but I'm more skeptical to change the default without at least some longer test simulations. Also, will diagnostics be able to hanle model output both before and after this PR is introduced? We might want to run earlier versions (e.g. NorESM2.3 with some BLOM/iHAMOCC updates) where it would be useful to have the standard diagnostics available.

matsbn:
> @matsbn - thanks for this update. Does this PR turn on ALE regridding/remapping by default or as an optional feature for further testing? I think it would be fine to include it as optional, but I'm more skeptical to change the default without at least some longer test simulations. Also, will diagnostics be able to hanle model output both before and after this PR is introduced? We might want to run earlier versions (e.g. NorESM2.3 with some BLOM/iHAMOCC updates) where it would be useful to have the standard diagnostics available.

@TomasTorsvik, this PR only impacts diagnostics and results when `vcoord_type = "cntiso_hybrid"`. For `vcoord_type = "isopyc_bulkml"` results are bit-identical in my testing (and `sigma` is kept as vertical dimension for layered diagnostics). iHAMOCC output format is unchanged, although one should consider a time-invariant vertical dimension here also when using `vcoord_type = "cntiso_hybrid"`. There is also an opportunity to use high-order reconstructions for remapping of iHAMOCC variables to diagnostic z-levels, since they are already computed, but this would require quite some work to interface with the iHAMOCC diagnostic routines.

I ran a 5-year simulation, branched of our current ocean/sea-ice spinup https://github.com/NorESMhub/noresm3_dev_simulations/issues/233 at model date 0051-01-01. With modest changes in the BLOM package of `diag_run`, I could run the standard diagnostics and also compare to the previous simulation with different dimensions for layered variables. The standard diagnostics difference for years 51-55 are available here:

https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/matsbn/NOIIAJRARYF8485OC_TL319_tn14_moddia_20251003/BLOM_DIAG/yrs51to55-NOIIAJRARYF8485OC_TL319_tn14_beta02_blom_v1.12.0_20250904/

Updated standard diagnostics should be available at the same time as a potential merge of the PR and I will discuss this with Yanchun.

TomasTorsvik:
@matsbn - OK, I think we need to do some work on the iHAMOCC side as well. It would be best if all tracer output use the same reconstruction. I will open a issue on this for now, and we can discuss how to proceed.

YanchunHe:
Hi @matsbn , I have made some changes in the diagnostic tools accordingly, https://github.com/NorESMhub/NorESM-Diagnostics/pull/8

Any of your comments will be very welcome. If looks OK, I will merge the update.

TomasTorsvik (approved):
I think this can be merged once the diagnostics update is in place.

JorgSchwinger:
I have a few questions/comments

* with the new remapping and adaptive layers, layer structure in principle changes every time step(?). I am wondering whether the averaging on layers in HAMOCC is then correct any more at all? Probably not, since it is meant to be on the old sigma layers.
*  For cmorization it then definitely does not make sense any longer to write out variables at "native" layers (since it is one variable per file, multiple month or years). It never made much sense (in my opinion) because it was confusing and not much used (if at all), so this option should be removed from the cmor tool.
* I can't really judge what it would take to implement a corresponding remapping for HAMOCC, but of course it would be preferable. If we want to do this for fast-track simulations we need a plan urgently...

Maybe it would also be go to discuss these things at a BLOM core meeting?

matsbn:
> I have a few questions/comments
> 
> * with the new remapping and adaptive layers, layer structure in principle changes every time step(?). I am wondering whether the averaging on layers in HAMOCC is then correct any more at all? Probably not, since it is meant to be on the old sigma layers.

It is not correct any more in the sense of an averaged quantity for a layer with constant potential density. So layer averaged quantities must be interpreted with care and the practical usefulness is likely reduced. For development purposes, layer averaged quantities are still essential in BLOM.

> * For cmorization it then definitely does not make sense any longer to write out variables at "native" layers (since it is one variable per file, multiple month or years). It never made much sense (in my opinion) because it was confusing and not much used (if at all), so this option should be removed from the cmor tool.

I agree on this. This is also a motivation to make the z-level diagnostics as good as possible.

> * I can't really judge what it would take to implement a corresponding remapping for HAMOCC, but of course it would be preferable. If we want to do this for fast-track simulations we need a plan urgently...

I had a quick look at the accumulation of HAMOCC diagnostic variables. What seems to complicate matters is that some accumulated fields are derived from prognostic BGC tracers (here prognostic tracers are what is in the `trc` array in BLOM). That would be messy to deal with in the BLOM ALE routines.
 
> Maybe it would also be go to discuss these things at a BLOM core meeting?

I think that is a good idea! I'll try to invite soon.

matsbn:
Confirmed bit-identical results in NOIIAJRARYF8485OC compset after merging in latest changes from master. Will proceed to merge PR.

PR #678, opened 2025-10-20
Use NorESMhub fork and noresm tag for GSW-Fortran submodule

### LINKED ISSUE #602: Do we need a fork of TEOS-10/GSW-Fortran ?
I recently had a problem with missing tags on the "Icepack" submodule for CICE, which is missing tags in the main repository. As a solution, I forked "Icepack" to NorESMhub in order to make a tag version. I see there might be a similar problem for "TEOS-10/GSW-Fortran" which also does not have any recent tags. However, GSW-Fortran does not evolve much (last update 2 years ago). 

Do we need a local for of TEOS-10/GSW-Fortran, or do we keep the branch reference (which should always point to the latest version)? I think it would be safer to have a local fork to avoid update issues later on.

TomasTorsvik:
This PR replaces the external https://github.com/TEOS-10/GSW-Fortran repository with a locally forked https://github.com/NorESMhub/GSW-Fortran repository, and makes the reference point at a new tag [gsw-fortran_v3.05-6_noresm_v1](https://github.com/NorESMhub/GSW-Fortran/releases/tag/gsw-fortran_v3.05-6_noresm_v1).

This ensures that the git-fleximod tool will check out a static tag version of GSW-Fortran, even if the TEOS-10/GSW-Fortran master branch should evolve.

Closes #602

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = https://github.com/CVMix/CVMix-src
 	fxtag = v0.98-beta
 	fxrequired = AlwaysRequired
-        fxDONOTUSEurl = https://github.com/CVMix/CVMix-src
+	fxDONOTUSEurl = https://github.com/CVMix/CVMix-src
 [submodule "pkgs/M4AGO-sinking-scheme"]
 	path = pkgs/M4AGO-sinking-scheme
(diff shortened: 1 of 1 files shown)


matsbn (approved):
Thanks, @TomasTorsvik! Good to include a static version in NorESM.

PR #682, opened 2025-11-02
Upgrade timing

matsbn:
The current implementation of the timing of various tasks is cumbersome to maintain and extend. Further, it only provides timing statistics of the master process, which might not give reliable results in parallel execution. This PR simplifies the interface to the timing module, which now handles all the bookkeeping and statistical output of various timers. During parallel execution, timers for all processes are now active allowing for more reliable timing statistics. The timing of various tasks during a BLOM time step has been revised and, particularly when the ALE method is used, more precise timing statistics are provided.

This PR does not change results and only modifies time statistics to stdout (ocn.log.* in NorESM).


TomasTorsvik (approved):

milicak (approved):

JorgSchwinger (approved):

matsbn:
The PR has been complemented with commits that gives better timing consistency during parallel execution and ensures the new timing functionality is supported when using MCT.

PR #684, opened 2025-11-03
Corrects layer thickness tendency in vertical mass flux diagnostics.

matsbn:
As discovered by @YanchunHe, the diagnosed vertical mass flux gives inconsistent results in its current implementation. This is because the previous time-step layer thickness is replaced with current layer thickness before a necessary layer thickness tendency is computed.

Except for changes in `wflx*` diagnostic fields, results are bit-identical with current master.


TomasTorsvik (approved):
@YanchunHe , @matsbn - thanks for bugfix!

milicak (approved):

YanchunHe:
Thanks, Mats for quickly fix this 👍

PR #687, opened 2025-11-12
Add PE-layouts for CPLHIST and JRA/tn21 forcings

JorgSchwinger:
This PR adds two PE-layouts for CPLHIST forcing (to be used for BLOM/CICE/iHAMOCC NorESM3 offline spinup) and one PE-layout for the tn21-grid and JRA forcing.


TomasTorsvik (approved):
Looks fine to me.

jmaerz (approved):

PR #691, opened 2025-11-14
fix interim restart capability

mvertens:
This requires an update CIME and needs to be coordinated across all components - hence it's a draft for now.

### DIFF
--- a/cime_config/config_archive.xml
+++ b/cime_config/config_archive.xml
@@ -6,7 +6,7 @@
     <hist_file_extension>hbgc[idmy]\d*.*\.nc$</hist_file_extension>
     <rest_history_varname>unset</rest_history_varname>
     <rpointer>
-      <rpointer_file>rpointer.ocn$NINST_STRING</rpointer_file>
+      <rpointer_file>rpointer.ocn$NINST_STRING.$DATENAME</rpointer_file>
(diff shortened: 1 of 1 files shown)


matsbn (approved):
I just approves this now, since when coordinated tests of the multi component and infrastructure PRs passes, this should obviously go in.

TomasTorsvik:
@mvertens - I was thinking it would be useful to have this in as the next PR, an tag it as `v1.12.16`, so that we have a minimal delta relative to `noresm3_0_beta06` (i.e.`v1.12.15`), and therefore putting on hold merging other PRs at the moment. When do you expect that we can merge this?

mvertens:
@TomasTorsvik - I think that would be fine. As long as you don't turn on interim restarts that should work. Can you run the aux_blom_noresm regression tests to verify this? Everything should be bit-for-bit.

TomasTorsvik (approved):

TomasTorsvik:
@mvertens - I did a run through `aux_blom_noresm` regression tests, and can confirm that the current PR is bit-for-bit with `noresm3_0_beta06`. Would it be OK to merge and tag this PR in BLOM now, or do you think there could be more changes coming in for this PR?

mvertens:
@tomas- thanks! Merging now sounds good.

TomasTorsvik:
This is now tagged as `v1.12.16`.

PR #692, opened 2025-11-18
Olivia pes settings

monsieuralok:
New Ocean only configuration that works on Olivia.

### DIFF
--- a/cime_config/config_pes.xml
+++ b/cime_config/config_pes.xml
@@ -163,6 +163,44 @@
   </grid>
 
   <grid name="a%T62.+oi%tnx1v4">
+    <mach name="olivia">
+      <pes pesize="M" compset="_DATM.*_BLOM.*%ECO">
+        <comment>none</comment>
+        <ntasks>
+          <ntasks_atm>140</ntasks_atm>
+          <ntasks_rof>140</ntasks_rof>
+          <ntasks_ice>140</ntasks_ice>
+          <ntasks_ocn>500</ntasks_ocn>
+          <ntasks_cpl>140</ntasks_cpl>
+          <ntasks_lnd>140</ntasks_lnd>
+          <ntasks_glc>140</ntasks_glc>
+          <ntasks_wav>140</ntasks_wav>
+        </ntasks>
+        <nthrds>
+          <nthrds_atm>1</nthrds_atm>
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):

JorgSchwinger (commented):
Are on Olivia always whole nodes charged (as on betzy)? In this case this setup would be highly inefficient (128 idle cpus), and I would suggest to use ntasks=623 for BLOM instead (only 5 idle cpus). Otherwise it's fine.

monsieuralok:
> Are on Olivia always whole nodes charged (as on betzy)? In this case this setup would be highly inefficient (128 idle cpus), and I would suggest to use ntasks=623 for BLOM instead (only 5 idle cpus). Otherwise it's fine.

Olivia has 255/256 cpu cores per node. I guess BLOM efficient version  has max 500 CPU cores. Therefore, I have chosen that. Second, rightnow I am committing what is working more than efficiency as some are failing bit strangely.

JorgSchwinger:
We do have a 623 BLOM configuration for the tn14 grid, and I am using this currently for the CPLHIST spinup. Scaling is maybe not extraordinary, but definitely more efficient than with 128 idle cpus...

monsieuralok:
> We do have a 623 BLOM configuration for the tn14 grid, and I am using this currently for the CPLHIST spinup. Scaling is maybe not extraordinary, but definitely more efficient than with 128 idle cpus...

Thanks, I did not knew that; I will try that

monsieuralok:
> We do have a 623 BLOM configuration for the tn14 grid, and I am using this currently for the CPLHIST spinup. Scaling is maybe not extraordinary, but definitely more efficient than with 128 idle cpus...

I have got following throughput 
NPES_OCN 623 36.24 models/year
NPES_OCN 500 31.86 models/year

I will update to 623 processors count

JorgSchwinger:
Thanks for testing this - looks like Olivia is substantially faster than betzy!

TomasTorsvik:
@monsieuralok , @JorgSchwinger , @matsbn - Is this ready for merging now, or do we need more testing of PES settings?

TomasTorsvik:
@monsieuralok , @JorgSchwinger , @matsbn - I will merge this now. If we need new PES settings later, it can come in a different PR.

PR #694, opened 2025-11-19
fix to prepare for NUOPC NH3 coupling

jmaerz:
Same fix, but for master.


TomasTorsvik (approved):

TomasTorsvik:
@jmaerz - Thans, good to have this in `master` as well.

PR #700, opened 2025-12-10
Make variable wlin accesible through user namelist

JorgSchwinger:
This PR is needed to make it possible to use wlin in `user_nml_blom` as a tuning variable. It has no effect on anything else, and it would be good to include it into a NorESM3 tag as soon as possible. It might be needed when re-coupling the BLOM CPLHIST spinup .

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -5050,8 +5050,8 @@
     <group>bgcparams</group>
     <values>
       <value>None</value>
-      <value use_wlin="yes">0.7</value>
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):

TomasTorsvik:
@JorgSchwinger - Thanks for the PR! The `noresm3_0_beta08` has not been merged yet (waiting for reviews), so we may include it there.

jmaerz (approved):
Looks good.

TomasTorsvik:
I think I will just go ahead and merge this, so that we can have it in `noresm3_0_beta08`

PR #707, opened 2026-01-11
fix pelayout for T62 and update test list

mvertens:
This PR fixes issue https://github.com/NorESMhub/BLOM/issues/706.

PR #708, opened 2026-01-12
Roll back latest changes to pe-layout

TomasTorsvik:
Roll back pe-layout changes from #707


JorgSchwinger (approved):
Thanks for fixing

mvertens (approved):
Thanks for the fix.

Issue #686, opened 2025-11-10
OMIP throughput NUOPC vs MCT


matsbn:
It seems running OMIP type simulations with noresm3_0_beta03b is 20-30% slower than using NorESM2.0.10 with identical BLOM code and a setup I believe is quite comparable. I had a suspicion before, but inspecting simulations with what I hope is more reliable timing in BLOM (https://github.com/NorESMhub/BLOM/pull/682) seems to confirm this.

I recently ran one year tests using BLOM including PR https://github.com/NorESMhub/BLOM/pull/682 on the tnx1v4 grid and 72 layers. The noresm3_0_beta03b based case is on Betzy here (calling it N3, hereafter):

`/cluster/projects/nn9560k/matsbn/NorESM/cases/NOIIAJRARYF8485OC_TL319_tn14_blom_timer2_20251105`

while the NorESM2.0.10 based case is here (calling it N2 hereafter):

`/cluster/projects/nn9560k/matsbn/NorESM/cases/N2NOIIAJRAOC_TL319_tn14_hyb_pnetcdf_timer2_20251110`

The PE-layout is the same in both cases. The CICE code differs with quite different throughputs according to the cesm_timing.* reports, but in neither simulations should BLOM be waiting. Also, I did not manage to use exactly the same type of JRA forcing, but I don't think that should matter either. In the cesm_timing.* reports, the total model throughputs are 14.3 and 18.3 simulated_years/day for N3 and N2, respectively. The BLOM timing statistics (at the end of ocn.log.* files) are for N3:
```
Timer statistics of group ModelAdvance:
      0.7052 sec average time for   8759 accumulations
   6177.1938 sec total time with contributions:
   1652.2219 sec  26.7471%  before blom_loop
   4508.4293 sec  72.9851%  blom_loop
     16.5426 sec   0.2678%  restart_write

Timer statistics of group blom_loop:
      0.5141 sec average time for   8759 accumulations
   4502.9234 sec total time with contributions:
      6.7672 sec   0.1503%  ocn_import
      5.7626 sec   0.1280%  ocn_stream_sss_interp
      0.0039 sec   0.0001%  ocn_stream_sst_interp
      5.3710 sec   0.1193%  ocn_stream_dust_interp
   4463.5047 sec  99.1246%  blom_step
      9.7191 sec   0.2158%  blom_accflds
     11.7950 sec   0.2619%  ocn_export

Timer statistics of group blom_step:
[...]

PR #709, opened 2026-01-28
Improved PE pelayouts for JRA compsets

JorgSchwinger:
This PR changes the PE layouts for BLOM JRA-forced compsets in connection with #686 and the fix provided in https://github.com/NorESMhub/CDEPS/pull/31 

With the new PE-layout the runoff is placed on it's own (exclusive) PEs to make sure it does not interfere with other components.

### DIFF
--- a/cime_config/config_pes.xml
+++ b/cime_config/config_pes.xml
@@ -321,14 +321,14 @@
       <pes pesize="L" compset="any">
         <comment>PE-layout with 500/640 pes yielding about 19.2/15.5 myears/wday for BLOM/total</comment>
         <ntasks>
-          <ntasks_atm>6</ntasks_atm>
-          <ntasks_rof>6</ntasks_rof>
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):
@JorgSchwinger - Thanks! Looks fine to me.
Could you briefly summarize what tests you have carried out with this PR?

JorgSchwinger:
I have tested that throughput is indeed better with the new PE layout compared to the old one ( with the option `rof.ryf8485_jra:mapalgo=mapfile:/cluster/shared/noresm/inputdata/cpl/cpl6/map_JRA025_to_tnx1v4_e1000r300_170928.nc` added to `user_nl_drof_streams`). I did this for the `--pecount M,L,XL` options. I also checked that results are bit-for-bit, both with the old PE-layout and among each other.

I will merge this now, since there is no risk that this affects anything else than BLOM JRA compsets.

PR #714, opened 2026-03-04
Include machine settings for olivia in testlist

TomasTorsvik:
This PR enables running the BLOM and iHAMOCC standard tests suites on olivia.


matsbn (approved):
Thanks for enabling testing on Olivia. Looks good to me.

PR #717, opened 2026-03-11
Bugfix for accfields where k0200 can be zero

TomasTorsvik:
The original implementation did not include a check to determine if `k0200` was a valid index.


jmaerz (approved):
Looks fine for me, could be just faster to do ``k=k0200(i,j)`` as before instead of looking up the same field all the time.

TomasTorsvik:
> Looks fine for me, could be just faster to do `k=k0200(i,j)` as before instead of looking up the same field all the time.

Done.

PR #712, opened 2026-02-09
Additional and revised diagnostics for CMIP7

matsbn:
This PR makes several changes to the BLOM diagnostics to deliver additional variables requested by CMIP7, but also to correct and modify existing variables.

Changed variable names, but identical output:

- mld -> bld ("Boundary layer depth")
- maxmld -> maxbld ("Maximum boundary layer depth")
- mlts -> mldl82 ("Mixed layer thickness (Levitus, 1982)")
- mltsmn -> mldl82mn ("Minimum mixed layer thickness (Levitus, 1982)")
- mltsmx -> mldl82mx ("Maximum mixed layer thickness (Levitus, 1982)")
- voltr -> masstr ("Section mass transports")

New variables:

- mldb04 ("Mixed layer thickness (de Boyer Montégut et al., 2004)")
- mldb04mn ("Minimum mixed layer thickness (de Boyer Montégut et al., 2004)")
- mldb04mx ("Maximum mixed layer thickness (de Boyer Montégut et al., 2004)")
- t17d ("17C isoterm depth")
- heattr ("Section heat transports")
- salttr ("Section salt transports")

Modified computation of variables:

- mmflxl ("Overturning streamfunction", now with time invariant potential density vertical coordinate and consistent computation for all BLOM's vertical coordinate choices)
- mmftdl ("Overturning streamfunction due to thickness diffusion", now with time invariant potential density vertical coordinate and consistent computation for all BLOM's vertical coordinate choices)
- mmfsml ("Overturning streamfunction due to submesoscale transport", now with time invariant potential density vertical coordinate and consistent computation for all BLOM's vertical coordinate choices)
- mmflxd ("Overturning streamfunction", corrected integration to vertical depth axis)
- mmftdd ("Overturning streamfunction due to thickness diffusion", corrected integration to vertical depth axis)
- mmfsmd ("Overturning streamfunction due to submesoscale transport", corrected integration to vertical depth axis)
- volgs ("Sea water volume", corrected erroneous division by gravitational acceleration)

New namelist variables SIGDIA* are introduced to specify the time invariant potential density vertical coordinate for overturning streamfunction computation using the ALE method. With the ALE method, these diagnostic potential densities are now output as the "sigma" vertical coordinate in BLOM output files, while reference potential densities are output as "sigref". With isopycnic vertical coordinate and bulk mixed layer, the "sigma" vertical coordinate is as before in the output, but the computation of overturning with "sigma" coordinate is revised.

[...]

PR #718, opened 2026-03-12
Bugfix in 2 readfld lines in mod_restart

TomasTorsvik:
It seems two lines for `ACC_VMFLTDL` and `ACC_VMFLSML` were swapped in the earlier PR #712 

Error triggered at restart when running `N2N1850` compset.


matsbn (approved):
Thanks for finding and fixing this!

PR #719, opened 2026-03-20
Functionality for horizontal alignment of mesoscale eddy-induced diffusion within the mixed layer

matsbn:
When neutral mesoscale eddy-induced diffusion is enabled with namelist option `ltedtp = "neutral"`, this PR adds functionality so that interfaces of layers constructed for diffusive flux computations may be gradually aligned with the surface within the mixed layer. This functionality is enabled by setting a new namelist variable `ndiff_surface_align = .true.`. The default setting of `ndiff_surface_align = .false.` gives bit-identical results with current master.

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -1853,6 +1853,17 @@
     <desc>Type of lateral tracer eddy diffusion: Valid methods:</desc>
   </entry>
 
+  <entry id="ndiff_surface_align">
+    <type>logical</type>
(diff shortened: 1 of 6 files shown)


JorgSchwinger (approved):

TomasTorsvik (approved):

TomasTorsvik:
Looks fine to me.

TomasTorsvik:
The `aux_blom_noresm` tests PASS, with expected name list fail (NLFAIL) when comparing with `noresm3_0_beta12a`.

PR #720, opened 2026-03-25
Adjusted iHAMOCC output frequencies for CMIP7 cmorization

JorgSchwinger:
This PR adjust the default settings for iHAMOCC output in `namelist_definition_blom.xml`, mainly
* add daily frequency for some of the newly added outputs
* remove outputs on layers almost completely (with some exceptions for N-cycle related fields)
* add monthly outputs for diffusive fluxes between sediment and water column

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -6438,7 +6438,7 @@
     <category>diabgc</category>
     <group>diabgc</group>
     <values>
-      <value>0,2,2</value>
+      <value>4,2,2</value>
       <value is_test="yes">4,2,2</value>
       <value is_test="yes" empty_hist="yes">0,0,0</value>
       <value hamocc_output_size="spinup">2,2</value>
@@ -6534,7 +6534,7 @@
       <value is_test="yes" empty_hist="yes">0,0,0</value>
       <value hamocc_output_size="spinup">2,2</value>
     </values>
-    <desc>Dissolved carbon (dissic) [mol C m-3]</desc>
+    <desc>Surface dissolved carbon (srfdissic) [mol C m-3]</desc>
   </entry>
 
   <entry id="srf_phyto">
@@ -6547,7 +6547,7 @@
       <value is_test="yes" empty_hist="yes">0,0,0</value>
       <value hamocc_output_size="spinup">2,2</value>
     </values>
-    <desc>Phytoplankton (phyc) [mol P m-3]</desc>
+    <desc>Surface phytoplankton (srfphyc) [mol P m-3]</desc>
   </entry>
 
   <entry id="srf_ph">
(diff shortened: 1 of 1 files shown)


cime_config/namelist_definition_blom.xml
```diff
@@ -7275,7 +7275,7 @@
       <value is_test="yes" empty_hist="yes">0,0,0</value>
       <value hamocc_output_size="spinup">2,2</value>
     </values>
-    <desc>Primary production (pp) [mol C m-3 s-1]</desc>
+    <desc>Vertically integrated primary production (ppint) [mol C m-3 s-1]</desc>
```
jmaerz: Are the units correct here (i.e. shouldn't it be: m-2)?

TomasTorsvik (approved):
@JorgSchwinger - thanks for checking the output list!

cime_config/namelist_definition_blom.xml
```diff
@@ -7301,7 +7301,7 @@
       <value is_test="yes" empty_hist="yes">0,0,0</value>
       <value hamocc_output_size="spinup">2,2</value>
     </values>
-    <desc>Vertically integrated denitrification</desc>
+    <desc>Vertically integrated denitrification (dnitint)</desc>
```
jmaerz: I am not sure, if this is used at all for the extended nitrogen cycle  - I don't think so.

cime_config/namelist_definition_blom.xml
```diff
@@ -8174,7 +8174,7 @@
       <value is_test="yes" empty_hist="yes">0,0,0</value>
       <value hamocc_output_size="spinup">0,2</value>
     </values>
-    <desc>PP consumption of NH4 [mol NH4 m-3 s-1] - ext. N cycle only </desc>
+    <desc>PP consumption of NO3 [mol NO3 m-3 s-1] - ext. N cycle only </desc>
```
jmaerz, line 8177: Thanks for fixing!

cime_config/namelist_definition_blom.xml
JorgSchwinger: Ok, yes the field is all 0 in the output. Would it be possible to add this diagnostic? It is used to provide the output `frn=minus_tendency_of_ocean_mole_content_of_elemental_nitrogen_due_to_denitrification_and_sedimentation`

(Otherwise I will just remove `frn`)

jmaerz: Hi Jörg, I don't know, how to read the data request here. Is by definition denitrification incorporating both, denitrifcation and anammox losses of N to N2 (i.e. more the traditional way)? - or are these separate things in the CMIP output? And what is meant as loss by sedimentation? - is it just particulate N-flux to the sediment or is it the net-flux (i.e. particulate sedimentation - diffusive inorganix reflux)? - technically, all the stuff can be delivered by correct post-processing of output data (if written out).

JorgSchwinger: I don't know how to read the request either, I would interpret it in a way that makes sense for our model. `dnitint` would provide a compact way of storing this in a 2d field. You can have a look in the data request table. If there is a possibility to calculate `frn` please add, otherwise I'll remove it.

jmaerz: Ok, I see. I am a bit time constrained these days. I try to look into it asap. Is there a time that it needs to be finished (if not already better by yesterday)?

cime_config/namelist_definition_blom.xml
JorgSchwinger: Corrected

cime_config/namelist_definition_blom.xml
JorgSchwinger: For now I switched off `dnitint`.

PR #722, opened 2026-03-26
Fix OCN_CO2_type for HIST compset

JorgSchwinger:
This PR fixes the setting of `OCN_CO2_TYPE` for a historical compset. This might be a preliminary fix since we will need to work further on the logic how to set this variable based on the compset (see also discussion in https://github.com/NorESMhub/NorESM/issues/782).

@mvertens, note that for the spinup (and probably also piControl) we use `OCN_CO2_TYPE=constant`, so I didn't follow the MOM6 example.

### DIFF
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -62,6 +62,8 @@
     <default_value>constant</default_value>
     <values>
       <value compset="_BLOM.*%ECO"           >constant</value>
+      <value compset="1850_CAM"              >constant</value>
+      <value compset="HIST_CAM"              >diagnostic</value>
       <value compset="_BLOM.*%ECO.*_BGC%BPRP">prognostic</value>
       <value compset="_BLOM.*%ECO.*_BGC%BDRD">diagnostic</value>
       <value compset="_DATM%CPLHIST.*_BLOM.*%ECO">constant</value>
--- a/cime_config/config_component.xml
+++ b/cime_config/config_component.xml
@@ -62,8 +62,7 @@
     <default_value>constant</default_value>
     <values>
       <value compset="_BLOM.*%ECO"           >constant</value>
-      <value compset="1850_CAM"              >constant</value>
-      <value compset="HIST_CAM"              >diagnostic</value>
+      <value compset="_CAM"                  >diagnostic</value>
       <value compset="_BLOM.*%ECO.*_BGC%BPRP">prognostic</value>
       <value compset="_BLOM.*%ECO.*_BGC%BDRD">diagnostic</value>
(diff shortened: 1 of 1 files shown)


tjiputra:
@JorgSchwinger , I wonder if we should set instead (not constant):
<value compset="1850_CAM"              >diagnostic</value>

I was thinking of many compsets with '1850_CAM*' such as PiCtrl, 1%CO2, 4xCO2, flat10, etc. which by default should have OCN_CO2_TYPE to be diagnostic.

JorgSchwinger:
We are currently running spinup with `1850_CAM` compset and `OCN_CO2_TYPE=constant`. We could change this, but this would then affect any new experiments. Of course it should be the same if CAM also uses constant 284.7 ppm, which it should(?).

JorgSchwinger:
I checked the Run9 CAM settings for CO2 concentration, it seems to use a LBC-file with monthly resolution (there is a seasonal cycle), is this correct @DirkOlivie?

The land is using the diagnostic CO2 from the atmosphere, so it would be probably good to switch to it for the ocean, too. I will then change to:

<value compset="_CAM">diagnostic</value>

So every compset with active CAM will trigger the ocean to read CO2 from CAM.

TomasTorsvik:
> So every compset with active CAM will trigger the ocean to read CO2 from CAM.

Seems like a good idea to have this as the default setting.

tjiputra:
Thank you all. We still have to figure out how to deal with biogeochemically-coupled (BGC) and radiatively-coupled (RAD) experiments, which I think can be addressed (overwritten) by the combination use of BPRP/BDRD/etc in the compset names.

JorgSchwinger:
Previously, we used `OCN_CO2_TYPE=constant` for the RAD experiment, but if we now switch to a CO2 concentration that has a seasonal cycle in the spinup/piControl, this is no longer a good option. So yes, we have to figure this out.

JorgSchwinger:
I have now changed this as @mvertens originally suggested: A compset with active CAM will trigger `OCN_CO2_TYPE=diagnostic`, which is consistent with what land is doing. This also means that we will continue the iHAMOCC spinup with diagnostic CO2 from CAM instead of constant CO2 (once this is included in a NorESM-tag). Once we switch to CMIP7 forcings, this will also introduce a latitudinal gradient (which is good, but will need some additional spin-up, potentially).

@TomasTorsvik @tjiputra  @mvertens please review this PR, since it would be good to merge as soon as possible - thank you!

mvertens (approved):

TomasTorsvik (approved):

PR #723, opened 2026-03-30
Enabling BLOM CMIP7 output by default namelist

matsbn:
With default namelist, BLOM should now output the CMIP7 requested variables at appropriate frequencies. Tested not to change the model state in a NOIIAJRARYF8485OC on Betzy.

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -3516,7 +3516,7 @@
     <category>diaphy</category>
     <group>diaphy</group>
     <values>
-      <value>0,0,1</value>
+      <value>0,0,4</value>
       <value is_test="yes">4,4,0</value>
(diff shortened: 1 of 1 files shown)


TomasTorsvik (approved):

cime_config/namelist_definition_blom.xml
```diff
@@ -3516,7 +3516,7 @@
     <category>diaphy</category>
     <group>diaphy</group>
     <values>
-      <value>0,4,0</value>
+      <value>0,0,1</value>
```
JorgSchwinger: Is the value of 1 correct? Should be probably 4?

JorgSchwinger (approved):

cime_config/namelist_definition_blom.xml
matsbn: Thanks for catching this, @JorgSchwinger! Corrected now.

PR #725, opened 2026-04-10
Alllow to run paramgen with python that does not have setuptools

mvdebolskiy:
Make paramgen work when setuptools are not installed with your python.


JorgSchwinger (approved):

TomasTorsvik (approved):

TomasTorsvik:
@mvdebolskiy - Thanks for fixing!

PR #728, opened 2026-04-13
Make extNcycle support optional natDIC tracers

### LINKED ISSUE #726: Natural DIC tracers
For iHAMOCC, we currently miss the full support for natural DIC tracers for the default model version. It suspect it would be good to implement this before starting the historical simulation. Has anyone worked on this over the past days? - if so, please could you let us know via this issue.

jmaerz:
Hi @JorgSchwinger an @tjiputra , with this PR, the extended nitrogen cycle now supports the natural DIC tracers. I haven't much tested it (compiles and runs in 1D vertical water column, but untested on betzy/olivia). I stayed with the shortcut for sediment biogeochemistry (I might look into it later). 

@TimotheeBrgs , in case that you also want to use the natDIC tracers, it seems as if they're currently not fully supported by the `river2OMIP` setting (e.g. in `mo_ocprod.F90`) 

Remaining issues after merging this branch:
- run in `OCN_CO2_TYPE=diagnostic` mode, CAM seems to feature an atmospheric CO2 concentration of about 284.957ppm (for the 1st month) 
- natDIC still isn't supported in the sediment   (see issue #730)
- not sure, how to proceed with the new `>=beta13` version; there was no preference within our group on this matter  
- river2omip not fully supported (see issue #731) 

Closes #726.

### DIFF
--- a/hamocc/mo_accfields.F90
+++ b/hamocc/mo_accfields.F90
@@ -199,8 +199,8 @@ contains
     real(rp) :: d13C(kpie,kpje,kpke)
     real(rp) :: d14C(kpie,kpje,kpke)
     real(rp) :: bigd14C(kpie,kpje,kpke)
-    real(rp) :: intpoc(kpie,kpje)               ! Vertically integrated detritus concentration 
-	real(rp) :: srfph(kpie,kpje)                ! Temporary variable for calculation of surface pH
+    real(rp) :: intpoc(kpie,kpje)               ! Vertically integrated detritus concentration
+    real(rp) :: srfph(kpie,kpje)                ! Temporary variable for calculation of surface pH
     real(rp) :: srfco3satarag(kpie,kpje)        ! Mole carbonate anion (CO3) for sea water in equilibrium with pure Aragonite @sfc
     real(rp) :: phyc_200(kpie,kpje)             ! phytoplankton 200m
     real(rp) :: ph_200(kpie,kpje)               ! pH 200m
@@ -208,8 +208,7 @@ contains
     real(rp) :: co3satarag_200(kpie,kpje)       ! Mole carbonate anion (CO3) for sea water in equilibrium with pure Aragonite @200m
     real(rp) :: o2_200(kpie,kpje)               ! o2 200m
     real(rp) :: o2min(kpie,kpje)                ! o2 minimum closest to surface
-	
-	real(rp),parameter :: himin = 1.0e-11       ! Minimum [H+] for calculation of pH
+    real(rp),parameter :: himin = 1.0e-11       ! Minimum [H+] for calculation of pH
 
     if (use_cisonew) then
       ! Calculation d13C, d14C and Dd14C: Delta notation for output
@@ -234,7 +233,7 @@ contains
     endif
 
     intpoc        (:,:)=0._rp
-	srfph         (:,:)=0._rp
+    srfph         (:,:)=0._rp
     srfco3satarag (:,:)=0._rp
     phyc_200      (:,:)=0._rp
     ph_200        (:,:)=0._rp
@@ -258,12 +257,12 @@ contains
             o2_200(i,j)         = ocetra(i,j,k,ioxygen)
             ph_200(i,j)         = -log10(max(hi(i,j,k),himin))
          endif
-		  
+
           ! integrated POC in kmol P m-2
           do k=1,kpke
             intpoc(i,j) = intpoc(i,j) + ocetra(i,j,k,idet)*pddpo(i,j,k)
           enddo
-          
+
           ! O2 minimum closest to surface (but below mixed layer)
           k1 = kmle(i,j)
           o2min(i,j) = ocetra(i,j,k1,ioxygen)
@@ -384,7 +383,7 @@ contains
     call accsrf(jsrfsilica,ocetra(1,1,1,isilica),omask,0)
     call accsrf(jsrfdic,ocetra(1,1,1,isco212),omask,0)
     call accsrf(jsrfphyto,ocetra(1,1,1,iphy),omask,0)
-	srfph = -log10(max(hi(:,:,1),himin))
+    srfph = -log10(max(hi(:,:,1),himin))
     call accsrf(jsrfph,srfph,omask,0)
     call accsrf(jdms,ocetra(1,1,1,idms),omask,0)
     call accsrf(jsrfpn2om,pn2om,omask,0)
@@ -413,7 +412,7 @@ contains
       call accsrf(jsrfnatdic,ocetra(1,1,1,inatsco212),omask,0)
       call accsrf(jsrfnatalk,ocetra(1,1,1,inatalkali),omask,0)
       call accsrf(jnatpco2,natpco2,omask,0)
-	  srfph = -log10(max(nathi(:,:,1),himin))
+      srfph = -log10(max(nathi(:,:,1),himin))
       call accsrf(jsrfnatph,srfph,omask,0)
     endif
     if (use_BROMO) then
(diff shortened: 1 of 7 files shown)


JorgSchwinger (approved):
@jmaerz  thanks a lot for fixing this! Could you also please adjust the value of `atm_co2_nat` (in `mo_para_bgc`) to the value used during NorESM3 the spin-up?

I also realize that we have a problem with the atmospheric CO2 concentration. After #722 (which will become effective for beta13) we will no longer use `OCN_CO2_TYPE=constant` which implies that we do not have a simple constant reference for `atm_co2_nat`.

I'm not sure what to do about this. Any suggestions?

@jmaerz would be nice to test it in an OMIP setting, by turning it on an checking that the `nat*` fields are bit identical with their counterparts (needs pre-beta13).

jmaerz:
Hi @JorgSchwinger , you refer to 284.7 ppm instead of the currently 284.32ppm, right?

With regards to the varying CO2: is there a way to always provide the spatio-temporally varying pre-industrial concentration, also during historical and scenario simulations? - then I would opt for using those all the time as boundary condition - that would be at least the cleanest way, as far as I understand.

JorgSchwinger:
> Hi @JorgSchwinger , you refer to 284.7 ppm instead of the currently 284.32ppm, right?

Yes, correct

JorgSchwinger:
> With regards to the varying CO2: is there a way to always provide the spatio-temporally varying pre-industrial concentration, also during historical and scenario simulations? - then I would opt for using those all the time as boundary condition - that would be at least the cleanest way, as far as I understand.

This is of course possible, but also quite a bit of additional work. This would need to come from an external input file.

jmaerz:
As far as I understand, the CO2 concentration is always coming from CAM (as of post-#722) - how is that CO2 generated/prescribed in CAM? - as far as I understand, CO2 is currently generally handled through the coupler (the moment that NorESM is run). I foresee the biggest problem with emission-driven, interactive simulations here. Would there be a possibility to always handle natDIC via DATM?  I don't now, how critical seasonality (and spatial distribution) is for the natural DIC tracers... - otherwise remain with the fix value.

JorgSchwinger:
In OMIP simulations we still use a constant CO2 value (not coming from CAM or DATM). For these kind of simulations the natDIC tracers still work straightforward.

For all other simulations I see three possibilities

1. We continue to use a constant scalar value of atmospheric CO2 for the and piControl. Then we would have a "correct" reference for `co2_atm_nat`.
2. We invest the work to make `co2_atm_nat` a 2d field that is read in from a climatology over the first x years of the piControl. This would work for both emission-driven and concentration-driven. The nat-tracers would differ though from the corresponding tracers in the piControl.
3. We do not use natDIC-tracers in fully coupled mode

I would actually vote for 3 - in fully coupled mode we do have a piControl, so the value of nat-tracers is quite limited anyway. For OMIP simulations, I still see a value in it (there you typically do not run a piControl)

TomasTorsvik (approved):

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , after some further debugging (for `use_sedbypass=True`), I have now run a JRA 1month simulation with bfb results for `nattalk`=`talk`, etc. For that I switched to `use_sedbypass=True` since we thus far don't have full support in the sediment for natDIC tracers (which would then result in non-bfb). I was running an additional simulation for the coupled system (compset `N1850`; using NorESM noresm3_0_beta11) - however, there, some discrepancies show up. I am currently not sure, what to read from it. Need to look deeper into it - if you have  suggestions, let me know.

JorgSchwinger:
Thanks a lot @jmaerz - very strange that you get differences in N1850, but not in a OMIP compset. How large are the differences? You use the sediment-bypass in both cases?

TomasTorsvik:
@jmaerz - I agree it looks strange that you get differences for N1850. Did you just include your BLOM branch for the build and ran with no further namelist changes? I can try to check with regression tests against the `noresm3_0_beta13` baseline simulations.

jmaerz:
Hi @JorgSchwinger , I will look deeper into it, but at first glance, it looks ice-related:

natDIC-biasDIC

Biases here for DIC - range is about -3e-5 to 3e-5 (for values of about 1 to 2 mol C/m3). Maybe the CO2 passed through the coupler is still different from the natDIC CO2 conc (but passed CO2 is raw CO2 concentration from the atmosphere, right)? 

@TomasTorsvik : tag 13 baseline cannot work, since there we clearly expect non-bfb. I didn't do many changes, just switched off the sediment.

TomasTorsvik:
@jmaerz - OK, thanks for clarifying.

jmaerz:
@JorgSchwinger , where did you get the updated value of 284.7ppm atmospheric CO2 from? - I ask, since to me, the picture attached indicates that the atmospheric concentration passed through the coupler might be higher than what we pre-set for natural DIC for the `N1850` compset (it is the same picture as above, but coloring makes it clearer - sorry for the initially bad color-coding). Ice as well as regions of outgassing seem to decrease differences. The graphics shows natDIC - DIC. 
natDIC-biasDIC

TomasTorsvik:
From `cmeps/cime_config/config_component_cesm.xml`:

```
  <entry id="CCSM_CO2_PPMV">
    <type>real</type>
    <valid_values></valid_values>
    <default_value>284.7</default_value>
    <values match="last">
      <value compset="^2000">367.0</value>
      <value compset="DATM.*_MOM6%[^_]*MARBL">284.317</value>
    </values>
    <group>run_co2</group>
    <file>env_run.xml</file>
    <desc>
      Mechanism for setting the CO2 value in ppmv for
      CLM if CLM_CO2_TYPE is constant or
      MOM6 if OCN_CO2_TYPE is constant.
    </desc>
  </entry>
```

jmaerz:
Hi @JorgSchwinger and @TomasTorsvik , Tomas and I, we just realized that there is output for `atmco2` - and it is spatially constant, with a value of 284.957ppm. I will setup a run now with this value for natDIC CO2 - and see if then the results become bfb for the coupled version as well. If it is the case, for some reason the coupler sends 284.957ppm instead of 284.7ppm.

jmaerz:
Strange, changing from `co2_atm_nat=284.7ppm` to `284.957ppm` alleviates some parts (seems to reduce the offset), but still differences visible.
 
natDIC-biasDIC-2

I start to wonder, if CAM/the coupler does something with a pre-set CO2 concentration.

JorgSchwinger:
There must be something weird with your setup (unless you already use beta13?). Pre-beta13 we are using constant CO2, and I just checked the spinup simulation ("Run10"), which correctly shows `atmco2=284.7`

Post-beta12a, I wouldn't expect bit-for-bit, because the CO2 comes through the coupler and has a seasonal cycle.

jmaerz:
@JorgSchwinger , yeah, after posting, I was thinking into the same direction... - I guess, I realized what happens - the BLOM `master` has already the 'diagnostic' switched on by default (instead of 'constant') triggered by CAM in the N1850 compset - even in the `beta11` that I use. I will try setting it explicitly now. Nevertheless, it is weird that CAM runs with a different CO2 concentration...

jmaerz:
Ok, after re-running and using `./xmlchange  OCN_CO2_TYPE=constant`, the natDIC tracers become bfb with the other DIC tracers. Hence, I would merge the branch, if you agree.

Remaining issues:
- run in `OCN_CO2_TYPE=diagnostic` mode, CAM seems to feature an atmospheric CO2 concentration of about 284.957ppm (for the 1st month) 
- natDIC still isn't supported in the sediment   
- not sure, how to proceed with the new `>=beta13` version; there was no preference within our group on this matter  
- river2omip not fully supported

JorgSchwinger:
@jmaerz thanks a lot for implementing and testing! I agree this should be merged now.

Regarding the fully coupled, the CAM concentrations that are read in, are varying monthly (e.g. in December they are 284.563). I would hope that the annual mean is ~284.7 (but I have not checked).

@jmaerz could you make an issue about natDIC for the sediment and R2OMIP that we do not forget about it?

TomasTorsvik:
@jmaerz - great that you solved this! I forgot that `master` already included the `diagnostic` switch.

Issue #729, opened 2026-04-17
Type bug in ncpack


TomasTorsvik:
It looks like there is a type bug in the subroutine `ncpack` in `mod_nctools.F90`:

```
    integer(kind = i2), parameter :: i2fill=-32768,i2max = 32767
    integer(i2), allocatable, dimension(:,:,:) :: fldout,fld_out
    real, allocatable, dimension(:,:,:) ::rfld

... (line 1402):
      !$omp parallel do private(j,i)
      do k = 1,kd
        do j = 1,jj
          do i = 1,ii
            fldout(i,j,k) = rfld(i,j,k)+i2fill    <-- ERROR in DEBUG mode
          end do
        end do
      end do
      !$omp end parallel do
      call xcgetrowint2(fld_out,fldout,kd)
      call ncerro(nfmpi_put_vara_int2_all(ncid,rhid,istart,icount,fld_out))
```

**Temporary fix**: #733 / Tag `v1.12.39`
- Use `real4` instead of `int2` when writing atmospheric CO2 to netCDF output.

TomasTorsvik:
See issue https://github.com/NorESMhub/noresm3_dev_simulations/issues/395

TomasTorsvik:
The `ncpack` subroutine looks to be fragile with respect to round-off errors. I write out in `mod_nctools.F90` after line 1340:

```
 ncpack1: fldmin,fldmax,scf,ofs:    284.780861509092     
   284.780861509092       4.336941042879117E-018   284.780861509092 
```

For a constant field we should have `scf = 1.d0`, but with the low value of `scf` we get in trouble when we do 
```
                rfld(i,j,k) = nint(((fld(ijk)*sfac)+offs-ofs)/scf)-i2fill
```
and later on
```
            fldout(i,j,k) = rfld(i,j,k)+i2fill
```
where `fldout` is an integer(i2) array.

JorgSchwinger:
I agree with your assessment @TomasTorsvik , according to your printout the field is constant but nevertheless it does not meet the condition 

`fldmin >= fldmax`  in line 1332

upon which `scf` would be correctly set to `1.d0` (could you print out more digits?).

So, a constant field coming from HAMOCC does not pose any problem, while if a constant CAM-field that has been interpolated to the BLOM-grid and passed to HAMOCC is prone to this type of (roundoff) error(?)

TomasTorsvik:
Here is the output with full precision `write(lp, '(g0)')` :
```
[...]

PR #733, opened 2026-04-19
Work-around for writing atmospheric CO2

TomasTorsvik:
This PR does not solve the problem with writing atmco2, but provides a work-around (writing real4 instead of int2 output) so that tuning simulations may continue while we work on the underlying problem.

REF: #729


mvertens (approved):

JorgSchwinger (approved):

jmaerz (approved):
Sounds reasonable as workaround.

PR #735, opened 2026-04-22
Bugfix of type conversion issue in packed format outputting

### LINKED ISSUE #729: Type bug in ncpack

matsbn:
This PR fixes a floating point to integer conversion issue in packed format outputting with scale factor and offset as reported in https://github.com/NorESMhub/BLOM/issues/729. The work-around in https://github.com/NorESMhub/BLOM/pull/733 has also been reverted. A case similar to https://github.com/NorESMhub/noresm3_dev_simulations/issues/395 has been run and successfully writes `atmco2` in packed format with `DEBUG=TRUE`.

Fix #729

### DIFF
--- a/phy/mod_nctools.F90
+++ b/phy/mod_nctools.F90
@@ -1214,7 +1214,7 @@ contains
     character(len=4) :: c4
     integer :: i,j,ij,ijk,k,n,kd
     integer, parameter :: maxdm=5, ijdm = (idm+2*nbdy)*(jdm+2*nbdy)
-    real :: scf,ofs,arng(2),fldmin,fldmax
+    real :: ri2max,scf,ofs,arng(2),fldmin,fldmax
     logical :: uvflg
     integer, dimension(maxdm) :: start,count
     integer(i2), allocatable, dimension(:,:,:) :: fldout,fld_out
@@ -1223,6 +1223,7 @@ contains
     real, dimension(itdm,jtdm) :: rfldt
 
     ! --- Initialise fields
+    ri2max = i2max
     uvflg = .false.
     kd = 1
 
@@ -1362,7 +1363,9 @@ contains
             do i = 1,ii
               ij = i+nbdy+(idm+2*nbdy)*(j+nbdy-1)
               if (msk(ij) == 1) then
-                rfld(i,j,k) = nint(((fld(ij+(k-1)*ijdm)*sfac)+offs-ofs)/scf) - i2fill
+                rfld(i,j,k) = &
+                  nint(max(-ri2max,min(ri2max, &
+                       (((fld(ij+(k-1)*ijdm)*sfac)+offs)-ofs)/scf)))-i2fill
               else
                 rfld(i,j,k) = 0
               end if
@@ -1378,7 +1381,9 @@ contains
               ij = i+nbdy+(idm+2*nbdy)*(j+nbdy-1)
               ijk = ij+(k-1)*ijdm
               if (msk(ij) == 1.and.fld(ijk) /= fillr8) then
-                rfld(i,j,k) = nint(((fld(ijk)*sfac)+offs-ofs)/scf)-i2fill
+                rfld(i,j,k) = &
+                  nint(max(-ri2max,min(ri2max, &
+                       (((fld(ijk)*sfac)+offs)-ofs)/scf)))-i2fill
               else
                 rfld(i,j,k) = 0
               end if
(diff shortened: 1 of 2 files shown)


JorgSchwinger (approved):
Thanks a lot for fixing this @matsbn

phy/mod_nctools.F90
```diff
@@ -1362,7 +1363,9 @@ subroutine ncpack(vnm,dims,fld,msk,mskflg,sfac,offs)
             do i = 1,ii
               ij = i+nbdy+(idm+2*nbdy)*(j+nbdy-1)
               if (msk(ij) == 1) then
-                rfld(i,j,k) = nint(((fld(ij+(k-1)*ijdm)*sfac)+offs-ofs)/scf) - i2fill
+                rfld(i,j,k) = &
+                  nint(max(-ri2max,min(ri2max, &
+                       (((fld(ij+(k-1)*ijdm)*sfac)+offs)-ofs)/scf)))-i2fill
```
TomasTorsvik, line 1368:
```suggestion
                rfld(i,j,k) = max(i2fill,min(i2max, &
                       nint(((fld(ij+(k-1)*ijdm)*sfac)+offs)-ofs)/scf)))-i2fill))
```
I don't think we need to define a new variable `ri2max` for this purpose. Just using the `i2fill` and ̀`ì2max` variables should be enough.

TomasTorsvik (changes requested):

phy/mod_nctools.F90
JorgSchwinger: But you need to have the correct data type `i2fill` in 16-byte integer, `ri2max` is real

matsbn: The reason for using `ri2max` instead of `i2max` in the min/max expressions is for performance. If one uses `i2max`, two additional integer->real conversions are done for evaluating the min/max expressions which in this case happens in an innermost loop 3D loop. Probably not something that has an enormous impact (and `i2fill` should ideally be replaced by `ri2fill` in the same expression), but I would keep it as it does not impact clarity too much.

matsbn: I see now that you suggest to have the `min/max` outside of `nint`. I deliberately put the limiting inside `nint` since if the floating point result from `(((fld(ij+(k-1)*ijdm)*sfac)+offs)-ofs)/scf` is a value outside of the integer range, a floating-point invalid operation would occur.

TomasTorsvik: OK, I see your point regarding the floating-point invalid operation. However, I don't think `nint(((fld(ij+(k-1)*ijdm)*sfac)+offs)-ofs)/scf)))-i2fill` results in a 16-byte integer, as long as we don't do
```
nint(((fld(ij+(k-1)*ijdm)*sfac)+offs)-ofs)/scf)), i2) - i2fill
```

In terms of performance, it's probably more of an issue that we do implicit type casting both here and in line 1421:
```
      !$omp parallel do private(j,i)
      do k = 1,kd
        do j = 1,jj
          do i = 1,ii
            fldout(i,j,k) = rfld(i,j,k)+i2fill
          end do
        end do
      end do
      !$omp end parallel do
``

TomasTorsvik (approved):

TomasTorsvik:
Maybe we need to look at the netCDF read/write modules in general at some stage. I approve now, since this solves the immediate problem.

matsbn:
> Maybe we need to look at the netCDF read/write modules in general at some stage. I approve now, since this solves the immediate problem.

I certainly agree to that.

PR #736, opened 2026-04-22
Include sanity check for input in ncpack

TomasTorsvik:
I suggest to stop processing in `ncpack` if input field is close to constant, but includes round-off errors.

We could also consider allowing a tolerance of round-off errors, i.e. replacing the field with the offset value when writing the output if the field is "close enough" to constant.

Meanwhile, @matsbn  opened PR #735 , so probably this one is redundant.

### DIFF
--- a/phy/mod_nctools.F90
+++ b/phy/mod_nctools.F90
@@ -1321,6 +1321,21 @@ contains
       end if
       call xcmin(fldmin)
       call xcmax(fldmax)
+
+      ! --- Sanity check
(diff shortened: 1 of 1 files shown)


JorgSchwinger:
> Meanwhile, @matsbn opened PR #735 , so probably this one is redundant.

Yes, I agree. I think @matsbn solution is simple and effective, so I would go for it.

TomasTorsvik:
Closing: outdated.

PR #742, opened 2026-04-27
New input4MIPs historical N2O time series for years 1750-2022

### LINKED ISSUE #739: Extend N2O historical atmospheric data
Currently, historical N2O data are only available until 2014. Ideally, N2O should get extended for the full historical period (i.e. present day).

jmaerz:
This is a PR that replaces the historical atmospheric N2O concentrations by those provided through input4MIPs for CMIP7 (1750-2022).  

CAM is using a spatially and monthly varying variant of these data, which could be at some point coupled similar to CO2, which can be envisaged for NorESM3.X

Closes #739


JorgSchwinger:
> @DirkOlivie , do you agree with these data? - or are you using different data in the atmosphere for the respective years?
> 

Wouldn't it make sense to update to the CMIP7 historical data? This should be available on input4MIPs? But anyway, probably no big difference.

jmaerz:
Hi @JorgSchwinger , you refer to these ones(?):
https://input4mips-cvs.readthedocs.io/en/latest/dataset-overviews/greenhouse-gas-concentrations/
https://zenodo.org/records/14892947
https://zenodo.org/records/19243739

Anyone re-created all of them already?

JorgSchwinger:
Hi @jmaerz 

I meant the data available on input4MIPs

https://esgf-metagrid.cloud.dkrz.de/search?project=input4MIPs&activeFacets=%7B%22mip_era%22%3A%22CMIP7%22%2C%22dataset_category%22%3A%22GHGConcentrations%22%2C%22variable_id%22%3A%22n2o%22%7D

jmaerz:
Hi @JorgSchwinger , thanks for pointing me to the atm. N2O data provided through input4MIPs. I now fully replaced the previous data by the new input4MIPs N2O data for the historical period (1750-2022). Main differences are in the period between 1750-1850, which now seem to better consider the atmospheric lifetime of N2O. This might also be relevant for your/our simulations @DirkOlivie. Is CAM updated in this respect as well (i.e. receiving the very same N2O input data for N2O)?

DirkOlivie:
@jmaerz for CMIP7, we will use in CAM GHG concentrations from the file (on betzy) :
/cluster/shared/noresm/inputdata/atm/waccm/lb/cmip7_ghg_version20260607/ghg_cmip7_CMIP_CR-CMIP-1-0-0_1750-2023_monthly_gnz_0p5degLat_version20260207.nc
It has (also for N2O) a seasonal cycle and a latitudinal gradient, and runs for the years 1750-2023.  It is based on the data from inpu4MIPs.

jmaerz:
@DirkOlivie , thanks for the information. That would suggest to go the same route as for CO2 on the longer run towards NorESM3.5, I would suspect/expect.

JorgSchwinger:
> @DirkOlivie , thanks for the information. That would suggest to go the same route as for CO2 on the longer run towards NorESM3.5, I would suspect/expect.

In principle, BLOM/HAMOCC could get everything through the mediator (including CFCs), but it would need to work for both compsets with CAM and DATM, so it requires a bit of work on that side.

JorgSchwinger (approved):

PR #745, opened 2026-04-30
Remove the option lkwrbioz_off to simplify HAMOCC code

JorgSchwinger:
This PR removes the option to revert back to the old way of splitting between euphotic zone and deep ocean in HAMOCC, thereby greatly simplifying the code of `mo_ocprod`.

I have tested that the new code results in bit-for-bit identical results in an OMIP compset. For these tests, I have also activated C-isotopes and DOM-classes, and for a second test I activated C-isotopes and R2OMIP (both tests were bit-for-bit).

I also have discovered a bug (see line 590 of `ocprod`), which erroneously applies the relaxation of iron to deep ocean values two times. I have not fixed this bug, since this will change results. I will open an issue and run tests to find out how large the effect is once this PR is merged.

@jmaerz @TomasTorsvik I would appreciate if you could review this PR thoroughly since it is quite a large code change.

### DIFF
--- a/hamocc/mo_ocprod.F90
+++ b/hamocc/mo_ocprod.F90
@@ -108,8 +108,7 @@ contains
     use mo_control_bgc,   only: dtb,io_stdo_bgc,with_dmsph,                                        &
                                 use_BROMO,use_AGG,use_PBGC_OCNP_TIMESTEP,use_FB_BGC_OCE,           &
                                 use_AGG,use_cisonew,use_natDIC, use_WLIN,use_sedbypass,use_M4AGO,  &
-                                use_extNcycle,lkwrbioz_off,lTO2depremin,use_river2omip,            &
-                                use_DOMclasses
+                                use_extNcycle,lTO2depremin,use_river2omip,use_DOMclasses
     use mo_vgrid,         only: dp_min,dp_min_sink,k0100,k0500,k1000,k2000,k4000,kwrbioz,ptiestu
     use mo_vgrid,         only: kmle
     use mo_clim_swa,      only: swa_clim
@@ -142,9 +141,9 @@ contains
     real(rp) :: tco(nsinkmax),tcn(nsinkmax),q(nsinkmax)
     real(rp) :: atten,avphy,avanut,avanfe,pho,xa,xn,ya,yn,phosy
     real(rp) :: avgra,grazing,avsil,avdic,graton
-    real(rp) :: gratpoc,grawa,bacfra,phymor,zoomor,excdoc,exud
-    real(rp) :: export, delsil, delcar, sterph, sterzo, remin
-    real(rp) :: docrem,opalrem,remin2o,aou,refra,pocrem,phyrem,tdoclc_rem,tdochc_rem
+    real(rp) :: gratpoc,grawa,phymor,zoomor,excdoc,exud
+    real(rp) :: export, delsil, delcar, remin
+    real(rp) :: docrem,opalrem,remin2o,aou,refra,pocrem,tdoclc_rem,tdochc_rem
     real(rp) :: zoothresh,phythresh
     real(rp) :: temp,temfa,phofa                  ! temperature and irradiation factor for photosynthesis
     real(rp) :: absorption,absorption_uv
@@ -156,7 +155,7 @@ contains
     real(rp) :: tiny_val = epsilon(1._rp)
     real(rp) :: zeu
     ! DOM
-    real(rp) :: exudsl,excdocsl,bacfrasl,docremsl,bacfrasr,docremsr,bacfrar,docremr,doclimfct
+    real(rp) :: exudsl,excdocsl,docremsl,docremsr,docremr,doclimfct
     ! sedbypass
     real(rp) :: florca,flcaca,flsil,flnatcaca
     ! cisonew
@@ -167,7 +166,6 @@ contains
     real(rp) :: grazing13,grazing14
     real(rp) :: graton13,graton14
     real(rp) :: gratpoc13,gratpoc14
-    real(rp) :: bacfra13,bacfra14
     real(rp) :: phymor13,phymor14
     real(rp) :: grawa13,grawa14
     real(rp) :: zoomor13,zoomor14
@@ -176,12 +174,9 @@ contains
     real(rp) :: export13,export14
     real(rp) :: delcar13,delcar14
     real(rp) :: dtr13,dtr14
-    real(rp) :: sterph13,sterph14
-    real(rp) :: sterzo13,sterzo14
     real(rp) :: pocrem13,pocrem14
     real(rp) :: docrem13,docrem14
     real(rp) :: tdoclc_rem13,tdochc_rem13,tdoclc_rem14,tdochc_rem14
-    real(rp) :: phyrem13,phyrem14
     real(rp) :: rem13,rem14
     real(rp) :: rco213,rco214,rdoc13,rdoc14,rdet13,rdet14
     real(rp) :: rtdoclc13,rtdochc13,rtdoclc14,rtdochc14
@@ -294,7 +289,7 @@ contains
           absorption    = 1._rp
           absorption_uv = 1._rp
 
-          vloop: do k = 1,merge(kpke,kwrbioz(i,j),lkwrbioz_off)
+          vloop: do k = 1,kpke
 
             if(pddpo(i,j,k) > 0.0_rp) then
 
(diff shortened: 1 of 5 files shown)


TomasTorsvik:
@JorgSchwinger - A general question:
Once we merge this, will this mean that we are no longer able to run the current `master` branch with NorESM2.x variants that use the bulk mixed layer formulation (perhaps NorESM2.5)? Or did we already break compatibility with the NorESM2.x framework earlier? We made some backwards-incompatible changes in v1.8 for the restart files, so that breaks with NorESM2.3, but we have not thrown out much functionality yet.

JorgSchwinger:
@TomasTorsvik you still can run this with NorESM2, there are no changes in any interface. What we do not have though, is a good tuning for use with isopycnic coordinates, is this correct @jmaerz ?

TomasTorsvik:
> @TomasTorsvik you still can run this with NorESM2, there are no changes in any interface. What we do not have though, is a good tuning for use with isopycnic coordinates, is this correct @jmaerz ?

OK, I suppose we are gradually moving away from supporting isopycninc coordinates as well. 
Still, I think this is a sufficient functionality change to justify bumping the version branch from `v1.12` to `v.1.13`.

hamocc/mo_ocprod.F90
```diff
@@ -595,54 +540,25 @@ subroutine ocprod(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,omask,ptho,pi_ph,psao,pp
...
-                                      &                       + tdochc_rem*rnit_tdochc
-              endif
-              ocetra(i,j,k,isco212) = ocetra(i,j,k,isco212) + tdoclc_rem*rcar_tdoclc               &
-                                    &                       + tdochc_rem*rcar_tdochc
-              ocetra(i,j,k,iiron)   = ocetra(i,j,k,iiron)   + (tdoclc_rem+tdochc_rem)*riron
+              remin_aerob(i,j,k) = (dtr+phosy)*rnit               ! kmol N/m3/dtb - Aerob remin to ammonium  (var. sources)
             endif
             ocetra(i,j,k,idet) = ocetra(i,j,k,idet)+export
-            ocetra(i,j,k,idms) = ocetra(i,j,k,idms)+dmsprod-dms_bac-dms_uv
+            ocetra(i,j,k,idms) = ocetra(i,j,k,idms)+dmsprod-dms_uv
```
TomasTorsvik, line 548: Is this a bug-fix w.r.t bacterial decomposition of DMS? I see it is subtracted in two places in the original code, and only once after the update. But it is intermingled with `lkwrbioz_off`, so I might just be confused about the whole thing. Just to double-check that it is not something that was forgotten about in the re-write process.

TomasTorsvik (approved):
@JorgSchwinger - thanks for the update!
I don't see anything that looks suspicious, except perhaps the DMS part. The update simplifies the code, which is much appreciated!

jmaerz:
Hi @JorgSchwinger , I will check the code change later this afternoon.

hamocc/mo_ocprod.F90
JorgSchwinger: No bug fix: Here, it was set to zero if `lkwrbioz_off=.true.` (line 570 of the old code), therefore it could be removed here. The bacterial decomposition of DMS is calculated further below.

hamocc/mo_ocprod.F90
```diff
@@ -1061,59 +796,110 @@ subroutine ocprod(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,omask,ptho,pi_ph,psao,pp
...
+              ! none of the processes at the current time is assumed to change
+              ! the size distribution (subject to change)
+              ! NOTE that phosy, exud etc. are in kmol/m3!
+              ! Thus divide by avmass (kmol/m3)
+              !**********************************************************************
+
               if(avmass > 0._rp) then
                 avnos = ocetra(i,j,k,inos)
-                ocetra(i,j,k,inos) = ocetra(i,j,k,inos)-remin*avnos/avmass
+                if (use_DOMclasses) then
```
jmaerz: Hi @JorgSchwinger , I might be wrong, but I have two potential issues here: First, don't we miss a `remin` here? Second, since you seem to have completely combined the first and second loop (useful/necessary?), `avmass` might require an update before applying `remin` (as it was done before). But I might miss that somehow on the way, things cancel out and you left it out for good reason.

JorgSchwinger: Regarding remin, yes you are right - thanks for spotting this! Will fix it later today.

`avmass` is calculated at the beginning of the loop before `phy` and `det` are modified. I think this is necessary to keep consistency between `avmass` and `nos`, but I am open to changing this if you think this would be more correct @jmaerz )

jmaerz: Hi @JorgSchwinger , as far I read it from the code before, all phytoplankton growth related changes are applied as before to the aggregation scheme, however, for `remin`, a new/updated `avmass` was calculated in the previous second loop (pretty much at the beginning of former loop2) - which is currently not really reflected in the code as it is (as far as I can see/understand). To not get out of sync, to my understanding, this would need to be done before applying `remin` (since `iphy` and `idet` also have been updated) - but please have a second look.

JorgSchwinger: There are a few things that are differently updated compared to the old `lkwrbioz_off=.false.` option, e.g. `dms_bac`. For these things I do not think this makes a difference.

My idea was to do only one update of `nos` (and I forgot to include remin here). Following the philosophy of the original code this would be done with avmass calculated at the beginning of the loop (which is consistent with `nos` that is being updated).

I really don't know how sensitive this is to these numerical changes - if you think it's better I can split the update of `nos` again into two parts one after production and one after remineralization.

jmaerz: Hi @JorgSchwinger , the problem with it is that other tracers are updated twice in in the very same loop (which I admittedly find very confusing - I personally would not have merged the two loops, since it makes the code generally more error-prone). `nos` is required to track these changes in terms of number of particles, which gets out of sync, if one isn't updating things correctly. That's what I see from the former code. I don't have further insights into how critical that is. 

The following line could be added to approach it: 
```
ocetra(i,j,k,inos) = ocetra(i,j,k,inos)-remin*ocetra(i,j,k,inos)/(ocetra(i,j,k,iphy) + ocetra(i,j,k,idet)+pocrem))
```  
so, rewritten and writing one more line to show, where I would put it:
```
ocetra(i,j,k,inos) = ocetra(i,j,k,inos)+anosloss
ocetra(i,j,k,inos) = ocetra(i,j,k,inos)*(1._rp-remin/(ocetra(i,j,k,iphy) + ocetra(i,j,k,idet)+pocrem))
```
This should give the same result as before (potentially to be checked - running a 1D setup should suffice).

**Further - just realizing that this likely qualifies a former bug:** - not sure, when it was introduced
It should actually read: 
```
ocetra(i,j,k,inos) = ocetra(i,j,k,inos)+anosloss
ocetra(i,j,k,inos) = ocetra(i,j,k,inos)*(1._rp-pocrem/(ocetra(i,j,k,iphy) + ocetra(i,j,k,idet)+pocrem))
```
Note the change from `remin` to `pocrem` since `remin` also includes DOM remineralization, which isn't affecting the number of particles.

JorgSchwinger: @jmaerz 

I have updated this now - I reverted the AGG code parts back to their original position. This should now give bfb-results with the previous code (except I corrected the bug with `remin` instead of `pocrem`).

hamocc/mo_ocprod.F90
```diff
@@ -882,43 +668,36 @@ subroutine ocprod(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,omask,ptho,pi_ph,psao,pp
...
-                docremr  = min(docr_remin*exp(-179.e3_rp/(8.314_rp*(temp+tzero)))*ocetra(i,j,k,idocr),&
-                         &     doclimfct*ocetra(i,j,k,ioxygen)/o2csmp)
+                ! Fractions of doc(s) remineralized into nutrients following Loenborg et al. 2018 (Frontiers):
+                ! A=2.96e6 (day-1) (Theoretical Kc in the absence of Ea)
+                ! R=8.314 J/mol/K (Universal gas constant)
+                ! Ea=38kJ/mol (Activation energy for DOC_L decay)
+                ! A_sl=7.96e9 (day-1) (Theoretical Kc in the absence of Ea)
+                ! R=8.314 J/mol/K (Universal gas constant)
+                ! Ea_sl=58kJ/mol (Activation energy for DOC_L decay)
+                docrem   = min(docl_remin *exp( -38.e3_rp/(8.314_rp*(temp+tzero)))*ocetra(i,j,k,idoc),  doclimfct*ocetra(i,j,k,ioxygen)/o2csmp)
```
jmaerz: Hi @JorgSchwinger , please introduce a line break here and in the following lines. The long lines lead to compilation errors when testing locally for a 1D setup.

hamocc/mo_ocprod.F90
```diff
@@ -882,43 +668,36 @@ subroutine ocprod(kpie,kpje,kpke,kbnd,pdlxp,pdlyp,pddpo,omask,ptho,pi_ph,psao,pp
...
               else
                 docrem = min(remido*ocetra(i,j,k,idoc), doclimfct*ocetra(i,j,k,ioxygen)/o2csmp)
               endif
 
               if (use_river2omip) then
-                tdoclc_rem = min(rem_tdoclc*ocetra(i,j,k,itdoc_lc), doclimfct*ocetra(i,j,k,ioxygen)&
-                           &     /merge(ro2utammo_tdoclc,ro2ut_tdoclc,use_extNcycle))
-                tdochc_rem = min(rem_tdochc*ocetra(i,j,k,itdoc_hc), doclimfct*ocetra(i,j,k,ioxygen)&
-                           &     /merge(ro2utammo_tdochc,ro2ut_tdochc,use_extNcycle))
+                tdoclc_rem = min(rem_tdoclc*ocetra(i,j,k,itdoc_lc), doclimfct*ocetra(i,j,k,ioxygen)/merge(ro2utammo_tdoclc,ro2ut_tdoclc,use_extNcycle))
```
jmaerz: Same here - too long lines.

hamocc/mo_ocprod.F90
jmaerz: I can confirm that at least in 1D, it becomes bfb when using `remin`.

jmaerz:
Hi @JorgSchwinger , general question, do you want to remain with the current structure? As in: with the combined previous 1st and 2nd loop?

hamocc/mo_ocprod.F90
JorgSchwinger: Ok, will fix this.

JorgSchwinger:
> Hi @JorgSchwinger , general question, do you want to remain with the current structure? As in: with the combined previous 1st and 2nd loop?

From my perspective, I thought this was the aim, to combine euphotic zone and deep ocean in one loop. Computationally this is the most efficient (and I don't see that this would be more error prone). But I can also re-introduce a second loop for remineralization if this is the preferred way.

jmaerz:
Hi @JorgSchwinger , I guess, it's good for now. However, we may want to discuss about it at some point. Thus far, I understood the loop structuring as displaying the operator splitting that we do and that is now broken up (which is why i find it a bit more confusing and error prone).

jmaerz (approved):
Hi Jörg, I haven't checked again since your last changes - I suspect that you just changed the line breaks.

JorgSchwinger:
> Hi Jörg, I haven't checked again since your last changes - I suspect that you just changed the line breaks.

Yes, I have only changed the line breaks. I will merge now, I'm fine with re-considering the general structure of `ocprod` at a later stage.

TomasTorsvik:
@JorgSchwinger , @jmaerz , @matsbn , @milicak 
I have tagged this PR as `v1.13.0`, to indicate that it breaks a functionality that has been present in the `v1.12.x` line of tags.

PR #751, opened 2026-05-22
Fix bug in relaxation of fe, slightly adjust default parameters

JorgSchwinger:
This PR fixes a bug in the relaxation of iron that was applied twice. Default sinking parameter are also changed slightly. Together these changes have practically no impact on coupled model simulations. The iron fix slightly increase primary production (and thereby DMS fluxes), but this is compensated by increasing `wmin`. The latter together with a decrease in `wlin` is also an attempt to improve persistent nutrient biases throughout the water column (too high in the deep ocean to little above 2000m).


TomasTorsvik (approved):
@JorgSchwinger - great, thanks for fixing!

jmaerz (approved):
Hi @JorgSchwinger , thanks for fixing and re-tuning!

PR #753, opened 2026-06-10
Extend the depth range of diagnostic levels

matsbn:
The lower depth bound of the deepest diagnostic level is extended from 8000m to 9500 m so that the full depth of the tnx1v4 grid is covered.

@YanchunHe, this should be consistent with what we discussed. I verified that the global sum of `templvl` weighted with
```
wgt(i,j,lev) = (min(pdepth(i,j), depth_bnds(2,lev)) - min(pdepth(i,j), depth_bnds(1,lev)))*parea(i,j)*pbot(i,j)/pdepth(i,j)
```
now gives good consistency with `tempga` for monthly output. Note that the weight uses `pbot`, which is time-varying.


jmaerz (approved):
Looks reasonable to me. Thanks.

TomasTorsvik (approved):

TomasTorsvik:
I suppose this will be an answer changing update. Think we can still try to get it into `noresm3_0_beta20`.

JorgSchwinger:
Hi @matsbn 

would we need a similar change for HAMOCC outputs, too? As far as I remember the weights used for BGC outputs are separate for BLOM and HAMOCC. I might be wrong. (I'm at a conference right now so I have no time to check this).

TomasTorsvik:
We use `depthslev` and `depthslev_bnds` in `mo_bgcmean.F90` and `mo_ncout_hamocc.F90`. In both cases this is imported from the `mod_dia` module. However, we have a modified `weights` field derived from `depthslev_bnds` that is defined in the `bgczlv` subroutine of `mo_bgcmean`. 

So, potentially, we may need to adjust something for HAMOCC output.

TomasTorsvik:
Looking at this again, I see that the `bgczlv` subroutine in `mo_bgcmean` duplicates/mimics the `diazlv` subroutine in `mod_dia`. So if it works for `diazlv` I don't see why it should not work for `bgczlv` (but I may be missing some details).

matsbn:
I made a short 2 year test with the increased depth bound. I can confirm that HAMOCC inherits the BLOM depth bounds so that the maximum depth bound in the HAMOCC output is 9500 m. The output is at NIRD here
```
/datalake/NS9560K/noresm3/cases/NOIIAJRARYF8485OC_TL319_tn14_b18_dbnds9500_20260602/ocn/hist
```
and can be compared to
```
/datalake/NS9560K/noresm3/cases/NOIIAJRARYF8485OC_TL319_tn14_b18_20260602/ocn/hist
```
HAMOCC diagnostics comparing these simulations are here: https://ns2345k.web.sigma2.no/datalake/diagnostics/noresm/matsbn/NOIIAJRARYF8485OC_TL319_tn14_b18_dbnds9500_20260602/HAMOCC_DIAG/yrs1to2-NOIIAJRARYF8485OC_TL319_tn14_b18_20260602/index.html

milicak (approved):

PR #756, opened 2026-06-18
Pass vertical diffusivity and buoyancy frequency squared to iHAMOCC.

### LINKED ISSUE #683: Vertical eddy diffusivity and turbulent energy dissipation rate as input to iHAMOCC
Hi @matsbn and @milicak,
as discussed previously, for new model development in the EU-project BioGeoSea, it would be great, if you could provide 

- (small scale) vertical eddy diffusivity ($k_v$ units: m<sup>2</sup>/s; i.e. the -total- vertical diffusive mixing coefficient for tracers)
- turbulent energy dissipation rate ($\epsilon$ units: m<sup>2</sup>/s<sup>3</sup>)

both ideally at  grid centers (where ocean biogeochemistry tracers live) to iHAMOCC. $\epsilon$ will allow to calculate the Kolmogorov microscale and turbulent shear rate. For further clarification of naming, $k_v$ and $\epsilon$ are typically connected via: 

$$
k_v=c\frac{\epsilon}{N^2}
$$

according to Osborn and Cox 1972, where $c$ is typically $c=0.2$ (Lindborg & Brethouver 2008). However, I believe that the turbulence library CVMIX should provide more insightful and direct values. 

**Timing**
Providing these values to iHAMOCC is not super urgent, but it would be great, if they could come available latest early next year (i.e. January 2026). 

If there are any questions, don't hesitate to contact me.

matsbn:
For the purpose of obtaining turbulent energy dissipation, $`\epsilon`$, this PR enables passing of vertical diffusivity, $`K`$, and buoyancy frequency squared, $`N^2`$, to iHAMOCC. Then the Osborn relationship ([Osborn, 1980](https://doi.org/10.1175/1520-0485(1980)010%3C0083:EOTLRO%3E2.0.CO;2))

$$K\approx\frac{\Gamma\epsilon}{N^2}$$

can be used to compute the energy dissipation, where $`\Gamma=0.2`$ is a commonly used value.

The computation of vertical diffusivity and buoyancy frequency differs depending on which vertical coordinate is used. With the ALE method these quantities are primarily layer interface variables, while they are layer variables with isopycnic coordinate and bulk mixed layer. With the ALE method I have chosen to interpolate $`K`$ and $`N^2`$ from interfaces to layers, so that the variables passed to iHAMOCC are independent of the choice of vertical coordinate.

This PR also changes the tracer vertical diffusivity to be the same as for salinity (was same as temperature before). With the parameterization choices used so far, temperature and salinity diffusivities have been identical.

This code change does not change results, except for diagnostic variables `bfsq` and `bfsqlvl`, which will differ when the ALE method is used because layer buoyancy frequency are computed differently within a time step (more consistent now with the diffusivity computation).

### DIFF
--- a/phy/mod_ale_vdiff.F90
+++ b/phy/mod_ale_vdiff.F90
@@ -79,7 +79,7 @@ contains
               do nt = 1, ntr
                 trc_1d(k,nt) = trc(i,j,kn,nt)
               enddo
-              nutrc_1d(k) = Kdiff_t(i,j,k)
+              nutrc_1d(k) = Kdiff_s(i,j,k)
             end if
           enddo
 
--- a/hamocc/mo_hamocc4bcm.F90
+++ b/hamocc/mo_hamocc4bcm.F90
@@ -27,7 +27,7 @@ contains
 
   subroutine hamocc4bcm(kpie,kpje,kpke,kbnd,kplyear,kplmon,kplday,kldtday,pdlxp,pdlyp,pddpo,prho,  &
                         pglat,omask, dust,rivin,ndep,oafx,pi_ph,pfswr,psicomo,ppao,pfu10,ptho,psao,&
-                        patmco2,pflxco2,pflxdms,patmbromo,pflxbromo,                               &
+                        vdiff,bfsq,patmco2,pflxco2,pflxdms,patmbromo,pflxbromo,                    &
                         patmn2o,pflxn2o,patmnh3,pflxnh3,shelfmask)
 
     !***********************************************************************************************
@@ -104,6 +104,8 @@ contains
     real(rp),intent(in)  :: pfu10  (1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd)      ! absolute wind speed at 10m height [m/s]
     real(rp),intent(in)  :: ptho   (1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd,kpke) ! potential temperature [deg C].
     real(rp),intent(in)  :: psao   (1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd,kpke) ! salinity [psu.].
+    real(rp),intent(in)  :: vdiff  (1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd,kpke) ! vertical diffusivity [m2 s-1].
+    real(rp),intent(in)  :: bfsq   (1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd,kpke) ! buoyancy frequency squared [s-2]
     real(rp),intent(in)  :: patmco2(1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd)      ! atmospheric CO2 concentration [ppm] used in fully coupled mode
     real(rp),intent(out) :: pflxco2(1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd)      ! CO2 flux [kg/m^2/s].
     real(rp),intent(out) :: pflxdms(1-kbnd:kpie+kbnd,1-kbnd:kpje+kbnd)      ! DMS flux [kg/m^2/s].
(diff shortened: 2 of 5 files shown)


TomasTorsvik (approved):
@matsbn - Thanks for the update! Looks fine to me, but I suggest to wait for review from @jmaerz before merging.

phy/mod_ale_vdiff.F90
```diff
@@ -79,7 +79,7 @@ subroutine ale_vdifft(m, n, mm, nn, k1m, k1n)
               do nt = 1, ntr
                 trc_1d(k,nt) = trc(i,j,kn,nt)
               enddo
-              nutrc_1d(k) = Kdiff_t(i,j,k)
+              nutrc_1d(k) = Kdiff_s(i,j,k)
```
jmaerz, line 82: @matsbn this looks like it was previously the vertical temperature diffusivity (and now corrected to the salinity diffusivity) - does this mean that the code is answer-changing for iHAMOCC?

jmaerz (approved):
Dear @matsbn , many thanks for this PR - I really appreciate! - it is perfect timing as a basis for new developments - being then well in line with the CMIP7 BLOM version. 

Could you briefly summarize, what mixing the vertical diffusivity entails (i.e. which sources)?   

I will likely branch off, once merged into master, for new developments.

phy/mod_ale_vdiff.F90
jmaerz: Sorry, I should have read your comment more carefully first. I see that thus far, the two values were equivalent. Resolving this comment.

matsbn: No, I tested it and as expected it was bit-identical. We have the option to enable double diffusion, which will give different vertical diffusivities for temperature and salinity, but has not been used so far in NorESM3 dev simulations. Enabling double diffusion is not yet implemented as a namelist option and would require a code change.

jmaerz: Thanks for re-confirmation!

matsbn:
> Dear @matsbn , many thanks for this PR - I really appreciate! - it is perfect timing as a basis for new developments - being then well in line with the CMIP7 BLOM version.
> 
> Could you briefly summarize, what mixing the vertical diffusivity entails (i.e. which sources)?
> 
> I will likely branch off, once merged into master, for new developments.

There are various options for vertical mixing and it depends also on the choice of vertical coordinate. I can list the commonly chosen contributions to the vertical diffusivity, which is a sum of various terms. Note that the majority of the terms do not relate to a prescribed or prognostic energy source.

ALE method:
- Background diffusivity, which is a specified constant diffusivity with a latitude dependence.
- Strong specified diffusivity when stability is weak (mimics mixing due to hydrostatic instability).
- Vertical shear instability mixing using a prescribed diffusivity dependence on gradient Richardson number.
- Tidally driven mixing. Energy dissipation climatology used and converted to diffusivity.
- Diffusivity in the surface boundary layer. Diffusivity profiles by KPP that depends on momentum and buoyancy forcing.

Isopycnic coordinate and bulk mixed layer:
- Background diffusivity, which is a specified constant diffusivity with a latitude dependence.
- Strong specified diffusivity when stability is weak (mimics mixing due to hydrostatic instability).
- Vertical shear instability mixing using k-epsilon model. Prognostic energy dissipation converted to diffusivity.
- Tidally driven mixing. Energy dissipation climatology used and converted to diffusivity.
- The bulk mixed layer formulation does not specify a vertical diffusivity in the surface boundary layer, but we estimate a KPP like diffusivity for the vertical exchange between the two model layers in the bulk mixed layer region.

jmaerz:
Dear @matsbn , many thanks for this informative list - highly appreciated! For your information: the envisaged work in iHAMOCC heavily relies on the hybrid coordinates (and better resolved mixed layer+polar region pressure levels and I will build - for now - on the default CMIP7 BLOM physics and to great extent iHAMOCC settings). Many thanks again for your work!

TomasTorsvik:
Tagged this as `v1.14.2`

PR #759, opened 2026-06-25
Ativate CFCs for historical; Activate nat-tracers for historical and scenarios; Activate shelf-sea residence time

### LINKED ISSUE #757: CFCs not activated in historical compsets
We would like to have CFCs (and SF6) activated (`use_cfc=.true.`) in historical compsets. The automatic activation currently fails (see https://github.com/NorESMhub/noresm3_dev_simulations/issues/460#issuecomment-4752908396)

We have in `namelist_definition_blom.xml` the block
````
  <entry id="use_cfc">
    <type>logical</type>
    <category>config_bgc</category>
    <group>config_bgc</group>
    <values>
      <value>.false.</value>
      <value compset="HIST_CAM60%NORESM.*_BLOM.*%ECO">.true.</value>
      <value compset="20TR_DATM.*_BLOM.*%ECO">.true.</value>
    </values>
    <desc>activates HAMOCC CFC code</desc>
  </entry>
````
but this is not picked up for the new historical compsets, because it wrongly indicates `HIST_CAM60` instead `HIST_CAM70`. @mvertens could you have a quick look and confirm if this is the only problem, or are there other changes needed? (sorry I am not so familiar with the compset-naming logic)

### PR DESCRIPTION
**Pull request by JorgSchwinger**
This PR fixes #757 and also activates natural HAMOCC tracers for historical and scenarios. Shelf sea residence time is now switched on by default, which is needed for the last phase of spin-up of NorESM3-LM

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -60,6 +60,7 @@ def buildnml(case, caseroot, compname):
     din_loc_root              = case.get_value("DIN_LOC_ROOT")
     caseroot                  = case.get_value("CASEROOT")
     rundir                    = case.get_value("RUNDIR")
+    compset                   = case.get_value("COMPSET")
     continue_run              = case.get_value("CONTINUE_RUN")
     ninst_ocn                 = case.get_value("NINST_OCN")
     run_type                  = case.get_value("RUN_TYPE")
@@ -227,6 +228,8 @@ def buildnml(case, caseroot, compname):
         config['ocn_ncpl'] = str(ocn_ncpl)
         config['pio_typename_ocn'] = pio_typename_ocn
         config['pio_netcdf_format_ocn'] = pio_netcdf_format_ocn
+        config["ishist"] = "yes" if "HIST_" in compset else "no"
+        config["isscen"] = "yes" if "SCEN7" in compset else "no"  # compset naming to be confirmed!
         config["continue_run"] = "yes" if continue_run else "no"
         config['blom_vcoord'] = blom_vcoord
         config["blom_river_nutrients"] = "yes" if blom_river_nutrients else "no"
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by TomasTorsvik — APPROVED** [review]

**Review by jmaerz — APPROVED** [review]
Looks reasonable. Thanks for fixing.

File: cime_config/buildnml [hunk 1]
```diff
@@ -227,6 +228,8 @@ def buildnml(case, caseroot, compname):
         config['ocn_ncpl'] = str(ocn_ncpl)
         config['pio_typename_ocn'] = pio_typename_ocn
         config['pio_netcdf_format_ocn'] = pio_netcdf_format_ocn
+        config["ishist"] = "yes" if "HIST_" in compset else "no"
```
mvertens, line 231: This looks fine. It seems however that you are triggering only on a transient scenario - so do you really need two config variables? Could you just use one config variable 
`config["transient"] = "yes" if ("HIST_" in compset or "SCEN7" in compset) else "no"`

cime_config/namelist_definition_blom.xml
```diff
@@ -4979,8 +4978,8 @@
     <group>config_bgc</group>
     <values>
       <value>.false.</value>
-      <value compset="HIST_CAM60%NORESM.*">.true.</value>
-      <value compset="20TR_DATM.*_BLOM.*%ECO">.true.</value>
+      <value ishist="yes">.true.</value>
+      <value isscen="yes">.true.</value>
```
mvertens, line 4982: If you use the "transient" config setting mentioned above - then this would only be one line.

mvertens (changes requested):

cime_config/buildnml
JorgSchwinger: We need a distinction, because currently the CFCs are only on for the historical, while the nat-tracers for both historical and scenarios. So if we would not distinguish between the two, we would produce garbage-output for the CFCs in the scenarios.

mvertens:
@JorgSchwinger - thanks for the clarification. This looks good.

mvertens (approved):

JorgSchwinger:
@mvertens thank you for checking!

PR #762, opened 2026-08-10
Change default parameters according to actual use in recent tests

TomasTorsvik:
This PR is intended to align the default namelist definitions according to namelist settings used for NorESM3 in recent tests.

For 1-degree atmosphere & ocean (`ne30pg3_tn14` grid) I have used the additional setting
```
LAU10F = .85
```
Is there any way to set the value of `lau10f` dependent on the atmosphere grid, or should we think of some other solution, e.g. usermods?

### DIFF
--- a/cime_config/namelist_definition_blom.xml
+++ b/cime_config/namelist_definition_blom.xml
@@ -453,7 +453,7 @@
     <category>limits</category>
     <group>limits</group>
     <values>
-      <value>.06</value>
+      <value>.09</value>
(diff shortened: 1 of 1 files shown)


matsbn (approved):
Thanks, @TomasTorsvik, looks good! We can maybe get this in now and figure out the grid configuration dependent settings at a later PR.

TomasTorsvik:
@matsbn - yes, I suppose we can find a solution for the `ne30pg3` grid later on. I think a usermod setting would be the cleanest option, as this would not require BLOM to be aware of the atmospheric grid.

Issue #510, opened 2025-03-10
Test run fail for NOIIAOC20TR


TomasTorsvik:
I am extending the default BLOM system test suite to cover all pre-defined compsets. There run fails when testing `NOIIAOC20TR`, in `mo_carchm`.
```
103: forrtl: error (75): floating point exception
103: Image              PC                Routine            Line        Source             
103: cesm.exe           0000000004A6E2CB  Unknown               Unknown  Unknown
103: libc.so.6          000014A3C7454DB0  Unknown               Unknown  Unknown
103: cesm.exe           00000000043EE593  mo_carchm_mp_carc         985  mo_carchm.F90
103: cesm.exe           00000000043A5523  mo_carchm_mp_carc         454  mo_carchm.F90
103: cesm.exe           0000000003FFFC15  mo_hamocc4bcm_mp_         256  mo_hamocc4bcm.F90
103: cesm.exe           000000000367708F  mo_hamocc_step_mp          83  mo_hamocc_step.F90
103: cesm.exe           000000000357360A  mod_tracers_updat         163  mod_tracers_update.F90
103: cesm.exe           0000000002CA346B  mod_blom_step_mp_         230  mod_blom_step.F90
103: cesm.exe           0000000002C2EB2C  ocn_comp_nuopc_mp         910  ocn_comp_nuopc.F90
```

Line 985:  `subroutine carchm_solve_dicsat`
```
ah2o = sqrt((K1*co2_sat)**2 + 4.*ac*2.*K1*k2*co2_sat)
```

It is strange that the `NOIIAOC` compset does not fail. The difference is 
```
    <alias>NOIIAOC</alias>
    <lname>2000_DATM%IAF_SLND_CICE_BLOM%HYB%ECO_DROF%IAF_SGLC_SWAV</lname>
    ...
    <alias>NOIIAOC20TR</alias>
    <lname>20TR_DATM%IAF_SLND_CICE_BLOM%HYB%ECO_DROF%IAF_SGLC_SWAV</lname>
    ...
    <entry id="RUN_STARTDATE">
    compset="20TR_DATM%IAF_SLND_CICE_BLOM%HYB%ECO_DROF%IAF_SGLC_SWAV">1700-01-01</value>
```

TomasTorsvik:
I looked again at this test. In the `ocn.log` file I see
```
 (mod_nuopc_methods):(blom_exportflds): brf flux not sent to coupler
Returning at first coupling interval
[...]

PR #767, opened 2026-09-04
Update list of expected test fails

TomasTorsvik:
Just a bit of clean up to keep the list of expected failing tests up to date with actual performance.

We have 3 ERP test that are failing for the same reason, at the `COMPARE_base_rest` stage. ERP is testing for difference when running with different processor counts. It is known that BLOM output is sensitive to changing processor count. I propose to leave one ERP test, to ensure that we do not have a `RUN` stage error when changing the processor count (test only on betzy), and remove 2 redundant tests.

### DIFF
--- a/cime_config/testdefs/ExpectedTestFails.xml
+++ b/cime_config/testdefs/ExpectedTestFails.xml
@@ -44,20 +44,6 @@
     </phase>
   </test>
 
-  <test name="ERP_Ld3_P256.T62_tn14.NOINYOC.betzy_gnu">
-    <phase name="COMPARE_base_rest">
-      <status>FAIL</status>
-      <issue>BLOM#501</issue>
-    </phase>
-  </test>
-
-  <test name="ERP_Ld3_P256.T62_tn14.NOINYOC.olivia_intel">
-    <phase name="COMPARE_base_rest">
-      <status>FAIL</status>
-      <issue>BLOM#501</issue>
-    </phase>
-  </test>
-
   <!-- ===================================== -->
   <!-- aux_hamocc_noresm test suite failures -->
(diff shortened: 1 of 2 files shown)


cime_config/testdefs/ExpectedTestFails.xml
```diff
@@ -44,23 +51,21 @@
...
 
-  <!-- aux_hamocc_noresm test suite failures -->
-
-  <test name="SMS_D_Ld1.T62_tn14.NOINYOC.betzy_intel.blom-hamocc-hamocc6">
-    <phase name="RUN">
+  <test name="ERP_Ld3_P256.T62_tn14.NOINYOC.olivia_intel">
+    <phase name="COMPARE_base_rest">
       <status>FAIL</status>
-      <issue>BLOM#645</issue>
+      <issue>BLOM#501</issue>
```
jmaerz: Hi @TomasTorsvik , not sure, but should this potentially read #510?

jmaerz (approved):
Other than the previous comment, looks reasonable. Thanks!

cime_config/testdefs/ExpectedTestFails.xml
TomasTorsvik: Hi @jmaerz , the test connected with #510 has been removed, so that one is no longer relevant.
But now I'm thinking it could make sense to just remove some of the ERP tests, since they are all failing for the same reason. Will update the PR.

jmaerz: Hi @TomasTorsvik , I guess, my confusion arose from the fact that there is no issue #501 (only a discussion) ...

PR #768, opened 2026-09-07
Feature/fix pacemaker settings

TomasTorsvik:
Leave entries for "pacemaker_SST_file" and "pacemaker_mask_file" as `unset` by default.
Prevent errors in case pacemaker files are not found on the system.

Including these files only makes sense in combination with `usermods_dirs/<SST_pacemaker_experiment>`, hence the preferred solution would be to include paths to pacemaker files in a `user_nl_blom` file under a relevant `usermods_dirs` experiment.

Solves #766

PR #769, opened 2026-09-08
Feature/fix pacemaker setting v2

TomasTorsvik:
The fix in #768 was incomplete, since it left two variables in the `limits` namelist group that is not defined by default in BLOM.

I suggest to solve this issue by creating a separate `pacemaker` namelist group, that will be defined but not read in by default. I'm not sure if we should include the `pacemaker` entry in the `cime_config/buildnml`. I don't think it is a problem to have extra group entries in the `ocn_in` file that we do not use on a regular basis, but perhaps there is a better way to solve this?

Tested with `NorESM2.0.10` code base, `N1850frc2esm` compset.

### DIFF
--- a/cime_config/buildnml
+++ b/cime_config/buildnml
@@ -249,6 +249,7 @@ def buildnml(case, caseroot, compname):
         groups=['limits','diffusion','merdia','secdia','diaphy']
 
         groups.append('cwmod')
+        groups.append('pacemaker')
 
         if case.get_value("BLOM_VCOORD") == "cntiso_hybrid":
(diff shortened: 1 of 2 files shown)


TomasTorsvik:
@matsbn , @JorgSchwinger , @jmaerz , @milicak 
Could someone please look at this PR? It moves some optional pacemaker input files outside of the `limits` namelist grouop, so that we don't need to define them on default setups.

JorgSchwinger (approved):
If this fixes the problem it's fine with me

